PR #1, opened 2022-10-09
Draft: new TCP server design

GlenDC:
Goals:

- Non Blocking
- Minimise Allocations
- Be Tower-Compatible
- Allow for graceful shutdown


GlenDC:
Already merged in.

Issue #70, opened 2024-02-29
provide get_or_insert functionality to the `Context` struct for extensions


GlenDC:
We already have `get`.

Next to that we can also provide wrappers:

```
self.get_or
self.get_or_default
```

Where the latter would require `T` the extension to implement `Default`.

PR #73, opened 2024-03-03
Implement get_or_default and get_or for the `Context` struct

AlvinKuruvilla:
Attempts to address #70.

### DIFF
--- a/src/http/client/mod.rs
+++ b/src/http/client/mod.rs
@@ -3,4 +3,4 @@
 
 mod service;
 #[doc(inline)]
-pub use service::HttpClient;
+pub use service::{HttpClient, HttpClientError};
--- a/src/http/client/service.rs
+++ b/src/http/client/service.rs
@@ -36,11 +36,26 @@ impl Default for HttpClient {
 }
 
 #[derive(Debug)]
+/// Error type for the [`HttpClient`].
 pub enum HttpClientError {
+    /// The HTTP version is invalid.
     InvalidVersion(Version),
+    /// The host information is missing.
+    ///
+    /// This information is required to be able to establish an L4 connection,
+    /// to serve the request over.
     MissingHost,
+    /// The host information is invalid.
+    ///
+    /// (e.g. could not be parsed as a [`SocketAddr`])
+    ///
+    /// [`SocketAddr`]: std::net::SocketAddr
     InvalidHost(String),
+    /// An IO error occurred.
+    ///
+    /// (e.g. during a handshake process)
     IoError(std::io::Error),
+    /// An HTTP error occurred during the http handshake or transfer process.
     HttpError(Error),
 }
 
@@ -123,6 +138,9 @@ where
                     .uri
                     .scheme()
                     .map(|s| match s.as_str() {
+                        // TODO is this scheme mapping complete enough?
+                        // and should we fail on unknown schemes?
+                        // should this be a shared utility somewhere?
                         "http" => 80,
                         _ => 443,
                     })
@@ -165,7 +183,7 @@ where
 
                 ctx.spawn(async move {
                     if let Err(err) = conn.await {
-                        // TOD: should this error level / handling be configurable?
+                        // TODO: should this error level / handling be configurable?
                         tracing::error!("connection failed: {:?}", err);
                     }
                 });


src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
     pub fn get<T: Send + Sync + 'static>(&self) -> Option<&T> {
         self.extensions.get::<T>()
     }
+    /// Get a reference to an extension or T's default().
+    ///
+    /// An extension is a type that implements `Send + Sync + 'static`,
```
GlenDC, line 156: [pedantic] I do not think there is much value in repeating the same documentation everywhere, also harder to maintain. So as an extended doc (everything after the first line) I would simply refer to `Context::get` for more information.

Also might be useful to link the `default()` part above to the actual `Default` std trait.

src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
     pub fn get<T: Send + Sync + 'static>(&self) -> Option<&T> {
         self.extensions.get::<T>()
     }
+    /// Get a reference to an extension or T's default().
```
GlenDC, line 154: You miss a line here. It can help to run `just qa` (does require to install just), but you can check the `justfile` to see what it does in case you do not want it. Pretty certain that running `cargo fmt` alone will already add the line here.

src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
...
+    /// # ctx.insert(5i32);
+    ///
+    /// // Test existing value
+    /// assert_eq!(ctx.get_or_default::<i32>(), 5i32);
+    ///
+    /// // Test default value for a type not inserted before (assuming `f64::default()` is 0.0)
+    /// assert_eq!(ctx.get_or_default::<f64>(), 0f64);
+    /// ```
+
+    pub fn get_or_default<T: Send + Default + Clone + Sync + 'static>(&self) -> T {
```
GlenDC, line 180: We do not wish to clone, and we do wish to add. The entire purpose, and sorry if that wasn't clear from the ticket, is that it would also insert `T` using its `Default` implementation if none was defined yet.

So in pseudo code what you want to do is:

1. call just the get method to try to fetch it (this step you do correctly)
2. in case it is found you can return the reference as usual, otherwise continue to step 3
3. call extensions.insert to insert the default value
4. call extensions.get again this time just unwrap as you know it should exist

Notes:

- this will require the method to take `&mut self` as there's a potential mutation
- the previous note will ensure the unwrap is safe
- this also means you no longer need the `Clone` trait bound

[nitpick] I usually put my regular trait bounds in front of the marker trait bounds, so something like `T: Default + Send + Sync + 'static`.

---

Bonus points:

- even better is if you add the same methods first to `Extensions`, as that one wraps around a map, so there you can actually call directly also `get_or`, `get_or_default` and `get_or_else`. This way we never have to call `unwrap` ourselves and do you also simplify your logic :)

src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
...
+    /// ctx.insert(5i32);
+    ///
+    /// // Test existing value - should return the inserted value
+    /// assert_eq!(ctx.get_or::<i32>(10), 5);
+    ///
+    /// // Test with a fallback value for a type not inserted before (fallback to 2.5 for f64)
+    /// assert_eq!(ctx.get_or::<f64>(2.5), 2.5);
+    /// ```
+
+    pub fn get_or<T: Send + Sync + Clone + 'static>(&self, fallback: T) -> T {
```
GlenDC, line 200: Same comments as on `get_or_default` applies.

src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
...
+    /// let mut ctx = Context::default();
+    /// ctx.insert(5i32);
+    ///
+    /// // Test existing value - should return the inserted value
+    /// assert_eq!(ctx.get_or::<i32>(10), 5);
+    ///
+    /// // Test with a fallback value for a type not inserted before (fallback to 2.5 for f64)
+    /// assert_eq!(ctx.get_or::<f64>(2.5), 2.5);
+    /// ```
+
```
GlenDC, line 199: This whiteline needs to be removed.

src/service/context/mod.rs
```diff
@@ -151,6 +151,55 @@ impl<S> Context<S> {
...
+    /// # let mut ctx = Context::default();
+    /// # ctx.insert(5i32);
+    ///
+    /// // Test existing value
+    /// assert_eq!(ctx.get_or_default::<i32>(), 5i32);
+    ///
+    /// // Test default value for a type not inserted before (assuming `f64::default()` is 0.0)
+    /// assert_eq!(ctx.get_or_default::<f64>(), 0f64);
+    /// ```
+
```
GlenDC, line 179: This whiteline needs to be removed.

src/service/context/mod.rs
GlenDC: Can you also add `get_or_else` while you are at it, as that will in generally be preferred over `get_or`, unless you are dealing with a simple primitive value, which will rarely be the case I imagine.

GlenDC (changes requested):
Thank you very much for making your first contribution to Rama! I very much appreciate it. I left a review with a request for changes. Anything that's marked as `[nitpick]` or `bonus points` you are free to ignore if you do not wish to. I can do those as a follow up commit if you didn't do these.

I already appreciate very much your effort. If for some reason you do not agree with something or are puzzled with something I said, please do reach out in this PR or via Discord. I'll gladly help you further where you want me to. If you want me to finish of this PR as you no longer have the time or desire to continue on this PR, do just tell and I'll glady do the last mile for you.

Either way, thank you very much for your PR and take care.

AlvinKuruvilla:
Thank you for your detailed feedback, I appreciate it :)

AlvinKuruvilla:
Just wanted to note, to my knowledge, we cannot remove the ```Clone``` bound because calling ```insert ``` requires that bound

src/service/context/extensions.rs
```diff
@@ -98,6 +98,13 @@ impl Extensions {
         });
         ext
     }
+    /// Checks if a value of type T exists in the extensions map
+    pub fn contains<T: 'static + Send + Sync>(&self) -> bool {
```
GlenDC, line 102: Don't think you need this function. Contains serves not much use in actual usage from non-context code.

I think better that you implement `get_or_insert_default`, `get_or_insert` and `get_or_insert_with` directly in `Extensions` and then expose these as-is from the `Context`.

I do realize I am asking perhaps a lot from you so do let me know at any point if you need me to take over.

And of course if you do not agree, do let me know.

src/service/context/mod.rs
```diff
@@ -152,6 +152,50 @@ impl<S> Context<S> {
...
+    /// # let mut ctx = Context::default();
+    /// # ctx.insert(5i32);
+    ///
+    /// // Test existing value
+    /// assert_eq!(ctx.get_or_default::<i32>(), &5i32);
+    ///
+    /// assert_eq!(ctx.get_or_default::<f64>(), &0f64);
+    /// ```
+    pub fn get_or_default<T: Send + Default + Clone + Sync + 'static>(&mut self) -> &T {
+        if !self.extensions.contains::<T>() {
```
GlenDC, line 169: Again, this should be implemented directly in `extensions`. Do think you should first try to use `.get(...)` and drop the contains concept. The `get` public method tries to also fetch from a parent extension which is still a concept I am supporting. Might in the end never the less drop it, but for now it is there, and if possible I'll keep it. But contains is not something we need to expose and currently anyway not support the parent concept if I see your approach to it.

GlenDC:
Also do feel free to resolve the review comment/threads in this MR if you think you resolved them :) As you probably can easier see if you resolved something or not.

src/service/context/extensions.rs
AlvinKuruvilla: I want to clarify what is the difference between getting_or_insert and get_or_insert with ?

GlenDC: - `get_or_insert`: insert a value `t: T` which is pre-created (so the function takes a `t: T`)
- `get_or_insert_with`: takes a closure/function that creates a value `T`, so is lazy
- `get_or_insert_default`: also lazy, but instead of needing a closure it can just call `T::default()`

An analog example of this concept can be found in the std lib: <https://doc.rust-lang.org/std/option/#modifying-an-option-in-place>

GlenDC:
As a friendly reminder @AlvinKuruvilla is this PR something you want to pursuit further, or you want me to finish of the great work you started? either way fine for me, and do take care :)

AlvinKuruvilla:
Hi @GlenDC I apologize for the radio silence, I got stuck while trying to implement `get_or_insert_with` 😅 
I committed all my other changes which include moving the logic for all the requested functions into `Extensions` and making it a pass through call on `Context`. I appreciate your help and patience throughout this PR and I'm excited to continue to hopefully contribute to Rama. I believe the only thing left on the table is fixing the implementation for `get_or_insert_with` :)

PR #84, opened 2024-03-12
Recreate pull request for get_or_nsert and the other associated functions

AlvinKuruvilla:
Welp that was an accident. Today I learned discarding changes when syncing a fork deletes the associated PR. Anyway I'll reiterate what I said in my original PR:
Thank you for your time and patience @GlenDC and I apologize for the radio silence. I managed to `get_or_insert` working and moved the core logic into `Extensions` as you suggested. I got stuck on the implementation of `get_or_insert_with` so that is really the only thing left on the table for this PR. `get_or_insert_default` uses `get_or_insert_with` internally, so that one won't work either till the crash on `get_or_insert_with` is fixed. (I'm sure there's a better way to write `get_or_insert_with` too but I got stuck 😅 )


GlenDC:
Ok. Do you want me to finish it off for you @AlvinKuruvilla or you would prefer further guidance.  Both are ok for me

AlvinKuruvilla:
I think it would be better if you finished it off, I'm curios to see how you would implement the last part

GlenDC:
Ok. I’ll finish it off tonight then. Thanks for all your effort and collaboration. Feel free to join our discord to ask questions or guidance. And feel free to pick up another “good first” issue if you are thirsty for more. I’ll gladly mentor and guide where you want me to.

GlenDC:
@AlvinKuruvilla I finished it of in https://github.com/plabayo/rama/pull/85.

But honestly, you nailed it. There is indeed no way around unwrapping/expecting in insert with. Your logic was correct. I just cleaned it up a bit the MR, but the code is pretty much what you had, so good on you.

AlvinKuruvilla:
Awesome thank you for helping me bring it to the finish line

PR #103, opened 2024-03-18
switch from browserstack (free) to lambdatest (paid)

### LINKED ISSUE #102: switch to lambdatest (paid) for rama-fp
Currently we were using BrowserStack, however:

- if we ever have to go paid it will cost us more
- we can only use mobile if we pay (with lambdatest seems we can do this for free)

So let's switch to lambdatest instead...

GlenDC:
Closes #102.


GlenDC:
Replaced by new plan to go for testgrid.

PR #113, opened 2024-03-21
Use workspace dependencies where possible

### LINKED ISSUE #89: use workspace dependencies for common workspace dep versionning
We are already defining our common properties in the workspace's `Cargo.toml`, as a single source of truth.

In a similar fashion we can define versions for dependencies once and only once in that `Cargo.toml`. This would make sure we only have to update dependencies once for the entire workspace.

LukeMathWalker:
Fix #89.

I took `cargo-autoinherit` for a spin since there was an open issue about it 😁

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,6 +12,64 @@ categories = ["asynchronous", "network-programming"]
 authors = ["Glen De Cauwsemaecker <glen@plabayo.tech>"]
 rust-version = "1.75.0"
 
+[workspace.dependencies]
+anyhow = "1.0"
+async-compression = "0.4"
+base64 = "0.22"
+bitflags = "2.4"
+brotli = "3"
+bytes = "1"
+clap = "4.4"
+crossterm = "0.27"
+flate2 = "1.0"
+futures = "0.3.29"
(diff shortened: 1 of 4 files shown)


GlenDC:
Amazing @LukeMathWalker, gave it a review. Seems to be all okay indeed.
Thanks for that! and GG with the tool.

GlenDC (approved):

LukeMathWalker:
Uh, package renames, the fun stuff. I'll have a deeper look to understand what's going on there, thanks for highlighting!

`cargo update` only updates the lockfile to the latest semver compatible release. It'll never give you `0.26` if your manifest says `0.25`. Perhaps you wanted `cargo upgrade`?

GlenDC:
> Uh, package renames, the fun stuff. I'll have a deeper look to understand what's going on there, thanks for highlighting!

To be clear. Your tool's choice did work. It had chosen to do the extra version rule and package rename on the individual crate (submodule) level. Which did work. Just not what I wanted, and it also made my Cargo.toml IDE tooling not find the package. But somehow it did still work with cargo etc.

PR #117, opened 2024-03-23
UA Parser and Identifier

### LINKED ISSUE #108: ability to parse UA strings and selectors
- ua strings:
   - transform them quad info ` { platform_version, platform_name, version, name } `
   - these have no operator for version, meaning match as close as possible
 - ua selectors:
    - start with `sel:`:
         - need to come up with a selector language, but KISS

The result is something that allows us to select an UA based on the parsed result from above and what we have available in the DB.

GlenDC:
Closes #108.


GlenDC:
superseded by https://github.com/plabayo/rama/pull/182

Issue #93, opened 2024-03-16
Provide “and(matcher)” and “or(matcher)” methods to enum matchers


GlenDC:
We already have in these http/stream enum matchers `and_*` and `or_*` methods. However currently there is no way to do complex chaining for the matcher of that passed in argument.

eg you can’t chain these, or negate them.

Before:

```rust
HttpMatcher::method(get).and_header(“foo”).negate()
```

the above is wrong as you would get `!(M AND H)`, instead of the desired `M AND !H`, which now would be possible as a result of completing this issue:

```rust
HttpMatcher::method(get).and(HttpMatcher::header(“foo”).negate())
```

Same applies to the stream/L4 enum matcher.

dmackdev:
Hi @GlenDC, I would be interested in picking this up.

Could you clarify what you mean that the first example is wrong? I think the resulting expression `!(M AND H)`is the expected behaviour from the way the code reads. 

Are you just saying that there is currently no way to express the second expression `M AND !H` using the current builder methods?

GlenDC:
Feel free to pick it up @dmackdev, is all yours. And that is correct. Currently there is indeed no way to express `M AND !H`. 

You can already express `!M AND H`, as that is something like `matcher_a().negate().and_matcher_b()` (sorry for the pseudo code).

But `X AND !Y` is indeed impossible given that `matcher_x().and_matcher_y().negate` will instead be equal to `!(X AND Y)`, which equals to `!X OR !Y`. Either way, not what you might have desired.

So @dmackdev you are correct that it's not that the first example is wrong, my wording was a bit too strong there. It's more like you say yourself, there is currently no way to express the negation of the other one. Of course we could also provide `and_matcher_x_negate` for each method. However there are also other constructions that are currently not possible.

So providing `and` and `or` methods to accept the `Enum` matcher directly solves all those cases. And once someone also tackles #92 we can also support in the same enum any custom matcher as well. But that's a different ticket. Of course if you wish feel free to pick them up together if you have the time/energy/desire and do it in one PR if you wish. All good for me.

GlenDC:
So the issue is all yours @dmackdev. If you do decide to not pick it up, do let it be known. Otherwise have fun. You can ask questions or guidance here, via email or discord. Input, remarks, feedback can be given the same way. Thanks for your time, energy and effort.

dmackdev:
If the current matcher was not negated and we were `AND`ing with an `AND` already, I think it would be safe to append the new matcher like so:
[...]

Issue #92, opened 2024-03-15
Provide support for boxed custom matchers in layer enums


GlenDC:
For both stream (L4) and http layers we have enums to allow for single type groupings. As an escape hatch might be good to allow custom boxed variants as well.

dmackdev:
Hi @GlenDC, could you please provide some guidance for the type of the inner field of the new enum variant?

Should we use the `Matcher` trait, so something like:
```rust
pub enum HttpMatcherKind
    //...
    Predicate(Arc<dyn crate::service::Matcher<State, Request<Body>>>)
}
```

Or the `Fn` trait, similar to how it is done in the `http::layer::cors` module, e.g. [here](https://github.com/plabayo/rama/blob/main/src/http/layer/cors/allow_credentials.rs#L88)? Would this need as arguments all the arguments in the `Matcher::matches` trait method, or just the `Request`?

Is it correct that this change will require introducing generics for `HttpMatcherKind` and `HttpMatcher`, in order for the `Body` parameter to align with their `Matcher` trait implementations?

GlenDC:
I would do it like this:

```rust
pub enum HttpMatcherKind<State, Body> {
    //...
    Custom(Box<dyn crate::service::Matcher<State, Request<Body>>>)
}
```

So a `Box` instead of an `Arc`, and I would call it `Custom` or something like that.

It does indeed also mean that you'll need to introduce generic parameters, not much that can be done about that.
Let me know if introducing that gives difficulties in some use cases @dmackdev , we can always give this some more thought if it does. I think it should be okay as those parameters should be able to be inferred in any use case I can think of, e.g. it's not much different of how many Services require this knowledge as well.

Other then that you're good to go. Thanks again.

dmackdev:
I've attempted to add the `Custom` variant https://github.com/dmackdev/rama/pull/1, but I am running into some issues - I have left some comments on that PR. Would appreciate some guidance @GlenDC.

GlenDC:
> I've attempted to add the `Custom` variant in this PR: [dmackdev#1](https://github.com/dmackdev/rama/pull/1). I have branched off of https://github.com/dmackdev/rama/tree/add-http-matcher-methods. But I am running into some issues - I have left some comments on that PR. Would appreciate some guidance @GlenDC.

[...]

PR #130, opened 2024-03-29
Implement `HttpMatcher::and` and `HttpMatcher::or`

dmackdev:
Addresses part of  https://github.com/plabayo/rama/issues/92 and https://github.com/plabayo/rama/issues/93 for `http` layer.

- Modifies `HttpMatcherKind::All` and `HttpMatcherKind::Any` variants to contain `Vec<HttpMatcher>` instead of `Vec<HttpKind>`. This way the negation of each matcher is preserved, and this fixes the bug I mentioned in the above issue.
- Implements `HttpMatcher::and` and `HttpMatcher::or` methods.
- Updates existing `HttpMatcher::and_*` and `HttpMatcher::or_*` methods to use new methods.
- Adds `HttpKind::Custom(_)` variant for custom matcher implementations.
- Adds tests.

### DIFF
--- a/src/http/matcher/mod.rs
+++ b/src/http/matcher/mod.rs
@@ -703,7 +703,6 @@ where
 
 #[cfg(test)]
 mod test {
-    use http::{HeaderName, HeaderValue};
     use itertools::Itertools;
 
     use crate::service::Matcher;
@@ -855,37 +854,6 @@ mod test {
 
             let matcher = (a.or(b)).and(c.or(d)).and(e.negate());
             let req = Request::builder().body(()).unwrap();
-            assert_eq!(
-                matcher.matches(None, &Context::default(), &req),
-                expected,
-                "({:#?}).matches({:#?})",
-                matcher,
-                req
-            );
-        }
-    }
-
-    #[test]
-    fn test_matcher_negation_and_header_contains() {
-        let matcher = HttpMatcher::method_get().negate().and_header_contains(
-            HeaderName::from_static("content-type"),
-            HeaderValue::from_static("text/plain"),
-        );
-
-        let cases = [
-            ("PUT", "text/plain", true),
-            ("PUT", "text/html", false),
-            ("GET", "text/plain", false),
-            ("GET", "text/html", false),
-        ];
-
-        for (method, header_value, expected) in cases {
-            let req = Request::builder()
-                .method(method)
-                .header("content-type", header_value)
-                .body(())
-                .unwrap();
-
             assert_eq!(
                 matcher.matches(None, &Context::default(), &req),
                 expected,
--- a/src/http/matcher/mod.rs
+++ b/src/http/matcher/mod.rs
@@ -5,6 +5,13 @@
 //! [`service::Matcher`]: crate::service::Matcher
 //! [`http::Request`]: crate::http::Request
 //! [`service::matcher` module]: crate::service::matcher
+use crate::{
+    http::Request,
+    service::{context::Extensions, matcher::IteratorMatcherExt, Context},
+    stream::matcher::SocketMatcher,
+};
+use std::fmt;
+use std::sync::Arc;
 
 mod method;
 #[doc(inline)]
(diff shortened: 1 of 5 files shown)


src/http/matcher/mod.rs
```diff
@@ -883,3 +605,131 @@ impl<State, Body> crate::service::Matcher<State, Request<Body>> for HttpMatcherK
...
+            .and(HttpMatcher::version(VersionMatcher::HTTP_2))
+            .and(HttpMatcher::domain("www.example.com"))
+            .negate();
+        let req = Request::builder()
+            .method("POST")
+            .version(Version::HTTP_2)
+            .uri("www.example.com")
+            .body(())
+            .unwrap();
+        assert!(!matcher.matches(None, &Context::default(), &req));
```
GlenDC, line 658: for these combo tests might be sane to verify that the logic is indeed correct (e.g. in this expect we expect to behave it like `!(A&B&C)`. Currently however this kind of test will pass even when it is `A&B&!C`.

This can be achieved by adding more request checks for the same combo matcher.

GlenDC (commented):
Looks already very good so far @dmackdev ! Well done.
I agree with your comment in https://github.com/plabayo/rama/issues/93#issuecomment-2027700683, if you feel motivated to pick that one up as well, feel free to take it as yours, and you can handle it as part of the same PR if you want. Because indeed, using request-agnostic booleans it will be indeed be a lot easier to test an entire truth table :)

All yours if you want. Do let it be known in that issue in the case.

Currently your code looks excellent for me and in fact you already solved the ticket for what is me concerned. Only remark is that currently you are just testing a single case of an entire truth table. I would prefer it that for each of these tests we did cover the entire truth table as expected. Especially for the complex queries.

src/http/matcher/mod.rs
```diff
@@ -880,6 +696,309 @@ impl<State, Body> crate::service::Matcher<State, Request<Body>> for HttpMatcherK
...
+                matcher,
+                req
+            );
+        }
+    }
+
+    #[test]
+    fn test_matcher_negation_and_header_contains() {
+        let matcher = HttpMatcher::method_get().negate().and_header_contains(
+            "content-type".parse().unwrap(),
```
GlenDC, line 977: You can use `HeaderName/HeaderValue`::from_static for these cases, just make sure to keep the names lower case as otherwise you get a panic, but that's okay for your cases.

GlenDC:
Your PR looks good to me @dmackdev , thank you for all your hard work.

Do you still plan to add some abstract truth table tests with hardcoded bools. Given your current test coverage I am honestly fine without, but might be the cherry on the cake if you do still go for it. Either way fine for me.

dmackdev:
Thank you @GlenDC , and thank you for the assistance along the way.

I was planning on amending the tests I added to use the hardcoded bool custom matchers - are you suggesting to add more tests instead of amending the ones I added?

I can make the changes this evening.

src/http/matcher/mod.rs
```diff
@@ -880,6 +696,504 @@ impl<State, Body> crate::service::Matcher<State, Request<Body>> for HttpMatcherK
...
+            _ctx: &Context<()>,
+            _req: &Request<()>,
+        ) -> bool {
+            self.0
+        }
+    }
+
+    #[test]
+    fn test_matcher_ands_combination() {
+        let cases = [
```
GlenDC: You can add `itertools` as a dev dependency.
This way you can just do:

```rust
use itertools::Itertools;

#[test]
fn test_matcher_ands_combination() {
    for (a, b, c) in [true, false].permutations(3) {
        let expected = a && b && c;
        let a = BooleanMatcher(a);
        let b = BooleanMatcher(b);
        let c = BooleanMatcher(c);

        let matcher = a.and(b).and(c);
        let req = Request::builder().body(()).unwrap();
        assert_eq!(
            matcher.matches(None, &Context::default(), &req),
            expected,
            "({:#?}).matches({:#?})",
            matcher,
            req
        );
    }
}
```

Seems a lot less trouble, and easier to read/maintain?
If you agree you can apply this for all these tests.

src/http/matcher/mod.rs
```diff
@@ -880,6 +696,517 @@ impl<State, Body> crate::service::Matcher<State, Request<Body>> for HttpMatcherK
...
+                expected,
+                "({:#?}).matches({:#?})",
+                matcher,
+                req
+            );
+        }
+    }
+
+    #[test]
+    fn test_matcher_or_and_or_and_negation() {
```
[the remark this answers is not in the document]
GlenDC: I would just apply the same technique I propose in my other comment. This way you can easily test it without having to write the cases yourself :) And easier to reason about then these manual http cases. So I would suggest to replace them, but only if you apply a technique such as the ones approached, because manual test cases won't make it that much easier.

GlenDC (changes requested):
Once the last comments are applied, let me know if it is finished from your POV. Because then I'll just merge it in also next time I review, in case I approve at that point.

dmackdev:
@GlenDC all done from my side.

GlenDC (approved):
Great job. Thank you for this. Merging it in.
Take care and feel free to take on another open issue if you can and wish.
Either way thanks for your devotion and work.

Issue #53, opened 2024-02-17
add integration tests for all examples


GlenDC:
Currently the examples found in the examples dir are already ensured to remain valid Rust programs in the sense that they can compile. However beyond manually running these they aren't automatically validated in the sense that they produce expected results.

Goal of this ticket is to have the examples tested.

For this we can add a single integration test per example for which we can write some utility code. In a sense this would allow all examples to be tested by doing the following steps:

1. build the example using <https://docs.rs/escargot/latest/escargot/struct.CargoBuild.html#method.example>
2. run the example for as long as the test is active
3. use our own http client to call its endpoints
4. validate the response by checking the response headers and/or inspecting the payload

Validation should remain as basic as possible while validating that the response clearly contains the generated content.

GlenDC:
Allowing others to pick this up if you want it let it be known here. Otherwise I'll pick it up myself one of these weeks. Once I do that I'll assign myself. But only if no one is working on it by then.

GlenDC:
A much simpler to the port issue might just be to give every example its own unique port. E.g. we can start at `40000` and bump the number up every time. It's a port range that has a low change of conflict (essentially a zero chance for our CI systems, but even on developer devices should be low enough).

This way we do not need to do anything special, and do also not need to listen to `stdout` somehow to get the port. We can just hardcode in our tests the port and move on.

GlenDC:
This story is being solved by @Hmikihiro

GlenDC:
@Hmikihiro pushed it most of the way forward, I am now going to take over upon his request to finish it off.

PR #134, opened 2024-04-01
add test for http_listener_hello and http_conn_state

Hmikihiro:
Hi, @GlenDC, I want to challenge #53  issue.
As an experiment, a test was created against `http_listener_hello.rs`.
I would like to know your thought about how to call example.
I'm concerned that the tests take a lot of times for examples be compiled.


tests/example_http_conn_state.rs
```diff
@@ -0,0 +1,32 @@
...
+            r##"
+            <html>
+                <head>
+                    <title>Rama — Http Conn State</title>
+                </head>
+                <body>
+                    <h1>Metrics</h1>
+                    <p>Alive: {is_alive}
+                    <p>Connection <code>{conn_index}</code> of <code>{conn_count}</code></p>
+                    <p>Request Count: <code>{request_count}</code></p>
```
GlenDC, line 22: I would honestly not assert on the entire response, as that is overly restrictive (E.g. we might want to change some tiny bits). I think you can just check a small part (e.g. contains `Request Count: 1`, that still tests business logic, but not too much.

tests/test_server/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+    }
+}
+
+pub fn run_example_server(example_name: &str, port_number: usize) -> ExampleServer {
+    let temp = assert_fs::TempDir::new().unwrap();
+    ExampleServer(
+        escargot::CargoBuild::new()
+            .example(example_name)
+            .manifest_path("Cargo.toml")
+            .target_dir(temp.path())
```
GlenDC, line 21: this is not needed, right? As far as I can see from the documentation it is not a required method for this builder to succeed, and I assume that it will just use your regular target dir, which is ok?

tests/test_server/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+            .run()
+            .unwrap()
+            .command()
+            .arg(port_number.to_string())
```
GlenDC, line 25: As I think it's easier and just as fine to just hardcode the port numbers as unique ones, starting with the first example with port `40_000`, there will be no need to pass in a port number.

tests/example_http_listener_hello.rs
```diff
@@ -0,0 +1,19 @@
...
+    let get = reqwest::get("http://127.0.0.1:40001/path").await?;
+    assert_eq!(get.text().await?, r#"{"method":"GET","path":"/path"}"#);
+
+    let client = reqwest::Client::new();
+    let post = client
+        .post("http://127.0.0.1:40001/")
+        .body("body")
+        .send()
+        .await?;
+    assert_eq!(post.text().await?, r#"{"method":"POST","path":"/"}"#);
```
GlenDC, line 16: same here, I would parse it as `json`, and then perhaps check on 1 or 2 properties, but certainly no literal string

GlenDC (changes requested):
Hi @Hmikihiro. Thank you for being a first time contributor :) Welcome!

If you need guidance or help, or have feedback then GitHub is fine for me, so feel free to continue to communicate like this. If you are ever stuck or want more interactive mentoring as part of your contribution, feel free to join our discord and start a chat there. Either way welcome!

I hope this review helps to unblock you a bit. All in all it does look pretty good, certainly I like the direction you are taking this here.

Regarding your concern of compile time, how much are we talking about? First of all I do think it doesn't help that you compile to a  tmp target dir which will mean you need to recompile probably everything each time. Keeping it all in same target dir (the default) will probably help a lot already.

And of course, I don't expect these tests to be as fast as regular tests, given we do real I/O and also have to compile and start actual (child) processes. So all in all, can take longer then a usual test, but of course not extremely long either.

Let me know how it is once you applied all changes. If you are still concerned please show here the numbers. I might run it myself then and see if iti s acceptable. This is either way mostly a test suite to be run in GitHub action. I do not expect developers to run the example tests.

Hmikihiro:
Thank you feed back.
I try to incorporate your feed back.

Sorry, I failed to run the test on removing the `temp_dir` from escargot.

Hmikihiro:
I want to rethink about the length of compile time after writing all the tests.

tests/example_http_key_value_store.rs
```diff
@@ -0,0 +1,44 @@
...
+use http::StatusCode;
+use rama::{error::BoxError, http::Request};
+
+use crate::test_server::recive_as_string;
+
+#[tokio::test]
+async fn test_http_key_value_store() -> Result<(), BoxError> {
+    let _example = test_server::run_example_server("http_key_value_store");
+
+    // because the example couldn't recive request yet.
```
GlenDC, line 12: Probably not a bad idea to just make a high level test client under `test_client.rs` of this crate which gives you easy methods to get `json/str` data, and which will also have graceful retries. E.g. if a failure to retry with a backoff policy. The latter could be made using a new middleware that could be added under `rama::service::layers`. Called `RetryService` / `RetryLayer`. Will look very much like the `Limit` stuff, but then for retrying. Also can be used using a policy,  so that for example exponential backoff is possible here as well, etc.

I get that this is asking a lot though, on top of the hard work you're putting with this request. So feel free to skip it for now and I can revisit that later. I'm only saying it because right now you are adding methods, e.g. `receive_as_string` and also small delays etc... That's a lot of sprinkled choices that in reality you can better solve by providing a small high level client for testing purposes.

But again, feel free to skip. Or come back to me if you do want to do it but need some guidance. We can also have it as a separate PR and keep this one scoped , a lot of possibilities. You can let me know.

tests/example_http_prometheus.rs
```diff
@@ -0,0 +1,45 @@
...
+}
+
+async fn test_metrics() -> Result<(), BoxError>{
+    let request = Request::builder()
+        .method("GET")
+        .uri(format!("{}{}",URL , "metrics"))
+        .body(String::new())
+        .unwrap();
+    let (_, res_str) = recive_as_string(request.clone()).await?;
+    assert_eq!(res_str,"# HELP example_counter example counter\n# TYPE example_counter counter\nexample_counter 2\n");
```
GlenDC, line 42: These tests are still to detailed. We are not concerned in these tests about formats, but more that the examples (1) run and (2) do more or less what we expect. So for example testing here that the str contains `example_counter 2` is good enough

tests/example_http_conn_state.rs
```diff
@@ -0,0 +1,41 @@
...
+    let _example = test_server::run_example_server("http_conn_state");
+
+    let get_request = Request::builder()
+        .method("GET")
+        .uri("http://127.0.0.1:40000/")
+        .body(String::new())
+        .unwrap();
+
+    let (_, res_str) = recive_as_string(get_request).await?;
+    let head = r##"
```
GlenDC, line 18: As I said in previous comment, these tests are too detailed. Just testing a small port of it (e.g. contains `Connection ...` is good enough).

GlenDC:
FYI you can run `just qa` (see CONTRIBUTIONS.md) to run most of these tests. Should also fix the formatting for you. As currently your tests do not get run because the formatting test (a pre-requirement) fails already.

> Sorry, I failed to run the test on removing the temp_dir from escargot. For the reason, I don't remove temp_dir yet.

Can you show the output and reasons why this fails? Then I can perhaps guide you in the direction, change my opinion or a combination of the two.

Hmikihiro:
When I replace from `temp.path()` to "/" or remove `target_dir(temp.path())`,  escargot compiles tests in `/debug`.

But, test programs occurred this error.
```
---- test_http_conn_state stdout ----
thread 'test_http_conn_state' panicked at tests/test_server/mod.rs:31:14:
called `Result::unwrap()` on an `Err` value: CargoError { kind: CommandFailed, context: Some("error: failed to create directory `/debug`\n\nCaused by:\n  Read-only file system (os error 30)\n"), cause: None }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

tests/example_http_key_value_store.rs
GlenDC: Added the stories:

* https://github.com/plabayo/rama/issues/138
* https://github.com/plabayo/rama/issues/137

Do not feel obliged to pick these up, or do it as part of this PR. It can be done by someone else, and it can be also done on another time. All up to you, either way fine.

Hmikihiro:
Thank you for the advice.
I've been able to reduce the compile time and use regular target dir.

tests/example_http_conn_state.rs
Hmikihiro: In my environment, the test returns 1 or 2 after `Connection`. I am attempting to check only that it contains the string `Connection`.

tests/example_http_key_value_store.rs
Hmikihiro: Thank you for your concern.
It is difficult for me to create high level client. I would skip it in this PR.

GlenDC: That is fair enough. You are already doing a lot! Thx for that!

tests/example_http_conn_state.rs
GlenDC: That is good enough for this PR, agreed. Could you put a comment though with “TODO: Investigate why this counter is not consistent 1”

GlenDC:
@Hmikihiro you are doing an awesome job, thank you very much, keep it up. Can you please close the comments on this PR once you resolved them, this way it’s clear to me which ones are still unresolved.

GlenDC:
@Hmikihiro you can merge main into your branch. Added the high level logic and retry support. With that branch merged you can add a `client.rs` in your root test folder with the following content (feel free to adapt where required):

```rust
use rama::{
    http::{
        client::HttpClient,
        layer::{
            compression::CompressionLayer,
            decompression::DecompressionLayer,
            retry::{ManagedPolicy, RetryLayer},
            trace::TraceLayer,
        },
        IntoResponse,
    },
    service::{util::backoff::ExponentialBackoff, BoxService, ServiceBuilder},
};

type HttpClientError = crate::error::Error;
type HttpClient<S> = BoxService<S, Request, Response, HttpClientError>;

fn client<S: Send + Sync + 'static>() -> HttpClient<S> {
    ServiceBuilder::new()
        .map_result(map_internal_client_error)
        .layer(TraceLayer::new_for_http())
        .layer(CompressionLayer::new())
        .layer(DecompressionLayer::new())
        .layer(RetryLayer::new(
            ManagedPolicy::default().with_backoff(ExponentialBackoff::default()),
        ))
        .service_fn(HttpClient::new())
        .boxed()
}
```

It's possible that there are warnings and errors in the code. You can see the original code in this style in https://github.com/plabayo/rama/blob/df54cd517a45dc267b5d0eb5d6e548347e295610/src/http/client/ext.rs#L693. You can also see a newly added example that shows how easy such a high level client API is to use: https://github.com/plabayo/rama/blob/df54cd517a45dc267b5d0eb5d6e548347e295610/examples/http_high_level_client.rs

Should help you to make your tests.

This way code in your PR such as

```rust
let get_request = Request::builder()
    .method("GET")
    .uri("http://127.0.0.1:40000/")
    .body(String::new())
    .unwrap();

let (_, res_str) = recive_as_string(get_request).await?;
```

should instead be:

```rust
let res_str = ::client().
    .get("http://127.0.0.1:40000/")
    .send()
    .await
    .unwrap()
    .try_into_string()
    .await
    .unwrap();
```

Let me know if it's clear. If there is guidance or assistance in any form needed from my part, do also let me know.
Take care.

Hmikihiro:
I have no ability for proxy and TLS connections.
I would appreciate it if you could take over from my branch.

GlenDC:
Sure thing. Thanks for letting me know. I'll take over and finish it today. Once again thank you very much for all your hard work. Soon you'll have your well deserved contributor badge.

My apologies for the sudden rush. But I have my last big sprint for this release and these tests will help with my time preservation and sanity.

GlenDC:
Merged this branch into a feature branch I own. Will take it from there. Thanks again @Hmikihiro ! You can follow https://github.com/plabayo/rama/pull/177 to see how that goes.

PR #136, opened 2024-04-02
feat-68/add form support

devanbenz:
- Adds support for Form within requests

### DIFF
--- a/examples/http_form.rs
+++ b/examples/http_form.rs
@@ -1,4 +1,4 @@
-//! An example to show how to create a minimal health check server,
+//! An example to show how to create a minimal server that accepts form data for a request.
 //! using the [`HttpServer`] and [`Executor`] from Rama.
 //!
 //! [`HttpServer`]: crate::http::server::HttpServer
@@ -9,7 +9,7 @@
 //! # Run the example
 //!
 //! ```sh
-//! cargo run --example http_health_check
+//! cargo run --example http_form
 //! ```
 //!
 //! # Expected output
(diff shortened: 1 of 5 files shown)


src/http/service/web/endpoint/extract/body.rs
```diff
@@ -131,6 +131,29 @@ where
     }
 }
 
+pub use crate::http::response::Form;
```
devanbenz, line 134: @GlenDC for some reason I'm seeing the following error with this module:

```
error: unreachable `pub` item
   --> src/http/service/web/endpoint/extract/body.rs:134:9
    |
134 | pub use crate::http::response::Form;
    | ---     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | |
    | help: consider restricting its visibility: `pub(crate)`
    |
    = help: or consider exporting it for use by other crates
```

Not entirely sure why.

devanbenz: Disregard I fixed it :)

src/http/response/form.rs
```diff
@@ -0,0 +1,86 @@
+use crate::http::dep::http::header::CONTENT_TYPE;
+use crate::http::dep::http::StatusCode;
+use crate::http::response::{IntoResponse, Response};
```
GlenDC, line 3: also add an import for `use crate::http::dep::mime;` please.

I know they are the same for now, but I try to use my own `dep` re-exports, so that if I ever need to I can easily swap it out with something else, without having to change import paths everywhere.

src/http/response/form.rs
```diff
@@ -0,0 +1,86 @@
...
+    T: Serialize,
+{
+    fn into_response(self) -> Response {
+        match serde_urlencoded::to_string(&self.0) {
+            Ok(body) => (
+                [(CONTENT_TYPE, mime::APPLICATION_WWW_FORM_URLENCODED.as_ref())],
+                body,
+            )
+                .into_response(),
+            Err(err) => (StatusCode::INTERNAL_SERVER_ERROR, err.to_string()).into_response(),
```
GlenDC, line 83: In the `FromRequest` part you only return a `StatusCode`. but here you also return the error? Might be better to just return the status code here, and trace the error, using `tracing::err`, and same in `FromRequest`? Or have I been doing that consistency myself for the other wrapper types as well? if so feel free to also align those while you are at it, or leave it and I'll do it later.

GlenDC (changes requested):
Looks good, thank you very much for this great contribution.
I have added two remarks, once these are resolved I am fully onboard to merge this pull request in. Thanks once again!

examples/http_form.rs
```diff
@@ -0,0 +1,34 @@
...
+//!
+//! The server will start and listen on `:8080`. You can use `curl` to check if the server is running:
+//!
+//! ```sh
+//! curl -v http://127.0.0.1:8080
+//! ```
+//!
+//! You should see a response with `HTTP/1.1 200 OK` and an empty body.
+
+use rama::{http::server::HttpServer, rt::Executor, service::service_fn};
```
GlenDC, line 25: Is this an example you are still planning to make? Because not sure if even needed. Either way for now it's just a copy paste of the health server example, so perhaps just delete this file? It's the last blocker for this PR To be able to be merged.

GlenDC (changes requested):
Except the odd example file I am okay with this being merged in.
Thank you very much for your work and devotion!

examples/http_form.rs
devanbenz: Just a placeholder - not quite finished yet! :)

GlenDC: Sure thing! not to worry. No pressure. I'll just leave you to it then. Take your time.
I suggest you just leave a comment here once you consider it final, and then we can take it from there to see if it can be merged in or not. Thanks for your dedication and work. Take care.

GlenDC (approved):
Thank you for the contribution, looks good!

PR #140, opened 2024-04-05
add stream custom matcher and combinator methods

dmackdev:
Closes https://github.com/plabayo/rama/issues/92.
Closes https://github.com/plabayo/rama/issues/93.


dmackdev:
@GlenDC The remaining step is to add tests. They would be a duplicate of the `HttpMatcher` tests only for `SocketMatcher`. Any ideas on how we could prevent duplication - a macro perhaps?

GlenDC:
> @GlenDC The remaining step is to add tests. They would be a duplicate of the `HttpMatcher` tests only for `SocketMatcher`. Any ideas on how we could prevent duplication - a macro perhaps?

Your PR looks good. I'm okay with what can be seen as duplication here though. No need to couple things that aren't necessary.

GlenDC (approved):

PR #147, opened 2024-04-14
Begin docummenting Rama's eventual squid-cache support

AlvinKuruvilla:
This commit adds a section to the book and begins the long process of creating summary documentation(options, and examples) for every directive on the squid cache website

### DIFF
--- a/docs/book/src/configuration/intro.md
+++ b/docs/book/src/configuration/intro.md
@@ -1,0 +1,4 @@
+Rama's configuration system is still in the works but this section aims to be a living document of Rama's configuration system. Specifically, this section showcases all the configuration options from squid-config that we do and do not support and provide examples where possible. As a rule of thumb, Rama aims to support the latest squid-config version (at the time of writing that is v7).
+
+During this section we will discuss:
+- [Squid Directives](squid-directives.md)
--- a/docs/book/src/configuration/squid-directives.md
+++ b/docs/book/src/configuration/squid-directives.md
@@ -1,0 +1,1826 @@
+# Squid Directives
+As previously mentioned Rama aims to be compatible with v7 of Squid Cache. Any items that are striked through are not available in v7. 
+<details>
+  <summary>accept_filter</summary>
+  Examples:
+
+    accept_filter httpready
+
+    accept_filter data
+</details>
+<details>
+  <summary>access_log</summary>
+  General Format:
+
+    access_log <module>:<place> [option ...]
+</details>
+<details>
+  <summary>acl</summary>
+  General Format:
+    
+    acl aclname acltype argument ...
+	acl aclname acltype "file" ...
+  
+  Examples:
+
+    acl localnet src 0.0.0.1-0.255.255.255	# RFC 1122 "this" network (LAN)
+    acl localnet src 10.0.0.0/8		# RFC 1918 local private network (LAN)
+    acl localnet src 100.64.0.0/10		# RFC 6598 shared address space (CGN)
+    acl localnet src 169.254.0.0/16 	# RFC 3927 link-local (directly plugged) machines
+    acl localnet src 172.16.0.0/12		# RFC 1918 local private network (LAN)
+    acl localnet src 192.168.0.0/16		# RFC 1918 local private network (LAN)
+    acl localnet src fc00::/7       	# RFC 4193 local private network range
+    acl localnet src fe80::/10      	# RFC 4291 link-local (directly plugged) machines
+    acl SSL_ports port 443
+    acl Safe_ports port 80		# http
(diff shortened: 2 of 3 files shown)


docs/book/src/configuration/intro.md
```diff
@@ -0,0 +1,4 @@
+Rama's configuration system is still in the works but this section aims to be a living document of Rama's configuration system. Specifically, this section showcases all the configuration options from squid-config that we do and do not support and provide examples where possible. As a rule of thumb, Rama aims to support the latest squid-config version (at the time of writing that is v7).
```
GlenDC, line 1: FYI squid-config is an inspiration, not something we want to be compatible with. It's a very common technology used by many, so clearly they did things right with that config. However it's a very complex config spec due to how much it supports and does. This is honestly not something I feel comfortable taking on in the project to be compatible, especially as the kind of project is not even that similar.

So for now we can look, besides how the config works and is defined, also what kind of rules do we want for rama. And initially I would already be happy if we can let the user define what ports to bind though, rules to control certain middleware (e.g. firewall, ratelimit, and other such rules).

But over time what we have and what they have will probably diverge more and more. And pretty certain that even at the start it might already look a bit different for the minimal stuff that is supported at that point.

GlenDC: That all said. I do appreciate your work, initiative and input here. It makes me think that perhaps this issue is a bit premature. Or perhaps it is not.

docs/book/src/configuration/squid-directives.md
```diff
@@ -0,0 +1,1826 @@
+# Squid Directives
```
GlenDC, line 1: These are a lot of rules and "directives". Also I would not mention squid at all in these documents as it's just an inspiration not really meant as a specific spec to follow.

That project most likely also didn't start with a giant list of directives, but grew over time I imagine.

I would much rather want to just start to come up with a small list of rules that proofs the value of such a config, and that can already be supported by a v0.2 or v0.3 of rama.

Not sure if I make much sense though.

GlenDC:
```
acl SSL_ports port 443

acl Safe_ports port 80		# http
acl Safe_ports port 443		# https

acl CONNECT method CONNECT

http_access deny !Safe_ports
http_access deny CONNECT !SSL_ports

http_port 8080
```

That was the example I gave but I honestly do not have experience myself with Squid.
So as I look at the example I am not sure the value it gives or how I Would rather see if not this.

Might be something we need to explore better.

GlenDC:
Left some comment also on the original issue. I think we need to step back a bit and first figure out some more fundamental questions prior to losing ourselves in details. This PR overwhelmed me tbh a bit, but that's good. It woke me up in aspect of this issue. A wake up call to provide clarity here for the sake of both of us.

AlvinKuruvilla:
Hi @GlenDC, thanks for the feedback. I admit, in hindsight, it probably wasn't the smartest idea on my part to try and tackle the spec in one shot like I did 😅. I like the idea of having a minimum configuration to grow organically. I also read your comments on the issue regarding TOML, and I agree that using TOML may be a more straightforward approach. I can sketch out the config in something like YAML or TOML unless you would rather I tackle another aspect of the issue. 
I will also address your other comments on the PR as well :)

GlenDC:
As I said, it's a big issue, despite the small scope that the first implementation will have.
The bigness comes mostly from all the explority work that is required for this issue before one can even start implementing.

While my own experience is inclined to rather go for something like `toml`, I do still think it's worth exploring what the squid-cache config language (if we can call it that) would give us. What are the pros and cons of it vs the toml.

It might be easier to do this exploritory work with some examples in mind:

Project:

- service A: health check service
- service B: http proxy with https support
- service C: socks5 opaque proxy (sticks to the transport layer)

Crucial is to also check some of the middleware already in the project implemented currently, on both the transport and application layers, and see how one could configure these? Are some opt-out rather then opt-in? Also perhaps come up with a custom http middleware that does something, and see how one would enable that as well.

This is still a rather basic example, but seeing this configured in the two approaches might tell us a lot already. WDYT?

GlenDC:
And of course, as this is not really a "good first issue", or "easy", feel free to pull out at any point after having said you want to do this. There's no shame in that either. It's a pretty undefined issue for which I have no answers from the top of my head either.

GlenDC:
I'm going to close this PR @AlvinKuruvilla as I think this was a bit premature, despite all the wonderful effort on your end which I do appreciate. This issue requires first some R&D work as discussed before, where no code changes are required for.  You might want to pick up another issue if you want to contribute to Rama, for now this is going to be a bit on hold. The README now also contains a bit of a roadmap overview.

PR #151, opened 2024-04-20
update deps: replaced futures crate with futures_lite

### LINKED ISSUE #150: replace futures crate dep with futures_lite
docs: https://docs.rs/futures-lite/latest/futures_lite/index.html

Remark: `join_all` is not there, but given we only join two futures in the usage found within rama, we can just replace our `join_all` usage with `futures::future::zip`.

dalprahcd:
Should close #150 by replacing `futures` and `futures-core` crate with `futures_lite` crate.

Could not replace `futures_util` crate, as I did not find an easy way to replace `futures_util::TryStream`.
The farthest I got was replacing
```
where
    S: TryStream,
    S::Ok: Into<Bytes>,
    S::Error: Into<BoxError>,
```
with 
```
where
  S: Stream<Item = Result<Bytes, BoxError>>
```

But that created more problems than solutions. I am open to guidance on this task.

Also, not sure how to fix clippy warning:

```
warning: unused `std::result::Result` in tuple element 0 that must be used
   --> src/stream/layer/tracker/bytes.rs:403:9
    |
403 |         futures_lite::future::zip(task_1, task_2).await;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this `Result` may be an `Err` variant, which should be handled
    = note: `#[warn(unused_must_use)]` on by default
```

Is ignoring the result the solution to go?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,8 @@ bytes = "1"
 argh = "0.1"
 crossterm = "0.27"
 flate2 = "1.0"
-futures = "0.3.29"
+futures-lite = "2.3.0"
 futures-core = "0.3"
-futures-util = { version = "0.3", default-features = false }
 h2 = "0.4"
 headers = "0.4"
 http = "1"
@@ -101,9 +100,8 @@ async-compression = { workspace = true, features = ["tokio", "brotli", "zlib", "
 base64 = { workspace = true }
 bitflags = { workspace = true }
 bytes = { workspace = true }
-futures = { workspace = true }
 futures-core = { workspace = true }
-futures-util = { workspace = true, features = ["alloc"] }
+futures-lite = { workspace = true }
 h2 = { workspace = true }
 headers = { workspace = true }
 http = { workspace = true }
--- a/src/http/layer/compression/mod.rs
+++ b/src/http/layer/compression/mod.rs
@@ -5,17 +5,18 @@
 //! Example showing how to respond with the compressed contents of a file.
 //!
 //! ```rust
-//! use bytes::{Bytes, BytesMut};
-//! use rama::http::{Body, Request, Response, header::ACCEPT_ENCODING};
+//! use bytes::Bytes;
+//! use futures_lite::stream::StreamExt;
+//! use rama::error::BoxError;
 //! use rama::http::dep::http_body::Frame;
-//! use rama::http::dep::http_body_util::{BodyExt, StreamBody, combinators::BoxBody as InnerBoxBody};
+//! use rama::http::dep::http_body_util::{BodyExt , StreamBody};
+//! use rama::http::dep::http_body_util::combinators::BoxBody as InnerBoxBody;
+//! use rama::http::layer::compression::CompressionLayer;
+//! use rama::http::{Body, Request, Response, header::ACCEPT_ENCODING};
+//! use rama::service::{Context, Service, ServiceBuilder};
 //! use std::convert::Infallible;
 //! use tokio::fs::{self, File};
 //! use tokio_util::io::ReaderStream;
-//! use rama::service::{Context, Service, ServiceBuilder, service_fn};
-//! use rama::error::BoxError;
-//! use rama::http::layer::compression::CompressionLayer;
-//! use futures_util::TryStreamExt;
 //!
 //! type BoxBody = InnerBoxBody<Bytes, std::io::Error>;
 //!
@@ -27,11 +28,14 @@
 //!     // Convert the file into a `Stream` of `Bytes`.
 //!     let stream = ReaderStream::new(file);
 //!     // Convert the stream into a stream of data `Frame`s.
-//!     let stream = stream.map_ok(Frame::data);
+//!     let stream = stream.map(|res| match res {
+//!         Ok(v) => Ok(Frame::data(v)),
+//!         Err(e) => Err(e),
+//!     });
 //!     // Convert the `Stream` into a `Body`.
 //!     let body = StreamBody::new(stream);
 //!     // Erase the type because its very hard to name in the function signature.
-//!     let body = body.boxed();
+//!     let body = BodyExt::boxed(body);
 //!     // Create response.
 //!     Ok(Response::new(body))
 //! }
@@ -406,7 +410,7 @@ mod tests {
         let compressed_with_level = {
             use async_compression::tokio::bufread::BrotliEncoder;
 
-            let stream = Box::pin(futures::stream::once(async move {
+            let stream = Box::pin(futures_lite::stream::once({
                 Ok::<_, std::io::Error>(DATA.as_bytes())
             }));
             let reader = StreamReader::new(stream);
--- a/src/http/server/hyper_conn.rs
+++ b/src/http/server/hyper_conn.rs
@@ -5,7 +5,6 @@ use crate::service::Service;
 use crate::service::{Context, HyperService};
 use crate::stream::Stream;
 use crate::tcp::utils::is_connection_error;
-use futures::FutureExt;
 use hyper::server::conn::http1::Builder as Http1Builder;
 use hyper::server::conn::http2::Builder as Http2Builder;
 use hyper_util::{rt::TokioIo, server::conn::auto::Builder as AutoBuilder};
@@ -51,7 +50,7 @@ impl HyperConnServer for Http1Builder {
         let mut conn = pin!(self.serve_connection(stream, service).with_upgrades());
 
         if let Some(guard) = guard {
-            let mut cancelled_fut = pin!(guard.cancelled().fuse());
+            let mut cancelled_fut = pin!(guard.cancelled());
 
             loop {
                 select! {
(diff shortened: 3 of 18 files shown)


src/stream/layer/tracker/bytes.rs
```diff
@@ -400,6 +400,6 @@ mod tests {
         assert_eq!(handle.read(), 9);
         assert_eq!(handle.written(), 9);
 
-        futures::future::join_all(vec![task_1, task_2]).await;
+        futures_lite::future::zip(task_1, task_2).await;
```
GlenDC, line 403: You can do

```
let (t1, t2) = futures_lite...
t1.unwrap();
t2.unwrap();
```

This might fail as it was previously ignored already as well. Do let me know in case that happens.

Cargo.toml
```diff
@@ -22,8 +22,7 @@ bytes = "1"
 argh = "0.1"
 crossterm = "0.27"
 flate2 = "1.0"
-futures = "0.3.29"
-futures-core = "0.3"
```
GlenDC, line 26: You should not get rid of `futures-core`, as that is where you'll find stuff like the `TryStream` trait. And as you'll see it's also still a dependency of futures-lite: https://crates.io/crates/futures-lite/2.3.0/dependencies

Cargo.toml
```diff
@@ -22,8 +22,7 @@ bytes = "1"
...
+futures-lite = "2.3.0"
 futures-util = { version = "0.3", default-features = false }
```
GlenDC, line 26: We probably do want to get rid of futures-util. You'll notice that we have places like `src/service/layer/trace_err.rs` that make use of it for stuff like `::map`, but honestly you can just get rid of that, and instead make it a regular async method impl (using the `async` syntax and await the future so yoiu can just deal with the result normally.

That's just one example. But just so you know, you can use what's in `futures_lite/futures_core` if there's a replacement, but otherwise it might just mean you need to just do it differently, as shown in this example

GlenDC:
> Could not replace futures_util crate, as I did not find an easy way to replace futures_util::TryStream

`TyStream` is from futures-core. So you can get rid of futures_util, just keep using futures_core. Eg. `TryStream` is just re-exported by `futures-util` as can be seen here: <https://github.com/rust-lang/futures-rs/blob/bb63c376ca2d3f277f7bfb2b178c913faa00b947/futures-util/src/stream/mod.rs#L14>

dalprahcd:
Thanks, got it most of it working now. Just need to update the example for ` http::layer::compression`, I do not fully understand how this layers interact with each other, what's your opinion on the best way to fix it?

```
//! async fn handle(req: Request) -> Result<Response<BoxBody>, Infallible> {
//!     // Open the file.
//!     let file = File::open("Cargo.toml").await.expect("file missing");
//!     // Convert the file into a `Stream` of `Bytes`.
//!     let stream = ReaderStream::new(file);
//!     // Convert the stream into a stream of data `Frame`s.
//!     let stream = stream.map(Frame::data);
//!     // Convert the `Stream` into a `Body`.
//!     let body = StreamBody::new(stream);
//!     // Erase the type because its very hard to name in the function signature.
//!     let body = body.boxed();
//!     // Create response.
//!     Ok(Response::new(body))
//! }
```

src/service/builder.rs
```diff
@@ -307,9 +306,11 @@ mod test {
             ctx: Context<State>,
             req: Request,
         ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_ {
-            self.0
-                .serve(ctx, req)
-                .map_ok(|res| res.as_ref().to_uppercase())
+            let serve = self.0.serve(ctx, req);
```
GlenDC, line 309: This is valid, but I would prefer if you instead do the more rama-idiomatic:

```rust
impl<S, State, Request> Service<State, Request> for ToUpper<S>
where
    S: Service<State, Request>,
    S::Response: AsRef<str>,
{
    type Response = String;
    type Error = S::Error;

    async fn serve(
        &self,
        ctx: Context<State>,
        req: Request,
        let res = self.0.serve(ctx, req).await;
        res.map(|msg| msg.as_ref().to_uppercase())
    }
}
```

src/service/layer/layer_fn.rs
```diff
@@ -75,7 +74,11 @@ mod tests {
                 ctx: Context<State>,
                 req: Request,
             ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_ {
-                self.0.serve(ctx, req).map_ok(|res| res.to_uppercase())
+                let serve = self.0.serve(ctx, req);
```
GlenDC, line 77: valid but you can just implement it as `async fn serve(...) -> Result<...> { ... }`, is more rama-idiomatic, and easier. Then you do not need the `async move`.

src/service/layer/trace_err.rs
```diff
@@ -69,7 +68,9 @@ where
         req: Request,
     ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_ {
         let level = self.level;
-        self.inner.serve(ctx, req).map(move |res| {
+        let serve = self.inner.serve(ctx, req);
```
GlenDC, line 71: valid but you can just implement it as `async fn serve(...) -> Result<...> { ... }`, is more rama-idiomatic, and easier. Then you do not need the `async move`.

GlenDC (changes requested):
You're almost there. Left some feedback already. Will also come back to your question regarding the compression part later today, when I find access to another developer machine.

dalprahcd:
Thanks for your comments, they are helping me a lot, but I have a question about your suggestions, they do not compile because neither `State`, `Request` and `Context` are `Send`, so the future returned by `serve` is not `Send`, then we can not satisfy the trait `Service`

``` rust
    /// Serve a response or error for the given request,
    /// using the given context.
    fn serve(
        &self,
        ctx: Context<S>,
        req: Request,
    ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_;
```

GlenDC:
> Thanks for your comments, they are helping me a lot, but I have a question about your suggestions, they do not compile because neither `State`, `Request` and `Context` are `Send`, so the future returned by `serve` is not `Send`, then we can not satisfy the trait `Service`
> 
> 
> 
> ``` rust
> 
>     /// Serve a response or error for the given request,
> 
>     /// using the given context.
> 
>     fn serve(
> 
>         &self,
> 
>         ctx: Context<S>,
> 
>         req: Request,
> 
>     ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_;
> 
> ```

You can check the rest of the codebase what the solution is. You need to specify the trait bounds (the list after 'where')

```
Request: Send + 'static,
State: Send + Sync + 'static,
```

src/http/layer/compression/mod.rs
```diff
@@ -27,7 +27,7 @@
 //!     // Convert the file into a `Stream` of `Bytes`.
 //!     let stream = ReaderStream::new(file);
 //!     // Convert the stream into a stream of data `Frame`s.
-//!     let stream = stream.map_ok(Frame::data);
+//!     let stream = stream.map(Frame::data);
```
GlenDC, line 30: This map will get a result. So you'll have to do smth like

```
map(|res| match res {
    Ok(v) => Ok(Frame::data(v)),
    Err(e) => Err(e),
})
```

GlenDC:
Let me know if after resolving and taking account my given feedback are blocked and need help/guidance. Can go more in depth where you want me to.

If you think the task is ready for an approval, do also let me know.

dalprahcd:
Hello @GlenDC the PR is ready for review, thanks very much for your help so far. I tried to clean up the example at `src/http/layer/compression/mod.rs` a bit as well as doing the required modifications.

src/http/server/hyper_conn.rs
```diff
@@ -5,7 +5,6 @@ use crate::service::Service;
 use crate::service::{Context, HyperService};
 use crate::stream::Stream;
 use crate::tcp::utils::is_connection_error;
-use futures::FutureExt;
```
GlenDC, line 8: Deleting this is incorrect, you no longer `::fuse()` which is going to be ok for the tests, but in production in can mean a panic.

Please use `futures_lite::stream::StreamExt`, so that you can re-add the `.fuse()` calls below @dalprahcd

GlenDC (changes requested):
Great job. One last change requested, once you resolved that last one, I'll merge this PR into the main branch. Thank you once again!

src/http/server/hyper_conn.rs
dalprahcd: Understood, but I may need help with this task. The `.fuse()` method is defined for `Streams`, but `guard.cancelled()` is a `Future`, so I am having a hard time using it.

GlenDC: Seems that they do not have such a fuse functionality for futures.
All good. I'll make something myself in a follow-up MR. This one is good to go then.

Thanks for the contribution!

GlenDC (approved):

Issue #154, opened 2024-04-25
revisit current root Error logic


GlenDC:
The current `rama::error` code was heavily inspired from Axum if I remember correctly.
However, it is a bit shortcoming and converting into or from it is not always as easy.

We might need to go a bit more complex here, while still keeping it essentially as easy as possible.
A re-study of anyhow and friends might help here.

GlenDC:
The first approach was to more or less follow the approach of anyhow, minus the backtracing. This can be seen in <https://github.com/plabayo/rama/pull/156>.

However... This is a mess, very invasive PR and was still pretty much as awkward, if not more...
It could have been see as a lot of wasted hours, except that I did learn from it:

- the awkward feeling comes from having an Error that is really not an `std::error::Error`.
   - this is fine for `anyhow` as its purpose is to be used as an easy to use high level ib
   - it is not fine for frameworks such as `rama` that operate on lower levels and higher levels alike,
      because as soon as you also need to interact with third party crates you might need to easily be seen as a regular std error (boxed or not)
     -  combine this with the fact that you have giant stacks of generics, like `rama` has, and you can get very crazy loco errors due to something as inocent like `From<E>`... Yihaa, Damn

In <https://github.com/plabayo/rama/pull/157> I'm going to implement tonight a new approach based on the original code, combined with the lessons learned from the mess of today:

- I really do not need an opaque `Error` type, and it does more harm then good, due to its awkwardness that comes from limitations because of auto trait impelmentations induced with `From<T> for T`...
   - This can be used to hide magic, or extra data like backtrace (e.g. what `anyhow` does)
   - we however do not need it
- Instead we can just work with `BoxError`, and also as extra give a `Result<T, BoxError>` and `Result::<T, BoxError>::Ok`
- We can then also provide `ErrorExt` with methods: `::context`, `::with_context`, `::with_backtrace`, `::backtrace`, and `::chain`
   - `::backtrace` will will work using the `backtrace` crate, and we should also support custom errors that can expose `backtrace`... Anyway, all without weird magic
- We can also provide `error!` macro that easily can work with `Into<Error>` expr, but also strings and template strings, easier to use, but all just exposed as `BoxError` in the result

The end result will be that everything will just be a `BoxError`, which is not awkward, with all utilities added around it, without having to have an awkward wrapper type that is both an error and not an error. And this way we improved the main situation without getting into the mess of earlier today.

Fin.

PR #156, opened 2024-04-26
upload mess (TODO: delete this mess branch)

GlenDC:
just keeping it for reference,
better to do a less invasive change,
for what we need this was a bad approach

going to take the lessons from here
and apply them in a less invasive manner


GlenDC:
This was in function of #154 , but the approach was too blunt.

Instead I think we can upgrade our current error module to more or less the same, but apply some extras...

The result will be dissapointing and less then hoped for...
but given the fact that we are sandwiched between the fact that From<T> for T is implemented and some third party crates depend on errors being std::error::Error.... Well... Damn :(

PR #177, opened 2024-05-04
add integration tests for all examples

### LINKED ISSUE #53: add integration tests for all examples

GlenDC:
Closes #53


GlenDC (commented):
All good. This can be merged.

GlenDC:
@Hmikihiro Thanks once again for all your hard work. You gave me the motivation to finish of this task <3 Thanks a lot! If you want to ever help more out on Rama in future do reach out. Take care and have a good one!

PR #219, opened 2024-05-27
refactor code a bit, improve network address handling and related improvements

### LINKED ISSUE #202: revisit proxy addresses
Currently these are always expected to be socket addresses in the actual proxy connect code.
The ProxyDB code keeps this more generic by having it as strings, which is neither good.

Instead we can perhaps have it as follows:

```rust
enum NetworkAuthority {
    SocketAddresses { addresses: Vec<Addresses> }
    Domain(String)
}

struct NetworkAddress {
     authority: NetworkAuthority,
     port: Option<u16>
}
```

Or something like this...

There might already be prior art about this in the rust eco system.
Ideally we have our entire codebase work around this, and easily also go from/to regular types (e.g. SocketAddr)...

This way it will all work the same.

It would also mean that we will validate our proxy addresses at parse/load time.

Let's do this after we merged in Hyper, as it's a huge distraction from that more important last remaining big piece.

### LINKED ISSUE #212: support x-forwarded-for header layer
and enable opt-in support for it in ip service

(https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)

### LINKED ISSUE #213: Support Via header
Contains proxy info: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

This info can also be found in forwarded header

### LINKED ISSUE #214: Support Forwarded Header
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded

### LINKED ISSUE #215: Support X-Forwarded-Proto header
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto

Can also be found in other headers such as Via and Forwarded

GlenDC:
- [ ] closes #215
- [ ] closes #214 
- [ ] closes #213 
- [ ] closes #212 
- [ ] closes #202

### DIFF
--- a/src/http/headers/common/forwarded.rs
+++ b/src/http/headers/common/forwarded.rs
@@ -1,0 +1,494 @@
+// //! This module implements parsing for the Forwarded header as defined by
+// //! [RFC 7239] and [RFC 7230].
+// //! 
+// //! [RFC 7239]: https://tools.ietf.org/html/rfc7239
+// //! [RFC 7230]: https://tools.ietf.org/html/rfc7230
+// //! 
+// //! # Example
+// //! 
+// //! TODO
+// //! 
+// //! # Fork
+// //! 
+// //! This code is forked from [`rust-forwarded-header-value`](https://crates.io/crates/forwarded-header-value):
+// //!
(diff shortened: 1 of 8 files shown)


GlenDC:
Replaced by other PRs.

PR #294, opened 2024-08-27
Remove const in HeapReader::new()

AlvinKuruvilla:
Rama is generating this error when building with the justfile or executing `cargo run`: `std::io::Cursor::<T>::new is not yet stable as a const fn`

To resolve this error remove the const from `HeapReader::new()`

### DIFF
--- a/src/net/stream/read.rs
+++ b/src/net/stream/read.rs
@@ -19,7 +19,7 @@ pin_project! {
 
 impl HeapReader {
     /// Creates a new `HeapReader` with the specified bytes data.
-    pub const fn new(data: Vec<u8>) -> Self {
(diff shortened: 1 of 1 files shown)


GlenDC:
Seems like it was only stabalized in 1.79.
The real solution however is to bump the MSRV up.

<https://releases.rs/docs/1.79.0/>

Going to just put it on 1.80 as rama 0.2.0 (non-alpha) will not release before 5 september,
after which we already will have even 1.81 so still giving people some room.

Thank you for raising this though!

GlenDC:
Thanks for your time and effort @AlvinKuruvilla ,
happy to see you around :)

PR #295, opened 2024-08-27
First cut of a new rama CLI 'probe' command

AlvinKuruvilla:
This commit introduces the first version of a new probe command to the rama CLI

It has two methods of input:
1. STDIN: pass a domain directly to the rama probe command like so:

`rama probe google.com`

2. FILE: pass `-l` to `rama probe` with a txt file containing all domains to probe:

`rama probe -l domains.txt`
where `domains.txt` is a file containing entries like:

`google.com`
`github.com`

### DIFF
--- a/rama-cli/src/cmd/mod.rs
+++ b/rama-cli/src/cmd/mod.rs
@@ -4,4 +4,5 @@ pub mod echo;
 pub mod fp;
 pub mod http;
 pub mod ip;
+pub mod probe;
 pub mod proxy;
--- a/rama-cli/src/cmd/probe/mod.rs
+++ b/rama-cli/src/cmd/probe/mod.rs
@@ -1,0 +1,232 @@
+//! rama probe service
+
+use std::{
+    fs::File,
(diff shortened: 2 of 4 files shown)


AlvinKuruvilla:
Hi @GlenDC, wanted to follow up on this PR. I noticed that when the command would run against a domain it would always give a 301 status code so I wanted to ask if you have any suggestions to resolve that 😃

GlenDC:
This is a very old PR . Lots of stuff changed.
We also have a probe command for some related things now:

```
target/debug/rama probe --help`
probe network services

Usage: rama probe [OPTIONS] <COMMAND>

Commands:
  tls   probe a server for its Tls capabilities
  tcp   probe a server for its Tcp capabilities
  help  Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose  enable debug logs for tracing (possible via RUST_LOG env as well)
  -h, --help     Print help
```

We could in future revisit this, but it will best be done from a new branch regardless.

PR #327, opened 2024-10-02
allow context state to be mutated

### LINKED ISSUE #316: allow context state to be mutated
At the time of writing this issue we still auto-wrap `State` used in `Context` automatically in an `Arc`.
There was good reason for it: as we expect context to be cloned in some fan services (e.g. tcp listener and http server),
as well as some middleware (e.g. retry middleware).

A user of `rama` has raised on Discord however a use case where `State`, given it is only used by 1 service at a time,
could be mutated and it would help their use case. RIght now that is only possible by putting such mutatable state as an `Extension` or by having a question `Request`, neither are very desirable for required custom state that you know is there and need to mutate. A non-considered option would be to use interior mutability but that would be a performance penalty with no real benefit.

As such not wrapping the state in `Arc` is probably the solution. For data that is heavy to clone and that is to be considered read-only, the users can still wrap those parts themselves as an `Arc`. Or even wrap the entire state in `Arc`, but it would be their choice.

GlenDC:
Closes #316


rama-http/src/layer/retry/mod.rs
```diff
@@ -260,7 +260,7 @@ mod test {
             retried: bool,
             service: &impl Service<State, Request, Response = Response, Error = E>,
         ) {
-            let state = ctx.state_clone();
+            let state = ctx.state().clone();
```
GlenDC, line 263:
```suggestion
            let state = ctx.state_clone();
```

rama-http/src/layer/retry/mod.rs
```diff
@@ -293,7 +293,7 @@ mod test {
...
```
GlenDC, line 296:
```suggestion
            let state = ctx.state_clone();
```

rama-macros/src/lib.rs
```diff
@@ -1,5 +1,9 @@
 //! Macros for [`rama`].
 //!
+//! There are no more macros for Rama. We used to have an `AsRef` one,
+//! but it is recommended to either not use a macro for that anymore,
+//! write one yourself or use a thirdparty crate such as `derive-more`.
```
GlenDC, line 5:
```suggestion
//! write one yourself or use a thirdparty crate such as `derive_more`.
```

rama-tcp/src/server/listener.rs
```diff
@@ -10,79 +10,99 @@ use std::{io, net::SocketAddr};
 use tokio::net::{TcpListener as TokioTcpListener, TcpStream, ToSocketAddrs};
 
 /// Builder for `TcpListener`.
-pub struct TcpListenerBuilder<S> {
+pub struct TcpListenerBuilder<S, T = ()> {
```
GlenDC, line 13: TODO: delete this state_transformer here

tests/integration/examples/example_tests/http_mitm_proxy.rs
```diff
@@ -50,6 +50,8 @@ async fn test_http_mitm_proxy() {
 
     let executor = Executor::default();
 
+    // TODO: rethink stateTransformer... it is putting silly bounds...
```
GlenDC, line 53:
```suggestion
```

tests/integration/examples/example_tests/http_mitm_proxy.rs
```diff
@@ -50,6 +50,8 @@ async fn test_http_mitm_proxy() {
...
+    // why did we introduce this in first place... Do we really need it?
```
GlenDC, line 54:
```suggestion
```

GlenDC (commented):
Looks good for most part, just couple of bits to fix still and then we can finally merge this hell of a refactor.

Issue #329, opened 2024-10-02
add "Deny All" Dns Resolver 


GlenDC:
A <https://ramaproxy.org/docs/rama/dns/trait.DnsResolver.html> implementation which returns an error for any incoming request. Can be named `DenyAllDns`.

parkma99:
Hi, I am from `[This Week in Rust](https://this-week-in-rust.org/)`, I would like to work it.

GlenDC:
Sure thing. How I see it is that it's basically an empty struct `DenyAllDns` which always returns an error, e.g. `DnsDeniedError` (can be made with `rama_utils::macros::error::static_str_error!`). And then implement the `DnsResolver` trait.

That's how I see it, but feel free to do it differently if you see another approach :)

Nice to meet you and thank you for your interest.

GlenDC:
Closed with https://github.com/plabayo/rama/pull/344 by @parkma99 
Thanks 🥳 !

PR #344, opened 2024-10-24
add "Deny All" Dns Resolver

parkma99:
implement #329

### DIFF
--- a/rama-dns/src/deny_all.rs
+++ b/rama-dns/src/deny_all.rs
@@ -1,0 +1,35 @@
+use crate::DnsResolver;
+use rama_net::address::Domain;
+use rama_utils::macros::error::static_str_error;
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+/// a [`DnsResolver`] implementation which
+/// denies all incoming DNS requests with a [`DnsDeniedError`].
+pub struct DenyAllDns;
+
+impl DenyAllDns {
+    #[inline]
+    /// Create a new [`Default`] [`DenyAllDns`].
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+static_str_error! {
+    #[doc = "Dns denied"]
+    pub struct DnsDeniedError;
+}
+
+impl DnsResolver for DenyAllDns {
+    type Error = DnsDeniedError;
+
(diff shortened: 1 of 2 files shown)


rama-dns/src/deny_all.rs
```diff
@@ -0,0 +1,22 @@
+use std::fmt::Error;
+
+use crate::DnsResolver;
+
+#[derive(Debug, Clone)]
+pub struct DenyAllDns {}
```
GlenDC, line 6:
```suggestion
#[non_exhaustive]
/// a [`DnsResolver`] implementation which
/// denies all incoming DNS requests with a [`DnsDeniedError`].
pub struct DenyAllDns;

impl DenyAllDns {
    #[inline]
    /// Create a new [`Default`] [`DenyAllDns`].
    pub fn new() -> Self {
        Self::default()
    }
}

```

rama-dns/src/deny_all.rs
```diff
@@ -0,0 +1,22 @@
+use std::fmt::Error;
+
+use crate::DnsResolver;
+
+#[derive(Debug, Clone)]
```
GlenDC, line 5:
```suggestion
#[derive(Debug, Clone, Default)]
```

GlenDC (changes requested):
Thanks, looks good.
Can you apply these changes,
and pull it onto your machine.

You might need to run `just qa` to quickly check if all passes.
It will also auto format in case there is some formatting to be done.

Should you not have `just` on your machine you probably can find install instructions
on <https://just.systems/man/en/installation.html> or for this instance you might get away to just check if `cargo fmt --all` doesn't give you new changes to commit.

If not too much to ask you can also squash your commits, but if not, also not ok, for now I'm still okay either way, it's what you want.

parkma99:
Thanks, now all tests is passed, I find git is little confusing😊

GlenDC:
> Thanks, now all tests is passed, I find git is little confusing😊

Yeah it is not something that one can just use without learning the nuances sadly.
Should you be interested I find <https://git-scm.com/book/en/v2> still the leading resource
that is worth reading at least once in your career.  Certainly most chapters, but not all.

Thanks for this contribution by the way, greatly appreciated.
Going to give it a last review, if it's okay and also all CI checks pass, I'll merge it in.
Otherwise I'll come back with more feedback.

Feel free to pick up another one, give feedback in any other way, or just leave it as-is.
Either way, have a nice life ahead of you and I wish you all the best!

GlenDC (approved):

Issue #332, opened 2024-10-03
support vec/array impl for DnsResolver


GlenDC:
It would allow you to chain multiple dns resolvers together, where it would try the next in case of error.

Note that this should not be used just to use multiple dns providers, as that can be configured already in a single resolver (eg Hickory)

GlenDC:
@parkma99 is picking this up

GlenDC:
This issue is available once again. Feel free to start from scratch or use @parkma99's original attempt as a starting point. All the same, whatever works best for the one picking this issue up.

NOOMA-42:
> This issue is available once again. Feel free to start from scratch or use @parkma99's original attempt as a starting point. All the same, whatever works best for the one picking this issue up.

I'd like to pick this up

GlenDC:
All yours @NOOMA-42 , let me know if you need guidance, help or feedback

PR #345, opened 2024-10-27
vec impl for DnsResolver

parkma99:
Hi @GlenDC I have a try for #332 But I am not sure my implement is totally that issue asked. And also, thanks for your helping

### DIFF
--- a/rama-dns/src/lib.rs
+++ b/rama-dns/src/lib.rs
@@ -79,128 +79,6 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
-#[derive(Debug)]
-pub struct DnsChainDomainResolveErr<E: 'static> {
-    errors: Vec<E>,
-}
-
-impl<E: std::fmt::Debug> std::fmt::Display for DnsChainDomainResolveErr<E> {
-    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(
-            f,
-            "domain resolver chain resulted in errors: {:?}",
-            self.errors
-        )
-    }
-}
-
-impl<E: std::fmt::Debug + Send + std::error::Error> std::error::Error
-    for DnsChainDomainResolveErr<E>
-{
-    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
-        self.errors.last().map(|e| e as &dyn std::error::Error)
-    }
-}
-
-impl<R, E> DnsResolver for Vec<R>
-where
-    R: DnsResolver<Error = E> + Send,
-    E: Send + 'static,
-{
-    type Error = DnsChainDomainResolveErr<E>;
-
-    async fn ipv4_lookup(&self, domain: Domain) -> Result<Vec<Ipv4Addr>, Self::Error> {
-        let mut errors = Vec::new();
-        for resolver in self {
-            match resolver.ipv4_lookup(domain.clone()).await {
-                Ok(ipv4s) => return Ok(ipv4s),
-                Err(err) => errors.push(err),
-            }
-        }
-        Err(DnsChainDomainResolveErr { errors })
-    }
-
-    async fn ipv6_lookup(&self, domain: Domain) -> Result<Vec<Ipv6Addr>, Self::Error> {
-        let mut errors = Vec::new();
-        for resolver in self {
-            match resolver.ipv6_lookup(domain.clone()).await {
-                Ok(ipv6s) => return Ok(ipv6s),
-                Err(err) => errors.push(err),
-            }
-        }
-        Err(DnsChainDomainResolveErr { errors })
-    }
-}
-
-impl<R, E, const N: usize> DnsResolver for [R; N]
-where
-    R: DnsResolver<Error = E> + Send,
-    E: Send + 'static,
-{
-    type Error = DnsChainDomainResolveErr<E>;
-
-    async fn ipv4_lookup(&self, domain: Domain) -> Result<Vec<Ipv4Addr>, Self::Error> {
-        let mut errors = Vec::new();
-        for resolver in self {
-            match resolver.ipv4_lookup(domain.clone()).await {
-                Ok(ipv4s) => return Ok(ipv4s),
-                Err(err) => errors.push(err),
-            }
-        }
-        Err(DnsChainDomainResolveErr { errors })
-    }
-
-    async fn ipv6_lookup(&self, domain: Domain) -> Result<Vec<Ipv6Addr>, Self::Error> {
-        let mut errors = Vec::new();
-        for resolver in self {
-            match resolver.ipv6_lookup(domain.clone()).await {
-                Ok(ipv6s) => return Ok(ipv6s),
-                Err(err) => errors.push(err),
-            }
-        }
-        Err(DnsChainDomainResolveErr { errors })
-    }
-}
-
-macro_rules! impl_dns_resolver_either_either {
-    ($id:ident, $($param:ident),+ $(,)?) => {
-        impl<$($param),+> DnsResolver for ::rama_core::combinators::$id<$($param),+>
-        where
-            $($param: DnsResolver<Error: Into<::rama_core::error::BoxError>>),+,
-        {
-            type Error = ::rama_core::error::BoxError;
-
-            async fn ipv4_lookup(
-                &self,
-                domain: Domain,
-            ) -> Result<Vec<Ipv4Addr>, Self::Error>{
-                match self {
-                    $(
-                        ::rama_core::combinators::$id::$param(d) => d.ipv4_lookup(domain)
-                            .await
-                            .map_err(Into::into),
-                    )+
-                }
-            }
-
-            async fn ipv6_lookup(
-                &self,
-                domain: Domain,
-            ) -> Result<Vec<Ipv6Addr>, Self::Error> {
-                match self {
-                    $(
-                        ::rama_core::combinators::$id::$param(d) => d.ipv6_lookup(domain)
-                            .await
-                            .map_err(Into::into),
-                    )+
-                }
-            }
-        }
-    };
-}
-
-rama_core::combinators::impl_either!(impl_dns_resolver_either_either);
-
 pub mod hickory;
 #[doc(inline)]
 pub use hickory::HickoryDns;
@@ -212,3 +90,7 @@ pub use in_memory::{DnsOverwrite, DomainNotMappedErr, InMemoryDns};
 mod deny_all;
 #[doc(inline)]
 pub use deny_all::{DenyAllDns, DnsDeniedError};
+
+pub mod chain;
+
+pub mod variant;
--- a/rama-dns/src/lib.rs
+++ b/rama-dns/src/lib.rs
@@ -79,6 +79,89 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
+#[derive(Debug)]
+pub struct DnsChainDomainResolveErr<E: 'static> {
+    errors: Vec<E>,
+}
+
+impl<E: std::fmt::Debug> std::fmt::Display for DnsChainDomainResolveErr<E> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "domain resolver chain resulted in errors: {:?}",
+            self.errors
+        )
+    }
+}
+
+impl<E: std::fmt::Debug + Send + std::error::Error> std::error::Error
+    for DnsChainDomainResolveErr<E>
+{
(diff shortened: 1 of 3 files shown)


rama-dns/src/lib.rs
```diff
@@ -79,6 +80,37 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
...
+static_str_error! {
+    #[doc = "domain not mapped in any dns resolver"]
+    pub struct DomainNotFoundErr;
+}
+
+impl<R: DnsResolver> DnsResolver for Vec<R> {
+    type Error = DomainNotFoundErr;
+
+    async fn ipv4_lookup(&self, domain: Domain) -> Result<Vec<Ipv4Addr>, Self::Error> {
+        let resolvers = self.iter().map(Arc::new).collect::<Vec<_>>();
```
GlenDC, line 92: Why are you doing this, should it not just be possible to do:

```rust
for resolver in self {
    // ...
}
```

if that gives an error you cannot work out, do feel free to paste it here and we can take it from there

rama-dns/src/lib.rs
```diff
@@ -79,6 +80,37 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
+static_str_error! {
+    #[doc = "domain not mapped in any dns resolver"]
+    pub struct DomainNotFoundErr;
+}
+
+impl<R: DnsResolver> DnsResolver for Vec<R> {
```
GlenDC, line 88: The issue requested the implementation for `Vec` and also `Array`.

If you only have bandwidth to do `Vec` that's okay, but know the array is pretty much the same.

The signature to impl Arrays would be something like:

```
impl<R: DnsResolver, const N: usize> DnsResolver for [R; N] {
    // ...
}
```

Something like that. Typing directly in comment here, so might have gotten some syntax wrong, but it's more or less like that

rama-dns/src/lib.rs
```diff
@@ -79,6 +80,37 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
+static_str_error! {
+    #[doc = "domain not mapped in any dns resolver"]
+    pub struct DomainNotFoundErr;
```
GlenDC, line 85: I think it's fine to have a custom error for it, but I would prefer if you implement a custom one by hand so we can keep track of _all_ errors:

```
pub struct DnsChainDomainResolveErr<E> {
    errors: Vec<E>,
}

impl<E: fmt::Debug> fmt::Debug for DnsChainDomainResolveErr<E> { ... }

impl<E: std::error::Error> std::error::Error for DnsChainDomainResolveErr<E> {
   fn source(...) { ... /* return last error as some */ }
}
```

This way you can keep track of errors with:

```rust
let mut errors = Vec::new();

// ...

Err(err) => errors.push(err),
```

GlenDC (changes requested):
Pretty much spot on. Hope my comments serve as the missing guidelines you were needing. However, you do seem to have understood the intent of the issue as far as I can see. So great start!

parkma99:
Thanks @GlenDC , I make some change in recent commits.  I would like to add some tests, can you give me some help?

GlenDC:
> Thanks @GlenDC , I make some change in recent commits. I would like to add some tests, can you give me some help?

Sure. What tests would you like to add? I can chip in if they are useful and how you might go on doing so.

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,88 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
+#[derive(Debug)]
+pub struct DnsChainDomainResolveErr<E: 'static> {
+    errors: Vec<E>,
+}
+impl<E: std::fmt::Debug + Send> std::fmt::Display for DnsChainDomainResolveErr<E> {
```
GlenDC, line 86:
```suggestion

impl<E: std::fmt::Debug> std::fmt::Display for DnsChainDomainResolveErr<E> {
```

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,88 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{:?}", self.errors)
```
GlenDC, line 88:
```suggestion
        write!(f, "domain resolver chain resulted in errors: {:?}", self.errors)
```

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,88 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
...
+    }
+}
+
+impl<E: std::fmt::Debug + Send + std::error::Error> std::error::Error
+    for DnsChainDomainResolveErr<E>
+{
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        if self.errors.is_empty() {
```
GlenDC, line 96: Pretty certain you can replace these lines with just a single call:

```
self.errors.last()
```

https://doc.rust-lang.org/std/vec/struct.Vec.html#method.last

(we want to return the last error btw, as source, given that's the reason why it failed in the end, as up to then it was still possible to have success)

GlenDC (changes requested):
Current version looks good except the minor comments I made.
With these changes applied I'm good with merging it, but curious to hear what kind of tests you want to still add.

parkma99:
> 
> Sure. What tests would you like to add? I can chip in if they are useful and how you might go on doing so.

Some demos for using DnsChainDomainResolve, just like in #332 mentioned

GlenDC:
> > Sure. What tests would you like to add? I can chip in if they are useful and how you might go on doing so.
> 
> Some demos for using DnsChainDomainResolve, just like in #332 mentioned

Sure. For general advise you can check existing tests in rama. E.g. how to do multiple cases of same tests (test cases), or fact that you'll need to use `tokio::test` for async tests, etc...

Regarding these tests. Wouldn't overdo it. Again use your own judgement, but following cases come to my mind:

1. test empty array/vec, will result in error but without inner errors (and that's ok)
2. test [Ok, Err]
3. test [Err, Ok]
4. test [Ok, Ok]
5. test [Err, Err, Ok]
6. test [Err, Err]

2-6 for both vec/array I guess.

You can use the existing https://ramaproxy.org/docs/rama/dns/struct.DenyAllDns.html for your err case, and https://ramaproxy.org/docs/rama/dns/struct.InMemoryDns.html for the ok case. I wouldn't check on the actual inner errors, just the high level Ok vs Err is good enough :) And if ok of course do check if you get the expected IP. Make sure to test both IPv4 and IPv6.

Feel free to cut on testing where you see fit, logic is simple enough that I don't think we really need that much testing here, but given you seem motivated to do so, do feel free to go for it :)

parkma99:
Hi @GlenDC  ,I still have no idea for testing, could you give me some example code , thanks

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,89 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
     }
 }
 
+#[derive(Debug)]
```
GlenDC, line 82: I think it might be a good idea to move all your new code here to a new file: `rama-dns/src/chain.rs`. Can be a private fail of which you pub use (re-export) types here.

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,89 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
...
+}
+
+impl<R, E> DnsResolver for Vec<R>
+where
+    R: DnsResolver<Error = E> + Send,
+    E: Send + 'static,
+{
+    type Error = DnsChainDomainResolveErr<E>;
+
+    async fn ipv4_lookup(&self, domain: Domain) -> Result<Vec<Ipv4Addr>, Self::Error> {
```
GlenDC, line 112: You can refactor these 2 definitions as:


```rust
macro_rules! dns_resolver_chain_impl {
    () => {
        async fn ipv4_lookup(&self, domain: Domain) -> Result<Vec<Ipv4Addr>, Self::Error> {
            let mut errors = Vec::new();
            for resolver in self {
                match resolver.ipv4_lookup(domain.clone()).await {
                    Ok(ipv4s) => return Ok(ipv4s),
                    Err(err) => errors.push(err),
                }
            }
            Err(DnsChainDomainResolveErr { errors })
        }
    
        async fn ipv6_lookup(&self, domain: Domain) -> Result<Vec<Ipv6Addr>, Self::Error> {
            let mut errors = Vec::new();
            for resolver in self {
                match resolver.ipv6_lookup(domain.clone()).await {
                    Ok(ipv6s) => return Ok(ipv6s),
                    Err(err) => errors.push(err),
                }
            }
            Err(DnsChainDomainResolveErr { errors })
        }
    }
}
```

so that you can do:

```rust
impl<R, E> DnsResolver for Vec<R>
where
    R: DnsResolver<Error = E> + Send,
    E: Send + 'static,
{
    type Error = DnsChainDomainResolveErr<E>;
    
    dns_resolver_chain_impl!()
}

impl<R, E, const N: usize> DnsResolver for [R; N]
where
    R: DnsResolver<Error = E> + Send,
    E: Send + 'static,
{
    type Error = DnsChainDomainResolveErr<E>;
    
    dns_resolver_chain_impl!()
}
```

Might give you an error around scoping or w/e, but nothing that can be resolved by in worst case
passing some stuff. With some cleverness you can probably even make it shorter, but this is already sufficient.

rama-dns/src/lib.rs
```diff
@@ -79,6 +79,89 @@ impl<R: DnsResolver<Error: Into<BoxError>>> DnsResolver for Option<R> {
...
+            match resolver.ipv6_lookup(domain.clone()).await {
+                Ok(ipv6s) => return Ok(ipv6s),
+                Err(err) => errors.push(err),
+            }
+        }
+        Err(DnsChainDomainResolveErr { errors })
+    }
+}
+
 macro_rules! impl_dns_resolver_either_either {
```
GlenDC, line 165: while you're at it, feel free to move this macro def and its usage to a separate file called `rama-dns/src/variant.rs`, keeps the lib.rs pretty clean

GlenDC:
> Hi @GlenDC ,I still have no idea for testing, could you give me some example code , thanks

Sure, in `rama-dns/src/in_memory.rs` there are already some example tests.

So to pick up some of the given example cases, you can start with:

```

#[cfg(test)]
mod tests {
    use std::net::{Ipv4Addr, Ipv6Addr};
    use rama_core::combinators::Either;

    use super::*;

    #[tokio::test]
    async fn test_empty_chain_vec() {
        let v = Vec::<DnsOverwrite>::new();
        assert!(v
            .ipv4_lookup(Domain::from_static("plabayo.tech"))
            .await
            .is_err());
        assert!(v
            .ipv6_lookup(Domain::from_static("plabayo.tech"))
            .await
            .is_err());
    }

    #[tokio::test]
    async fn test_empty_chain_array() {
        let a: [DnsOverwrite; 0] = [];
        assert!(a
            .ipv4_lookup(Domain::from_static("plabayo.tech"))
            .await
            .is_err());
        assert!(a
            .ipv6_lookup(Domain::from_static("plabayo.tech"))
            .await
            .is_err());
    }

    #[tokio::test]
    async fn test_chain_ok_err_ipv4() {
        let v = vec![
            Either::A(serde_html_form::from_str("example.com=127.0.0.1").unwrap()),
            Either::B(DenyAllDns::new()),
        ];
        assert!(
            v.ipv4_lookup(Domain::from_static("example.com"))
                .await
                .unwrap()
                .into_iter()
                .next()
                .unwrap(),
            Ipv4Addr::new(127, 0, 0, 1)
        );
        assert!(v
            .ipv6_lookup(Domain::from_static("example.com"))
            .await
            .is_err());
    }
    
    #[tokio::test]
    async fn test_chain_err_err_ok_ipv6() {
        // ... TODO
    }
    
    // TODO ...
}
```

Haven't tested the code or ran it so you might need to solve some issues in the code, but hopefully it gets the idea across. I certainly wouldn't test all possible combinations. Just try to combine a couple so that you more or less have tested some couple things. I'm not interested in full test coverage, so try to be minimal and smart about it, as this is also all code to maintain in the end. And thanks to how we write the code, e.g. using a macro for a single implementation, it is already more guaranteed that we have working code for both sequence implementations.

GlenDC:
Hi @parkma99 hope you are doing fine. Do you want to hand over this issue to someone else? Or do you want to continue yourself. In case of the latter, are you in need of assistance or help? Or mentoring of any kind?

GlenDC:
> Hi @GlenDC I want to hand over this issue to someone else because of busy working 😫😫

All good. Not a problem at all. Take care remote friend. Hope your work and rest of your life goes well!

PR #346, opened 2024-10-30
Feat-328/add ExtensionMatcher struct

calebbourg:
Adds two implementations. One for Const and one for closures.

### DIFF
--- a/rama-core/src/matcher/ext.rs
+++ b/rama-core/src/matcher/ext.rs
@@ -1,0 +1,78 @@
+use crate::{
+    matcher::{Extensions, Matcher},
+    Context,
+};
+use std::any::Any;
+
+pub struct ExtensionMatcher<T: Send + Sync> {
+    marker: T,
+}
+
+struct Const<T: Send + Sync>(T);
+struct FnBox(Box<dyn Fn(&dyn Any) -> bool + Send + Sync>);
+
+impl ExtensionMatcher<FnBox> {
+    pub fn with<T: Send + Sync + 'static, F: Fn(&T) -> bool + Send + Sync + 'static>(f: F) -> Self {
+        let wrapped_fn = move |v: &dyn Any| {
+            if let Some(concrete) = v.downcast_ref::<T>() {
+                f(concrete)
+            } else {
+                false
+            }
+        };
+        Self {
+            marker: FnBox(Box::new(wrapped_fn)),
+        }
+    }
+}
+
+impl<T: Send + Sync + Sync + std::cmp::PartialEq> ExtensionMatcher<Const<T>> {
+    pub fn constant(value: T) -> Self {
+        Self {
+            marker: Const(value),
+        }
+    }
+}
+
+impl<T: Send + Sync + std::cmp::PartialEq + 'static, State, Request> Matcher<State, Request>
+    for ExtensionMatcher<Const<T>>
+{
+    fn matches(&self, _ext: Option<&mut Extensions>, ctx: &Context<State>, _req: &Request) -> bool {
+        ctx.get::<T>()
+            .map(|v| v == &self.marker.0)
+            .unwrap_or_default()
+    }
+}
+
+impl<State, Request> Matcher<State, Request> for ExtensionMatcher<FnBox> {
+    fn matches(&self, _ext: Option<&mut Extensions>, ctx: &Context<State>, _req: &Request) -> bool {
+        ctx.get::<FnBox>()
+            .map(|v| (self.marker.0)(v))
+            .unwrap_or_default()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::context::Extensions;
+
+    #[derive(Clone, PartialEq)]
+    struct MyMarker(i32);
+
+    #[test]
+    fn test_extension_matcher() {
+        let mut ext = Extensions::new();
+        ext.insert(MyMarker(10));
+        let matcher = ExtensionMatcher::constant(MyMarker(10));
+        assert!(matcher.matches(Some(&mut ext), &Context::default(), &10));
+    }
+
+    #[test]
+    fn test_fn_extension_matcher() {
+        let mut ext = Extensions::new();
+        ext.insert(MyMarker(10));
+        let matcher = ExtensionMatcher::with(|v: &MyMarker| v.0.eq(&10));
+        assert!(matcher.matches(Some(&mut ext), &Context::default(), &10));
+    }
+}
--- a/rama-core/src/matcher/ext.rs
+++ b/rama-core/src/matcher/ext.rs
@@ -2,77 +2,135 @@ use crate::{
     matcher::{Extensions, Matcher},
     Context,
 };
-use std::any::Any;
 
-pub struct ExtensionMatcher<T: Send + Sync> {
-    marker: T,
+use std::marker::PhantomData;
+/// A matcher which allows you to match based on an extension,
+/// either by comparing with an extension ([`ExtensionMatcher::with_const`]),
+/// or using a custom predicate ([`ExtensionMatcher::with_fn`]).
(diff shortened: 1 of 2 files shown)


rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+    matcher::{Extensions, Matcher},
+    Context,
+};
+use std::any::Any;
+
+pub struct ExtensionMatcher<T: Send + Sync> {
+    marker: T,
+}
+
+struct Const<T: Send + Sync>(T);
```
GlenDC, line 11: This I would replace with something like the following:

```rust
mod private {
    use std::marker::PhantomData;

    pub(super) trait ExtensionPredicate<T>: Send + Sync + 'static {
        fn call(&self, value: &T) -> bool;
    }

    pub(super) struct PredicateConst<T>(pub(super) T);

    impl<T> ExtensionPredicate<T> for PredicateConst<T>
    where
        T: Clone + Eq + Send + Sync + 'static,
    {
        #[inline]
        fn call(&self, value: &T) -> bool {
            self.0.eq(value)
        }
    }

    pub(super) struct PredicateFn<F, T>(pub(super) F, pub(super) PhantomData<fn() -> T>);

    impl<F, T> ExtensionPredicate<T> for PredicateFn<F, T>
    where
        F: Fn(&T) -> bool + Send + Sync + 'static,
        T: Clone + Eq + Send + Sync + 'static,
    {
        #[inline]
        fn call(&self, value: &T) -> bool {
            self.0(value)
        }
    }
}
```

rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
+use crate::{
+    matcher::{Extensions, Matcher},
+    Context,
+};
+use std::any::Any;
+
+pub struct ExtensionMatcher<T: Send + Sync> {
```
GlenDC, line 7: This is how I would define the struct:

```rust
/// A matcher which allows you to match based on an extension,
/// either by comparing with an extension ([`ExtensionMatcher::with_const`]),
/// or using a custom predicate ([`ExtensionMatcher::with_fn`]).
pub struct ExtensionMatcher<P, T> {
    predicate: P,
    _marker: PhantomData<fn() -> T>,
}
```

rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+    marker: T,
+}
+
+struct Const<T: Send + Sync>(T);
+struct FnBox(Box<dyn Fn(&dyn Any) -> bool + Send + Sync>);
+
+impl ExtensionMatcher<FnBox> {
```
GlenDC, line 14: This is how you can create the fn variant:

```rust
impl<F, T> ExtensionMatcher<private::PredicateFn<F, T>, T>
where
    F: Fn(&T) -> bool + Send + Sync + 'static,
    T: Clone + Eq + Send + Sync + 'static,
{
    /// Create a new [`ExtensionMatcher`] with a predicate `F`,
    /// that will be called using the found `T` (if one is present),
    /// to let it decide whether or not the extension value is a match or not.
    pub fn with_fn(predicate: F) -> Self {
        Self {
            predicate: private::PredicateFn(predicate, PhantomData),
            _marker: PhantomData,
        }
    }
}
```

rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+                false
+            }
+        };
+        Self {
+            marker: FnBox(Box::new(wrapped_fn)),
+        }
+    }
+}
+
+impl<T: Send + Sync + Sync + std::cmp::PartialEq> ExtensionMatcher<Const<T>> {
```
GlenDC, line 29: This can be replaced with the following:

```rust
impl<T> ExtensionMatcher<private::PredicateConst<T>, T>
where
    T: Clone + Eq + Send + Sync + 'static,
{
    /// Create a new [`ExtensionMatcher`] with a const value `T`,
    /// that will be [`Eq`]-checked to find a match.
    pub fn with_const(value: T) -> Self {
        Self {
            predicate: private::PredicateConst(value),
            _marker: PhantomData,
        }
    }
}
```

rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn constant(value: T) -> Self {
+        Self {
+            marker: Const(value),
+        }
+    }
+}
+
+impl<T: Send + Sync + std::cmp::PartialEq + 'static, State, Request> Matcher<State, Request>
```
GlenDC, line 37: Thanks to the private trait trick you can now just implement the `ExtensionMatcher` with just 1 impl block:

```rust
impl<State, Request, P, T> Matcher<State, Request> for ExtensionMatcher<P, T>
where
    State: Clone + Send + Sync + 'static,
    Request: Send + 'static,
    T: Clone + Send + Sync + 'static,
    P: private::ExtensionPredicate<T>,
{
    fn matches(&self, _ext: Option<&mut Extensions>, ctx: &Context<State>, _req: &Request) -> bool {
        ctx.get::<T>()
            .map(|v| self.predicate.call(v))
            .unwrap_or_default()
    }
}
```

rama-core/src/matcher/ext.rs
```diff
@@ -0,0 +1,78 @@
...
+impl<State, Request> Matcher<State, Request> for ExtensionMatcher<FnBox> {
+    fn matches(&self, _ext: Option<&mut Extensions>, ctx: &Context<State>, _req: &Request) -> bool {
+        ctx.get::<FnBox>()
+            .map(|v| (self.marker.0)(v))
+            .unwrap_or_default()
+    }
+}
+
+#[cfg(test)]
+mod test {
```
GlenDC, line 56: This is how I Would implement the tests, I left 2 todo's in there, as to allow you to add some test cases of your own:

```rust
#[cfg(test)]
mod test {
    use super::*;

    #[derive(Debug, Clone, PartialEq, Eq)]
    struct MyMarker(i32);

    #[test]
    fn test_extension_matcher() {
        let matcher = ExtensionMatcher::with_const(MyMarker(10));
        let mut ctx = Context::default();

        assert!(!matcher.matches(None, &ctx, &()));

        ctx.insert(MyMarker(10));
        assert!(matcher.matches(None, &ctx, &()));

        // TODO: test for 2 more cases or so :)
    }

    #[test]
    fn test_fn_extension_matcher() {
        let matcher = ExtensionMatcher::with_fn(|v: &MyMarker| v.0 % 2 == 0);
        let mut ctx = Context::default();

        assert!(!matcher.matches(None, &ctx, &()));

        ctx.insert(MyMarker(4));
        assert!(matcher.matches(None, &ctx, &()));

        // TODO: test for 2 more cases or so :)
    }
}
```

GlenDC (changes requested):
You were on the right path, I added some code suggestions that you are free to copy as-is or change where you see fit. Some other remarks:

* In general don't box unless you have to. Thanks to embracing generics we do not need to box here, the way I showed you in my code suggestions is one way on how to avoid it while still allowing multiple variants;
* `Matcher`s take an optional `Extensions` struct, yes. But that's not the one to check against. This can be used by matchers to inject extra data (side-effects) that was generated as part of matching. E.g. a URI path matcher can insert there key fragments for its dynamic parts. We have to check instead against the `Context` which embeds an `Extensions`. And for the `ExtensionMatcher` there is no side-effect data to insert. So no need to test for that either.

I know it's a bit tricky, hope this makes sense to you. If not, don't mind asking more follow up questions :)

calebbourg:
@GlenDC Thank you for the feedback and help! I learned a tremendous amount in just reading and applying your suggestions. 
0b6d8d72946894d4edfb053d92e1819e9e17c29a

GlenDC (approved):

PR #347, opened 2024-11-03
port hyperium/* as rama-http-core

### LINKED ISSUE #86: Embed hyper and friends directly into rama
The design of hyper is similar enough that we ideally can just use it as is. And in a way that has been done so far. However akin to Tower and friends it also became clear that some of our goals and scope is sufficiently different that there will always be a kind of friction or trade off.

We developed most of rama 0.2 with hyper. In prototypes it became clear that a fork is needed for certain preservation goals. At that point even better is to just embed it directly into the codebase.

We can keep track in our FORK document what commit we are in sync with so that we can keep the codebases in sync.

This ticket is mostly about the embedding work and the use of it. We can already adapt code and drop code where we see it fit. But extensive modifications in the name of traffic preservation and control is best to be left for later tickets, to keep the scope of this ticket somewhat manageable.

GlenDC:
- dropped logic we do not need
- dropped wrapper/duplicate code no longer needed
- modify signatures to avoid having to need to lock
- keep file structure as similar as possible to easier facilitate keeping in sync with upstream (already proven to work as this PR already contains a couple of syncs with upstream, was very easy to do)

Closes #86


GlenDC (commented):
GTG

Issue #331, opened 2024-10-03
Support IP modes in connector and resolver


GlenDC:
enum IpMode
- dual (default)
- singleIpV4
- singleIpV6
- dualPreferIpv4

Then wrapper types can be made:

- DnsResolveIpMode(IpMode): control what Dns can resolve to
- ConnectIpMode(IpMode): define what ip can connect to

GlenDC:
These wrapper types are then to be used in the core tcp connector code, as to make sure we respect these desires when establishing a tcp connection.

AnkurRathore:
@GlenDC I would like to work on this. let me know what I need to do to start contributing.

GlenDC:
Hi @AnkurRathore , welcome and nice to meet you. It's all yours for now.

I can give you the high level steps that I would undertake if i would do it myself, but feel free to diviate where you see fit at your own best judgement:

1. implement the enum types as I'll lay them out at end of my comment;
2. use `DnsResolveIpMode` in <https://github.com/plabayo/rama/blob/45848f3e0276d7490e0f71e021823c897137dbed/rama-tcp/src/client/connect.rs#L153> to respect that enum (both to not resolve what shouldn't be resolved, and also to make it respect the preference. Because currently both IPv4 and IPv6 are allowed, and Ipv6 is currently always preferred (see delay))
3. use `ConnectIpMode` in (a) https://github.com/plabayo/rama/blob/45848f3e0276d7490e0f71e021823c897137dbed/rama-tcp/src/client/connect.rs#L121 to ensure we don't connect to IPv4/IPv6 if not desired, also use that same mode to overwrite `DnsResolveIPMode` in case ConnectIPMode is stricter. E.g. if IPv4 only is desired, than it's pointless to request Ipv6 addresses :)

```txt
enum DnsResolveIpMode
    dual (default)
    singleIpV4
    singleIpV6
    dualPreferIpv4

enum ConnectIpMode
    dual (default)
    ipv4
    ipv6
```

I do not know how good your rust or network knowledge is. So in case something is not clear now, or at any point, do feel free to ask. I'm here to mentor where you see fit. Also feel free to open a PR even if not yet "finished", in case you are stuck, want guidance or have questions of any other kind. I'm here. I do not always respond immediately, but I do promise I'll always get back to you eventually.

AnkurRathore:
[...]

PR #348, opened 2024-11-04
added wrapper type and todos

AnkurRathore:
@GlenDC I have attempted to add some wrapper types, and I have some doubts implementing them which I have put as TODO in the comments. I would need your guidance on this.


rama-tcp/src/client/connect.rs
```diff
@@ -23,6 +23,49 @@ use tokio::{
...
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum IPModes {
+    Dual,
+    SingleIpV4,
+    SingleIpV6,
+    DualPreferIpV4
+}
+
+impl Default for IPModes {
```
GlenDC, line 34: You can easier define this by using `Default` in the `#[derive(..)]` above and set `#[default]` above `Dual,`.

rama-tcp/src/client/connect.rs
```diff
@@ -23,6 +23,49 @@ use tokio::{
...
+    fn default() -> Self {
+        Self::Dual
+    }
+}
+
+//DNS Resolver
```
GlenDC, line 40: just a note for future commits: do put some effort in these comments once you are preparing for final review please, if you need help do let me know. Not important at this stage though, just remarking it already.

rama-tcp/src/client/connect.rs
```diff
@@ -23,6 +23,49 @@ use tokio::{
     },
 };
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum IPModes {
```
GlenDC, line 27: Originally I wanted to go for a single IP Mode indeed, and work with wrapper types. However I think it makes more sense to define 2 enums:

```
enum DnsResolveIpMode
    dual (default)
    singleIpV4
    singleIpV6
    dualPreferIpv4

enum ConnectIpMode
    dual (default)
    ipv4
    ipv6
 ```

Don't think these belong in `rama-tcp` though. Probably more like in `rama-net` or so.

File: rama-tcp/src/client/connect.rs [hunk 4]
```diff
@@ -23,6 +23,49 @@ use tokio::{
...
+
+impl<C>ConnectIpMode<C>{
+    fn new(connector: C, ip_mode: IPModes) -> Self {
+        Self {connector, ip_mode}
+    }
+}
+
+/*
+TODO - Using the wrapper types
+in the tcp_connect fn, update the dns type to use the DnsResolveIpMode
```
GlenDC, line 66: I think it's in `async fn tcp_connect_inner<State, Dns, Connector>(` that you'll want to check for `DnsResolveIpMode`.

And it's in the location where we actually establish a tcp stream that you'll check for `ConnectIpMode`.

Originally I was thinking to also automatically let one influence the other, but I think it makes more sense to just let the user worry about that. E.g. if user selects Ipv6-only for DNS and ipv4-only for Connect, then that obviously will fail, but that is up to the user to then fix. No point in trying to do anything about that, as the only reasonable thing to do would be to return some kind of error, so let's not bother with that as the current error will already be fine enough.

GlenDC (changes requested):
Let me know if my comments in this review are sufficient guidance or in case you need more detailed guidance. If the latter try to be more specific on where exactly you want my guidance with.

rama-tcp/src/client/connect.rs
AnkurRathore: Do I need to put these in rama-net?

rama-tcp/src/client/connect.rs
AnkurRathore: I was in two minds whether to commit as  I was still thinkig about the IPMode matching implementation. I will keep note of that. Thanks

GlenDC:
Hi @AnkurRathore , not that it's a blocker or urgent. But given this is already a month stale. I do want to at least check in.

* Are you stuck?
* Need guidance, mentoring, or help of any kind?
* Or did you move one and want to hand this over to someone else?

EIther and everything is fine. And hope you're in good health and condition. Take care!

GlenDC:
Understood. Do take care. For now I'll leave it with you then. It's not blocking any release schedule, as I still have plenty of other and much bigger fish to fry.

Should it still be left open in some weeks, I might re-assign it in case nothing changed by than. Either way good though, this is all voluntary. If you do feel ready by than to continue work on it, also good, I'll gladly support you with that. But otherwise also ok. Take care either way!

AnkurRathore:
@GlenDC  please review the latest changes. Let me know if I am on the correct path.

rama-tcp/src/client/connect.rs
```diff
@@ -210,37 +207,41 @@ where
...
-        connector,
-        IpKind::Ipv4,
-        ipv4_domain,
-        port,
-        ipv4_tx,
-        ipv4_connected,
-        ipv4_sem,
-    ));
+    match dns.mode {
+        DnsResolveIpMode::SingleIpV4 | DnsResolveIpMode::DualPreferIpV4 | DnsResovleIpMode::Dual =>{
```
GlenDC, line 211: You probably want to provide utility methods to `DnsResolveIpMode`, such as:

```
fn ipv4_supported(&self) -> bool;
fn ipv6_supported(&self) -> bool;
```

Is same-same, but easier to have to update such statements only in a single loc.

rama-tcp/src/client/connect.rs
```diff
@@ -274,39 +275,31 @@ async fn tcp_connect_inner_branch<Dns, Connector>(
...
     Connector: TcpStreamConnector<Error: Into<BoxError> + Send + 'static> + Clone,
 {
-    /*TODO
-    modify the match to use dns.ip_mode and match for 
-    SingleIpv4 or DualPreferIpv4 or Dual
-     */
+    
     let ip_it = match ip_kind {
-        IpKind::Ipv4 => match dns.ipv4_lookup(domain).await {
+        IpKind::Ipv4 if matches!(dns.mode, DnsResolveIpMode::Dual |DnsResolveIpMode::SingleIpV4 |DnsResolveIpMode::DualPreferIpV4 ) =>{
```
GlenDC, line 280: I would suggest to do:

```
match ip_kind => {
     IpKind::Ipv4 => if dns.mode.ipv4_supported() {
           // ... logic
     } else {
          trace::debug ipv4 not supported by DnsMode: no connection can be established
     }

     IpKind::Ipv6 => if dns.mode.ipv6_supported() {
           // ... logic
     } else {
          trace::debug ipv6 not supported by DnsMode: no connection can be established
     }
```

This way we have a clear trace in our logs / spans.

Btw, I don't think this part of the code is to use the Dns mode as no dns is to be resolved here. Here you need to use the IpMode or however we decided to call it.

rama-tcp/src/client/connect.rs
GlenDC: This is still to move to rama-net and `ConnectIpMode` is still to be used as well in the tcp connection part itself (unrelated from DnsResolveIpMode)

AnkurRathore:
@GlenDC I am doing good. Hope the same from your end.
I will go through the review comments.
Thanks.

AnkurRathore:
@GlenDC I have made the changes as per the comments. Please review them.

rama-net/src/mode.rs
```diff
@@ -0,0 +1,43 @@
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+/// Enum representing the IP modes that can be used by the DNS resolver.
+pub enum IPModes {
```
GlenDC, line 3: There are 2 enums required:

```
enum DnsResolveIpMode
    dual (default)
    singleIpV4
    singleIpV6
    dualPreferIpv4

enum ConnectIpMode
    dual (default)
    ipv4
    ipv6
```

The struct `DnsResolveIpMode` can be removed as there should be no need to wrap the resolver, this would be bad for ergonomics and doesn't really solve anything.

The difference between the 2 enums is that:

- ConnectIpMode is used only for establishing a connection
- DnsResolveIpMode is used only for resolving a domain
- these two are separate and do not influence each other.
- if they are not present in the `Context<S>` of the incoming request we would resort to the default
- DnsResolveIpMode has one more variant as there we also need to choose which one we are biased against, `Ipv4` or `Ipv6`, where the latter is the default to help progress the adoption of ipv6.

Sorry for not catching this earlier. Seens I missed it in my initial scans of your code.

Do let me know here or on Discord if something is not clear in that set of requirements.

rama-tcp/src/lib.rs
```diff
@@ -19,4 +19,4 @@
 
 pub mod client;
 pub mod server;
-pub mod utils;
+pub mod utils;
```
GlenDC, line 22: You removed a whiteline here

GlenDC:
> @GlenDC I have made the changes as per the comments. Please review them.

Added the requested review, some comments that already existed are still unresolved as well. Do let me know if you have more questions, need more guidance or anything else. Here, in the original github ticket (https://github.com/plabayo/rama/issues/331) or on Discord.

AnkurRathore:
@GlenDC addressed the required changes in enum. Please review

GlenDC:
> @GlenDC addressed the required changes in enum. Please review

Those changes look excellent. Do you need guidance for the remaining TODOs:

- respect ip version preference when resolving dns
- respect ipConnectMode when establishing a transport connection

AnkurRathore:
> > @GlenDC addressed the required changes in enum. Please review
> 
> Those changes look excellent. Do you need guidance for the remaining TODOs:
> 
> * respect ip version preference when resolving dns
> * respect ipConnectMode when establishing a transport connection

@GlenDC need your guidance on the remaining TODOs. Can you elaborate more on this.

GlenDC:
> @GlenDC need your guidance on the remaining TODOs. Can you elaborate more on this.

Sure thing

1. `respect ip version preference when resolving dns`

These sleep intervals would need to be swapped in case Ipv4 is prefered over ipv6

https://github.com/plabayo/rama/blob/1e2acaecb4dd868492ea076426112103eee817a5/rama-tcp/src/client/connect.rs#L279-L280

2. `respect ipConnectMode when establishing a transport connection`

In these locations you would need to block connections with an error in case of an Ipv4 address but ipv4 not allowed, and same for ipv6:

- https://github.com/plabayo/rama/blob/1e2acaecb4dd868492ea076426112103eee817a5/rama-tcp/src/client/connect.rs#L138
- https://github.com/plabayo/rama/blob/1e2acaecb4dd868492ea076426112103eee817a5/rama-tcp/src/client/connect.rs#L272

AnkurRathore:
@GlenDC Happy New Year!
I have addressed the above TODOs. Please review them.

rama-net/src/mode.rs
```diff
@@ -0,0 +1,31 @@
...
+impl DnsResolveIpMode{
+   
+    /// checks if IPv4 is supported in current mode
+    pub fn ipv4_supported(&self) -> bool {
+        matches!(self.mode, DnsResolveIpMode::Dual | DnsResolveIpMode::SingleIpV4 | DnsResolveIpMode::DualPreferIpV4)
+    }
+
+    /// checks if IPv6 is supported in current mode
+    pub fn ipv6_supported(&self) -> bool {
+        matches!(self.mode, DnsResolveIpMode::Dual | DnsResolveIpMode::SingleIpV6)
```
GlenDC, line 20:
```suggestion
        matches!(self.mode, DnsResolveIpMode::Dual | DnsResolveIpMode::SingleIpV6 | DnsResolveIpMode::DualPreferIpV4)
```

rama-net/src/mode.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+///Mode for establishing a connection
```
GlenDC, line 24:
```suggestion
/// Mode for establishing a connection
```

rama-tcp/src/client/connect.rs
```diff
@@ -117,6 +118,16 @@ where
     let domain = match host {
         Host::Name(domain) => domain,
         Host::Address(ip) => {
+            //check if IP Version is allowed
+            match (ip.is_ipv4(), connector) {
```
GlenDC, line 122: I would prefer you work with enums to have true exhaustive checking, e.g.

```rust
match (ip, connector) => {
    (IpAddr::V4(_), ConnectIpMode::Ipv6)) => ...
    (IpAddr::V6(_), ConnectIpMode::Ipv4)) => ...
    _ => (),
}
```

rama-tcp/src/client/connect.rs
```diff
@@ -167,37 +178,32 @@ where
...
-        IpKind::Ipv6,
-        ipv6_domain,
-        port,
-        ipv6_tx,
-        ipv6_connected,
-        ipv6_sem,
-    ));
+    if dns.ipv4_supported() {
+            ctx.spawn(tcp_connect_inner_branch(
+                dns.clone(),
```
GlenDC, line 183: You can add `Copy` derive trait to these mode Enums, then you don't need to Clone

rama-tcp/src/client/connect.rs
```diff
@@ -231,37 +237,69 @@ async fn tcp_connect_inner_branch<Dns, Connector>(
...
-            Ok(ips) => Either::A(ips.into_iter().map(IpAddr::V4)),
-            Err(err) => {
-                let err = OpaqueError::from_boxed(err.into());
-                tracing::trace!(err = %err, "[{ip_kind:?}] failed to resolve domain to IPv4 addresses");
-                return;
-            }
-        },
-        IpKind::Ipv6 => match dns.ipv6_lookup(domain).await {
+        IpKind::Ipv4 =>{
+            if dns.ipv4_supported(){
```
GlenDC, line 243: Why do that check here again? You already check if ipv4/ipv6 is supported prior to calling `tcp_connect_inner_branch`

rama-tcp/src/client/connect.rs
```diff
@@ -231,37 +237,69 @@ async fn tcp_connect_inner_branch<Dns, Connector>(
...
-            let delay = match ip_kind {
-                IpKind::Ipv4 => Duration::from_micros((21 * 2 * index) as u64),
-                IpKind::Ipv6 => Duration::from_micros((15 * 2 * index) as u64),
+            let delay = match (ip_kind, dns) {
+                //When IPv4 preffered give shorter delay
+                (IpKind::Ipv4,DnsResolveIpMode::DualPreferIpV4) => Duration::from_micros((15 * 2 * index) as u64),
+                (IpKind::Ipv6, DnsResolveIpMode::DualPreferIpV4) => Duration::from_micros((21 * 2 * index) as u64),
+                
+                //IPv6 preffered 
+                (IpKind::Ipv4, _) => Duration::from_micros((21 * 2 * index) as u64),
```
GlenDC, line 300: This logic is flawed I think, e.g. what if only Ipv4 is allowed in that case this would still give more delay for ipv4 even though it is the only one.

I would do it like this:


```rust
// move to before the for loop !!!

let (ipv4_delay_scalar, ipv6_delay_scalar) = match dns {
   DnsResolveIpMode::DualPreferIpV4 | DnsResolveIpMode::SingleIpv4 => (15 * 2 ,  21 * 2)
   DnsResolveIpMode::Dual, DnsResolveIpMode::SingleIpv6 => (21 * 2, 15 * 2)
};

// ...

for (index, ip) in ip_it.enumerate() {

// ..

let delay = match ip_kind {
     IpKind::Ipv4 => Duration::from_micros((ipv4_delay_scalar * index) as u64),
     IpKind::Ipv6 => Duration::from_micros((ipv6_delay_scalar * index) as u64),
};

// ..
```

GlenDC (changes requested):
Happy new year to you as well! Thanks for continuing this PR. Have left a couple of comments, ones these are resolved I believe we can merge it in!!!! Very exciting.

Once these comments are resolved can you please also run `just qa` and fix any warnings, errors that pop your way there, as well as commit any new changes that the formatter might have caused. If you don't have just yet you can find how to install it at <https://just.systems/man/en/packages.html>

rama-tcp/src/client/connect.rs
AnkurRathore: @GlenDC I have addressed the review. I think all that will be left is to run the just qa command.

GlenDC: Great. Are you able to run that command?

AnkurRathore: I was able to install it using cargo install command. Not familiar on how to use it. Can you guide me through the process?

GlenDC: Once https://just.systems/man/en/packages.html is installed you can run 'just qa' from the root directory of this project. It might complain that some cargo tools aren't installed, if that happens you need to install that tool manually and rerun 'just qa', until successful.

If that doesn't work for you for some reason I think it's fine if for this PR you only run these commands manually:

```
cargo fmt --all
cargo clippy --workspace --all-targets --all-features
cargo test --all-features --workspace
```

rama-tcp/src/client/connect.rs
```diff
@@ -154,8 +166,8 @@ async fn tcp_connect_inner<State, Dns, Connector>(
     ctx: &Context<State>,
     domain: Domain,
     port: u16,
-    dns: Dns,
-    connector: Connector,
+    dns: DnsResolveIpMode,
```
GlenDC, line 169: Does this still even compile? You remove `Dns` but `DnsResolveIpMode` doesn't provide actual dns functionality, yet you call `lookup_ipv4` later on it...

shouldn't this instead be `dns_mode: DnsResolveIpMode` while keeping `dns: Dns`?

GlenDC:
Hi @AnkurRathore hope you're doing fine? No pressure, just checking in here with you :) Hope you're doing ok! In case there's something I can be of meaning to you, do let me know :)

AnkurRathore:
@GlenDC Thanks for asking. I am doing good. Overwhelmed by some projects at work. I did get to compile the code and see some errors. I have been going through the code and trying to resolve the errors.

AnkurRathore:
@GlenDC Sorry for delaying the work on the pull request. I have made the changes, ran the tests, and got no issues. Please review and check. Hope this gets merged.

GlenDC:
Looks good. I have a couple last remarks but I'll finish it of for you. It's pretty much there and you did so much already.

AnkurRathore:
@GlenDC Thanks. Can I pick another issue?

GlenDC:
> @GlenDC Thanks. Can I pick another issue?

Always! Do as you please.
Perhaps https://github.com/plabayo/rama/issues/374 could be a good one for you,
and very useful for the entire project at the same time.

If you like it after checking it out do feel free to comment there and do ask any questions you might have about them.

The benchmark code that will be developed for that issue is pretty important to detect performance regressions and also aid in tuning the performance once we have the luxary to do that.

Issue #330, opened 2024-10-02
figure out how we can explicitly allow Anonymous users in layers such as ProxyAuth


GlenDC:
Currently these require an <https://ramaproxy.org/docs/rama/net/user/auth/trait.Authority.html> which on its turns requires a <https://ramaproxy.org/docs/rama/http/headers/authorization/trait.Credentials.html> and that is the issue for supporting Anonymous users.

We essentially need to be able to support an authority without it being hard-linked to a credential. This way we could easily support an Authority such as (NoAuthority == No UserId == Anonymous)

calebbourg:
@GlenDC I've looked a bit at the code involved here and wanted to put my thoughts on paper to see if this is at all in line with what you might have envisioned. If I'm way off that's ok too and I can take a look at other things when they come up :). I realize this is low priority and I want to respect your time. I very much appreciate all of the support.

My first thought (in somewhat pseudo code) would be to extract the credential handling into it's own trait something like

```rust
pub trait CredentialHandler<T> {
    fn validate_credentials(&self, credentials: &Self) -> bool;
}
```

and then the implementations for [Basic](https://github.com/plabayo/rama/blob/main/rama-net/src/user/credentials/basic.rs)

```rust
// move the responsibility of parsing here as well

impl<T: UsernameLabelParser>  CredentialHandler for Basic {
    fn validate_credentials(&self, credentials: &Self) -> bool {
        let username = credentials.username();
        let password = credentials.password();

        if password != self.password() {
            return false;
        }

        let mut parser_ext = Extensions::new();
        let username = match parse_username(&mut parser_ext, T::default(), username) {
            Ok(t) => t,
            Err(err) => {
                tracing::trace!("failed to parse username: {:?}", err);
                return if self == credentials {
                    ext.insert(UserId::Username(username.to_owned()));
                    true
                } else {
                    false
                };
            }
        };

        if username != self.username() {
            return false;
        }

        ext.extend(parser_ext);
        ext.insert(UserId::Username(username));
        true
    }
}
```

and 

```rust
impl AuthoritySync<Basic> for Basic {
    fn authorized(&self, ext: &mut Extensions, credentials: &Basic) -> bool {
        self.validate_credentials(credentials)
    }
}
```

With this approach I hit an decision [here](https://github.com/plabayo/rama/blob/main/rama-http/src/layer/proxy_auth.rs#L149-L155) 

```rust
[...]

PR #349, opened 2024-11-05
Feat-330/allow anonymous

calebbourg:
Adds `allow_anonymous` concept to existing auth layers.


calebbourg:
@GlenDC I provided the logic in `ProxyAuthService`'s `Service` implementation to acknowledge `allow_anonymous` but did not for `AsyncRequireAuthorization`. It seemed it may be better handled in a given `AsyncAuthorizeRequest` implementation as that is where operations on header values and things like that were happening in the examples. Wasn't 100% on that choice but curious what your vision was.

Also would it make sense for me to add an example showing how this might be used?

rama-http/src/layer/proxy_auth.rs
```diff
@@ -162,6 +196,12 @@ where
                     .body(Default::default())
                     .unwrap())
             }
+        } else if self.allow_anonymous {
+            let mut user_ext = Extensions::new();
+            user_ext.insert(UserId::Anonymous);
+            ctx.extend(user_ext);
```
GlenDC, line 202:
```
            ctx.insert(UserId::Anonymous);
```

You can make this a one-liner

rama-http/src/layer/auth/async_require_authorization.rs
```diff
@@ -130,12 +130,28 @@ use std::future::Future;
 #[derive(Debug, Clone)]
 pub struct AsyncRequireAuthorizationLayer<T> {
     auth: T,
+    allow_anonymous: bool,
```
GlenDC, line 133: This is the wrong layer/service to implement it for. As this is a generic layer which allows any kind of custom authorization. Good one if you want to implement your own authorization, but not in need of `allow_anonymous` as that would be something the used `auth: T` will implement themselves.

The proxy auth is good, that one you can keep.

You can still add it to `ValidateRequestHeaderLayer` and its service however. That one is about the `Authorization` header. Logic to be done there is pretty much what you did for `proxy_auth.rs`.

GlenDC (changes requested):
Well done, it is exactly in the right direction and partly spot on.
Added a couple of remarks and one guide post/comment to help you reach the finish line. Thanks for your contribution and do let me know if there's any questions, remarks, feedback or anything else I can be of help with.

rama-http/src/layer/auth/async_require_authorization.rs
calebbourg: Thank you very much!

Question on this [here](https://github.com/plabayo/rama/blob/main/rama-http/src/layer/validate_request/validate_request_header.rs#L176-L178) just to make sure I understand the intention: 

```rust
    async fn serve(
        &self,
        ctx: Context<State>,
        req: Request<ReqBody>,
        match self.validate.validate(ctx, req).await {
            Ok((ctx, req)) => self.inner.serve(ctx, req).await,
            Err(res) => Ok(res),
        }
    }
}
```
Is the expectation to insert `UserId::Anonymous` into the `Context` as part of the`Err(res)` arm? If so it seems like that differs from the logic of `proxy_auth.rs`  you had outlined in https://github.com/plabayo/rama/issues/330#issuecomment-2452376257

> If and only if the header used by that layer is absent and the bool is true we will authorize the user as UserID::Anonymous

In the `ValidateRequestHeader` service, if I'm understanding it correctly, there could be a situation where `validate(ctx, req)` fails because an `Accept: application/json` header isn't present. Would it follow that we would then still continue the chain of layers if `allow_anonymous == true` in that situation?

Something like this:

```rust
async fn serve(
        &self,
        ctx: Context<State>,
        req: Request<ReqBody>,
        match self.validate.validate(ctx, req).await {
            Ok((ctx, req)) => self.inner.serve(ctx, req).await,
            Err(res) => {
                if self.allow_anonymous {
                   // these are borrow after moves
                    ctx.insert(UserId::Anonymous);

                    self.inner.serve(ctx, req).await
                } else {
                    Ok(res)
                }
            }
        }
    }
```

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -258,6 +260,53 @@ where
...
+        req: Request<B>,
+    ) -> Result<(Context<S>, Request<B>), Response<Self::ResponseBody>> {
+        match req.headers().get(header::AUTHORIZATION) {
+            Some(_) => self.credential.validate(ctx, req).await,
+            None if self.allow_anonymous => {
+                ctx.insert(UserId::anonymous());
+
+                Ok((ctx, req))
+            }
+            None => self.credential.validate(ctx, req).await,
```
calebbourg, line 296: Obviously this part is awkward and requires checking headers twice.  hadn't put much thought into quite yet.

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -258,6 +260,53 @@ where
     }
 }
 
+pub(crate) struct AuthorizeContext<C> {
```
calebbourg, line 263: This winds up leaking into the public interface

GlenDC:
```suggestion
pub struct AuthorizeContext<C> {
```

You're correct. Just make this `pub(crate) struct AuthorizeContext<C>`. It's still private in creation as it cannot be created outside of the crate, but at least it lets you name the type :)

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -258,6 +260,53 @@ where
...
+impl<C> AuthorizeContext<C> {
+    pub(crate) fn new(credential: C) -> Self {
+        Self {
+            credential,
+            allow_anonymous: false,
+        }
+    }
+}
+
+impl<S, B, C: ValidateRequest<S, B>> ValidateRequest<S, B> for AuthorizeContext<C>
```
GlenDC, line 277: Why do you add this implementation? `C` is always a credential in this context. And we want to have control over what can be inside of `AuthorizeContext` that's why it's semi private. As such I don't see a reason for this impl block to exist at all?

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -258,6 +260,53 @@ where
...
+                ctx.insert(UserId::anonymous());
+
+                Ok((ctx, req))
+            }
+            None => self.credential.validate(ctx, req).await,
+        }
+    }
+}
+
+impl<C: Clone> Clone for AuthorizeContext<C> {
```
GlenDC, line 301: also implement `fmt::Debug` like this please

GlenDC (changes requested):
You are getting close.

Things to do, next to my remarks is:

- The new `impl` block you added for `ValidateRequest` needs to be removed.
- update the existing `impl` blocks for that trait:
   - `impl<S, B, ResBody> ValidateRequest<S, B> for Bearer<ResBody>` <- this replaced with the wrapper `AuthorizeContext<Bearer<ResBody>>`
   - `impl<S, B, ResBody> ValidateRequest<S, B> for Basic<ResBody>` <- this replaced with the wrapper `AuthorizeContext<Basic<ResBody>>`
 - You also need to provide setter methods for setting that `allow_anonymous` on the layer + service by adding these to:
    - `impl` blocks for things like `impl<S, C> ValidateRequestHeaderLayer<S, AuthorizeContext<C>>
    - ` impl<S,  C> ValidateRequestHeader<S, AuthorizeContext<C>> `
    - no need to differentiate `C` between `Basic` and `Bearer` here, given the constructors that we provide already limit what `C` can be, and `C`'s value is not used here. So this keeps it easy and minimal

rama-http/src/layer/auth/require_authorization.rs
GlenDC: This is a mis-use/interpretation of the API. `impl` block can be removed as remarked in the other comments.

GlenDC:
Thank you @calebbourg for your contribution, effort and patience on this one. I hope the feedback does not discourage you. As always I'm here for you if there are remarks, feedback, questions or anything else of such nature. Take care.

rama-http/src/layer/auth/require_authorization.rs
calebbourg: TBH I didn't consider updating the existing implementations for `Basic` and `Bearer` to include `AuthorizeContext`, I think because I was hesitant for the type to be more pervasive. I was a bit fixated on the "privateness" of `AuthorizeContext` which I see now was not necessary. Without adding it to the other `ValidateRequest` implementations I had no way (that I could see) of accessing the `allow_anonymous` value and so this was the approach I took. 
Given your suggestions I can totally see why that approach is nicer. Thank you for that.

rama-http/src/layer/auth/require_authorization.rs
calebbourg: added 922cfc4eaf06ec4ad9964cafb71520a3aa8e02f3

calebbourg:
Thank you very much. Not discouraged at all. I welcome and appreciate the feedback. It's very helpful.

rama-http/src/layer/validate_request/validate_request_header.rs
```diff
@@ -90,7 +90,7 @@ where
 /// See the [module docs](crate::layer::validate_request) for an example.
 pub struct ValidateRequestHeader<S, T> {
     inner: S,
-    validate: T,
+    pub(crate) validate: T,
```
calebbourg, line 93: Changed this so that I could access it in `require_authorization` with the intention of avoiding circular dependency between this module and `require_authorization` (if I had implemented the `x_allow_anonymous` functions here)

Could expose this through a method if preferable.

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -183,8 +213,14 @@ where
         request: Request<B>,
     ) -> Result<(Context<S>, Request<B>), Response<Self::ResponseBody>> {
         match request.headers().get(header::AUTHORIZATION) {
-            Some(actual) if actual == self.header_value => Ok((ctx, request)),
+            Some(actual) if actual == self.credential.header_value => Ok((ctx, request)),
```
GlenDC, line 216: It's a bit philosophical, but I consider this not correct implementation. Anonymous users should be anonymous because the header is not defined. IMHO if the header is defined it has to be correct. In Proxy-Auth you do that correct, but here in this file you also are allowing it to be anonymous if credential is wrong, which is not correct in my opinion.

rama-http/src/layer/proxy_auth.rs
```diff
@@ -162,6 +196,10 @@ where
                     .body(Default::default())
                     .unwrap())
             }
+        } else if self.allow_anonymous {
+            ctx.insert(UserId::Anonymous);
+
```
GlenDC, line 201:
```suggestion
```

nit pick , but please delete this empty line

rama-http/src/layer/validate_request/validate_request_header.rs
GlenDC: No this is fine. It's still not part of public rama-http api so this is ok as-is. We still have freedom to change it easily this way. Well done.

GlenDC (changes requested):
Well done, and fast indeed. Don't sweat it though, unless you're having fun :)
A couple of minor remarks, other than that I think we're good to go once these are resolved I'll give it a last review and if all is well I can merge it in if you're ok with it as well.

rama-http/src/layer/proxy_auth.rs
calebbourg: No problem 71063ce1abd54aa1e4fd76cf48600749849a2ba4

GlenDC (approved):
Well done, thank you for going the extra mile.
The added unit tests and documentation is greatly appreciated!

💯🎈

PR #351, opened 2024-11-07
Feat 330/allow anonymous trait implementation

calebbourg:


GlenDC:
Problem with this approach is that you are now adding it to multiple parts of the stack.
E.g. you have it in ProxyAuth but also in the credentials themselves.
It also doesn't make really sense to add it to a credential as what's the point of a creating that credential in first place.
And finally you implement it from stuff like `Bearer<()>` but nothing else. 

I do like the initiative, but I believe it's the wrong approach, regardless of the flaws that would need to be fixed if you would go for it. Still, nice attempt though, it shows that you start to feel more comfortable with the code base and that is nice to see. Good job regardless.

PR #359, opened 2024-12-08
Add SocketAddress Abstraction for IP Address and Port Handling

### LINKED ISSUE #358: Add SocketAddress to rama-net
It's pretty much like Authority from the same rama-net package (https://ramaproxy.org/docs/rama/net/address/index.html)

But it is a IpAddr Port (u16) pair, so IpAddr instead of Host.

We can also make it convertible to and from std's SocketAddr.

This type is convenient if you want to enforce it to be a socket compatible addr, which Authority can't do for you.

hafihaf123:
# Changes

## New File: socket_address.rs

- Defines the SocketAddress struct.
- Implements key methods like new, ip_addr, port, into_ip_addr, and into_parts.
- Provides From conversions between SocketAddr and SocketAddress.
- Adds fmt::Display implementation for intuitive string formatting of addresses.
- Adds TryFrom and FromStr implementation for conversions
- Adds serde::Serialize and serde::Deserialize implementations for seamless serialization/deserialization.

## Modified File: mod.rs

- Adds the socket_address module.
- Re-exports SocketAddress with inline documentation.

## Modified File: authority.rs

- changed visibility of the `split_port_from_str` function to lessen code redundancy to `pub(crate)`

# Tests

- copied the tests from *authority.rs* and modified them to fit the structure
- runs **successfully**

#  Purpose

- Fixes #358


rama-net/src/address/socket_address.rs
```diff
@@ -0,0 +1,257 @@
...
+            ".-",
+            "::1",
+            "127.0.0.1",
+            "[::1]",
+            "2001:db8:3333:4444:5555:6666:7777:8888",
+            "[2001:db8:3333:4444:5555:6666:7777:8888]",
+            "example.com",
+            "example.com:",
+            "example.com:-1",
+            "example.com:999999",
```
GlenDC, line 231:
```suggestion
            "example.com:999999",
            "example.com:80",
```

Make sure to also add a valid Authority, as that's valid for an authority but shouldn't be valid for a SocketAddress

rama-net/src/address/socket_address.rs
```diff
@@ -0,0 +1,257 @@
...
+            sock_address.ip_addr().to_string(),
+            ip_addr,
+            "parsing: {}",
+            s
+        );
+        assert_eq!(sock_address.port(), port, "parsing: {}", s);
+    }
+
+    #[test]
+    fn test_parse_valid() {
```
GlenDC, line 174: You copied this from Authority I think. Where I see the same `fn test_parse_valid() {`.

But looking at it, I think that test doesn't do anything useful. This is not your fault as my original test is similar. But now that you're working on this it might be good to verify and if needed fix it.

A line like:

```
assert_eq(s, s.parse().expect(&msg), expected_ip_addr, expected_port)
```

The first is a string (`s`) and the second is `s.parse().expect(...)`, meaning the second one parses into the same type as the first, ... which is a string.

Probably need like a valid actual SocketAddr as a new added first parameter, so that `parse()` actually deserialized into a SocketAddress. Can you verify this and if needed fix it for both `authority.rs` and `socket_address.rs`.

GlenDC (changes requested):
Looks good as a starter. Have two comments, in case those are resolved I think we're pretty much there.

hafihaf123:
Hello @GlenDC ,

I have addressed the issues you mentioned and updated the tests accordingly. They are now running successfully, but I’m not entirely confident I’ve implemented them in the correct manner. I’d appreciate your feedback to ensure everything aligns with best practices.

If these changes look good, I can apply similar updates to authority as needed. 

Additionally, I made the following changes:
- Adjusted the visibility of one function in `authority.rs` and another in `host.rs` to `pub(super)` to reduce code duplication. Please let me know if this is an appropriate approach.

rama-net/src/address/socket_address.rs
```diff
@@ -0,0 +1,295 @@
...
+            ("[::1]:80", SocketAddress::new("::1".parse().unwrap(), 80)),
+            ("127.0.0.1:80", SocketAddress::new("127.0.0.1".parse().unwrap(), 80)),
+            (
+                "[2001:db8:3333:4444:5555:6666:7777:8888]:80",
+                SocketAddress::new("2001:db8:3333:4444:5555:6666:7777:8888".parse().unwrap(), 80),
+            ),
+        ] {
+            let msg = format!("parsing '{}'", s);
+
+            assert_eq(s, s.parse().expect(&msg), expected_socket_address.ip_addr(), expected_socket_address.port());
```
GlenDC, line 222:
```suggestion
            assert_eq(expected_socket_address, s.parse().expect(&msg), expected_socket_address);
```

Not sure if it will compile, but this is more or less the change you need to make for also all the assert_eq's below this. Goal is to compare to an actual address, otherwise you're just validating string with string, which again to be fair is also due to my fault of having that buggy test in the first place in `authority.rs`

GlenDC (changes requested):
Good, added a last comment, You can apply that. If you then just `just qa` and all runs fine you can commit. Try to also run `just hack` That should also pass. If it does, you can immediately apply the same fix for the analogue test in `authority.rs`.

Thanks upfront!

Btw can you move those utility functions from `authority.rs` that you marked as `pub(super)` to a new `rama-net/src/address/parse_utils.rs` file which can remain `pub(super)`. But this way we avoid having these address files refer to one another.

hafihaf123:
Hello @GlenDC,

I have updated the tests in both socket.rs and authority.rs. Hopefully, they meet expectations this time and I’m open to further feedback if needed.

While running `just qa`, I noticed warnings in `authority.rs` related to a `function cannot return without recursing` issue. Since this file was among those I modified, I wanted to bring it to your attention. However, I believe this is a separate issue and not directly caused by my changes.

Please let me know if further adjustments are required!

rama-net/src/address/authority.rs
```diff
@@ -229,43 +218,47 @@ impl<'de> serde::Deserialize<'de> for Authority {
 mod tests {
     use super::*;
 
-    fn assert_eq(s: &str, authority: Authority, host: &str, port: u16) {
```
GlenDC, line 232: So sorry! I totally forgot and didn't read that I had defined a custom `assert_eq` function... Can you please revert this test. E.g. leave `test_parse_valid` untouched and put this code back.

Also please do a similar thing for your `socket_address.rs`' `test_parse_valid` test. This old approach is much cleaner.

So sorry. I had read it wrongly as `assert_eq!`. But now that I see you deleting `assert_eq` (a custom fn I had defined) I now realise I did not make a mistake after all.

You did a lot already though, and I don't want to put you into a ping-pong show. So do tell me if this is too much and I'll take it over for this minor change. But if you are up for it please do so, as this is probably the last bit before we can merge.

GlenDC (changes requested):
Awesome work. Thanks already. Let me know if you're up for that revert + minor change. Once that is done (by you or me) I'll merge this in.

Either way, thanks for the amazing contribution. Very much appreciated.

rama-net/src/address/authority.rs
hafihaf123: Thanks for clarifying! No problem, I’ll revert the `test_parse_valid` changes in both `authority.rs` and `socket_address.rs` to restore the use of the custom`assert_eq` function.

I’ll update the PR shortly!

rama-net/src/address/parse_utils.rs
```diff
@@ -0,0 +1,24 @@
+use rama_core::error::{ErrorExt, OpaqueError};
+use std::net::{IpAddr, Ipv6Addr};
+
+pub(in crate::address) fn split_port_from_str(s: &str) -> Result<(&str, u16), OpaqueError> {
```
GlenDC, line 4:
```suggestion
pub(super) fn split_port_from_str(s: &str) -> Result<(&str, u16), OpaqueError> {
```

rama-net/src/address/parse_utils.rs
```diff
@@ -0,0 +1,24 @@
...
+        match s[colon + 1..].parse() {
+            Ok(port) => Ok((&s[..colon], port)),
+            Err(err) => Err(err.context("parse port as u16")),
+        }
+    } else {
+        Err(OpaqueError::from_display("missing port"))
+    }
+}
+
+pub(in crate::address) fn try_to_parse_str_to_ip(value: &str) -> Option<IpAddr> {
```
GlenDC, line 15:
```suggestion
pub(super) fn try_to_parse_str_to_ip(value: &str) -> Option<IpAddr> {
```

rama-net/src/address/socket_address.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[inline]
+    fn from((ip, port): (Ipv4Addr, u16)) -> Self {
+        (ip, port).into()
+    }
+}
+
+impl From<([u8; 4], u16)> for SocketAddress {
+    #[inline]
+    fn from((ip, port): ([u8; 4], u16)) -> Self {
+        (<[u8; 4] as Into<IpAddr>>::into(ip), port).into()
```
GlenDC, line 83:
```suggestion
        let ip: IpAddr = ip.into();
        (ip, port).into()
```

rama-net/src/address/socket_address.rs
```diff
@@ -0,0 +1,296 @@
...
+    #[inline]
+    fn from((ip, port): (Ipv6Addr, u16)) -> Self {
+        (ip, port).into()
+    }
+}
+
+impl From<([u8; 16], u16)> for SocketAddress {
+    #[inline]
+    fn from((ip, port): ([u8; 16], u16)) -> Self {
+        (<[u8; 16] as Into<IpAddr>>::into(ip), port).into()
```
GlenDC, line 97:
```suggestion
        (let ip: IpAddr = ip.into();
        (ip, port).into()
```

GlenDC (changes requested):
Looks great, added some minor suggestions, you can apply them from within github if you agree. Other than that let me know if you think it's ready, because if you think it's done I'll prepare it today still to get it merged. Thanks for the work!

hafihaf123:
From my side, I think it’s ready, but let me know if there’s anything else to address. Appreciate your help in getting this prepared for merge!

GlenDC (approved):
All good. Thanks!

PR #361, opened 2024-12-11
Fix boring tls termination example and test

soundofspace:
This example was writing an http response directly without waiting for a proper http request. In itself there was nothing wrong with this example directly, but it was not a valid http server for sure. This broke the test that was using this example as a server because hyper will crash when there is unexpected data in the connection buffers.

In some occasions this example did work: either due to very lucky timing or because the retry layer repeated enough until it got lucky. This also explains the problems we saw where the retry layer seemed stuck from the outside, which we thought was a bug, but might have just been this bug infinitely repeating


GlenDC (approved):
LGTM

PR #362, opened 2024-12-14
Ability to store peer certificate chain

soundofspace:


rama-http-backend/src/client/svc.rs
```diff
@@ -47,11 +47,12 @@ where
         // directly instead of here...
         let req = sanitize_client_req_header(&mut ctx, req)?;
 
-        let resp = match &self.0 {
+        let mut resp = match &self.0 {
             SendRequest::Http1(sender) => sender.lock().await.send_request(req).await,
             SendRequest::Http2(sender) => sender.lock().await.send_request(req).await,
         }?;
 
+        resp.extensions_mut().insert(ctx);
```
soundofspace, line 55: Do we want to return ctx, similar to how it is done in other places, or inject it in request extensions? Injecting it in extensions is pretty easy and transparent to use

soundofspace: Did not add e2e test for this as it is very small, and usage will be included in the dynamic certificate issuer e2e test.

Example of how to use this
```rust
inner_client.set_tls_config(ClientConfig {
    store_server_certificate_chain: true,
    ..Default::default()
});

...

let neg_tls_params = response
    .extensions()
    .get::<Context<()>>()
    .unwrap()
    .get::<NegotiatedTlsParameters>()
    .unwrap();

let certificates = neg_tls_params.peer_certificate_chain.clone().unwrap();
```

rama-net/src/tls/mod.rs
```diff
@@ -95,3 +95,58 @@ pub enum DataEncoding {
     /// Privacy Enhanced Mail (PEM) (plain text)
     Pem(NonEmptyString),
 }
+
+#[cfg(feature = "boring")]
```
GlenDC, line 99: You should be able to implement these directly in boring/rustls modules. Child modules are AFAIK not subject to the orphan rule, but might recall this wrongly so.

rama-tls/src/boring/client/connector.rs
```diff
@@ -437,9 +438,24 @@ impl<S, K> TlsConnector<S, K> {
...
+                    .is_some_and(|data| data.connect_config_input.store_server_certificate_chain);
+
+                let server_certificate_chain = store_server_cert_chain
+                    .then(|| {
+                        stream
+                            .ssl()
+                            .peer_cert_chain()
+                            .map(|chain| chain.try_into().ok())
+                    })
+                    .flatten()
```
GlenDC, line 452: you might be trying to be a bit too clever with the functional style here. E.g. using it on booleans in this way seems to force you to flatten twice. Might want to write it a bit different

rama-tls/src/boring/server/service.rs
```diff
@@ -15,11 +15,11 @@ use rama_core::{
 use rama_net::{
     http::RequestContext,
     stream::Stream,
-    tls::{client::NegotiatedTlsParameters, ApplicationProtocol},
+    tls::{client::NegotiatedTlsParameters, ApplicationProtocol, DataEncoding},
     transport::TransportContext,
 };
 use rama_utils::macros::define_inner_service_accessors;
-use std::{io::ErrorKind, sync::Arc};
+use std::{io::ErrorKind, sync::Arc, vec};
```
GlenDC, line 22: `vec` is already in prelude

rama-http-backend/src/client/svc.rs
GlenDC: Injecting context into the response feels dirty/wrong.

But to be fair, the fact that the context dies at the end of a chain did always botter me...

Still in spirit of current codebase this doesn't feel like the right solution. Instead one probably would use a layer to extract that information out before the context is being passed on to die.

It's unconvenient, that's true, but for now it's the only way to go I think.

GlenDC (changes requested):

rama-tls/src/rustls/client/connector.rs
```diff
@@ -432,14 +432,11 @@ impl<S, K> TlsConnector<S, K> {
...
-        let server_certificate_chain = store_server_cert_chain
-            .then(|| {
-                conn_data_ref
-                    .peer_certificates()
-                    .map(|chain| chain.try_into().ok())
-            })
-            .flatten()
-            .flatten();
+        let server_certificate_chain = if store_server_cert_chain {
+            conn_data_ref.peer_certificates().map(DataEncoding::from)
```
GlenDC, line 436: You can normally also do `map(Into::into)`, which is already in the prelude :)

rama-tls/src/boring/client/connector.rs
```diff
@@ -442,15 +442,15 @@ impl<S, K> TlsConnector<S, K> {
...
-                let server_certificate_chain = store_server_cert_chain
-                    .then(|| {
-                        stream
-                            .ssl()
-                            .peer_cert_chain()
-                            .map(|chain| chain.try_into().ok())
-                    })
-                    .flatten()
-                    .flatten();
+                let server_certificate_chain = if store_server_cert_chain {
```
GlenDC, line 445: I have nothing against flatten btw, just the double flatten and the removal of the error was a bit unwanted, e.g. flatten can play a role here, just in a different way: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2c71e9399ed9a1fc851a18cbb044be54

rama-net/src/tls/mod.rs
soundofspace: Implemented this in the root module as there are currently no separate modules for boring and rustls at this level. Would need a decent refactor. But as a rework is planned in the future for tls this could be done then

rama-http-backend/src/client/svc.rs
soundofspace: As discussed in private, this is part of a bigger problem that is out of scope for this PR. As such a new issue will be created where we will further discuss this problem and solutions. Once a solution is merged, I will also use update code here to use this new approach to expose inner layer stuff to the outside

rama-tls/src/rustls/client/connector.rs
soundofspace: Cool trick thanks

rama-tls/src/boring/client/connector.rs
soundofspace: True agree that the initial approach was way too complex. Just lots of different ways to do this and playing around a bit to find the optimal approach, feedback here is greatly appreciated

Issue #354, opened 2024-11-13
support dynamic issuer in tls cert issuer logic


GlenDC:
At the time of writing this issue there is already the following support for tls acceptor auth config:

```rust
pub enum ServerAuth {
    SelfSigned(SelfSignedData),
    Single(ServerAuthData),
    CertIssuer(ServerCertIssuerData),
}
```

Where `ServerCertIssuerData` is:

```rust
pub struct ServerCertIssuerData {
    pub kind: ServerCertIssuerKind,
    pub max_cache_size: u64,
}
```

and `ServerCertIssuerKind` is:

```rust
pub enum ServerCertIssuerKind {
    SelfSigned(SelfSignedData),
    Single(ServerAuthData),
}
```

The goal of this issue is to provide a third variant to `ServerCertIssuerKind` such that it becomes:

```rust
pub enum ServerCertIssuerKind {
    SelfSigned(SelfSignedData),
    Single(ServerAuthData),
    Dynamic(???),
}
```

The dynamic "thing" would support any kind of custom logic to support getting certs. The thing could receive some kind of `CertIssueInput`, for example:

```rust
struct CertIssuerInput {
     pub sni: Host,
}
```

Proposed steps to resolve this issue:

1. choose an approach, or create your own, work out on paper and align here in the issue in the comment section with me.
2. implement the approach and open a PR as soon as you feel ready, want early feedback, are stuck, have questions, or want to collaborate.

Some approaches we can make this happen:

1. provide a trait `ServerCertIssuer` and work with a Box<dyn ServerCertIssuer>`; This approach is the most obvious but also not one I like as it means we have to box by default

2. provide a channel based approach such that the variant would become something like (pseudo code):

```
  // ...
  Dynamic(ServerCertInputSender),
}

struct ServerCertInputSender {
     sender: Sender<ServerCertInputReceiver>,
}

struct ServerCertInputReceiver {
    sender: oneshot<Result<ServerAuthData, BoxError>>,
}

// the internal details of `ServerCertInputSender` and `ServerCertInputReceiver` would be hidden behind an api that simply uses
// the `send` methods. E.g. `send(input)` and `send(result)`. Wthout having to know the internal implementation
// or that a channel is even used
```

3. something else. Feel free to propose something other than the above proposals

Generics is something we do want to avoid here as it would make the entire type and using it a lot more cumbersome / complicated. Even when not using this dynamic approach.

---

[...]

PR #365, opened 2024-12-17
Issue 354/support dynamic issuer in tls cert issuer logic

soundofspace:
Implements #354 for generic tls config and boring ssl acceptor.


examples/tls_boring_dynamic_certs.rs
```diff
@@ -0,0 +1,220 @@
+// TODO docs
```
GlenDC, line 1:
```suggestion
```

?

examples/tls_boring_dynamic_certs.rs
```diff
@@ -0,0 +1,220 @@
...
+            },
+            None => Ok(self.example_data.clone()),
+        }
+    }
+}
+
+pub fn example_self_signed_auth() -> Result<ServerAuthData, OpaqueError> {
+    Ok(ServerAuthData {
+        private_key: DataEncoding::Pem(
+            std::str::from_utf8(include_bytes!("./assets/example.com.key"))
```
GlenDC, line 180: Seems you want to use include_str! instead

examples/tls_boring_dynamic_certs.rs
```diff
@@ -0,0 +1,220 @@
...
+                .expect("should work"),
+        ),
+        ocsp: None,
+    })
+}
+
+async fn http_service<S>(_ctx: Context<S>, _request: Request) -> Result<Response, Infallible> {
+    Ok(Response::new(
+        format!(
+            "hello client, you were served by boring tls terminator proxy issuing a dynamic certificate\r\n"
```
GlenDC, line 216: Pretty certain you can just call into_response() on the &'static str. Why the newlines btw?

rama-net/src/tls/server/config.rs
```diff
@@ -72,6 +73,9 @@ pub struct ServerCertIssuerData {
     /// The max amount of certs to cache,
     /// a default value of 8096 is used if 0.
     pub max_cache_size: u64,
+    /// If cache should be disabled when
+    /// using a dynamic issuer.
+    pub disable_cache_for_dynamic_issuer: bool,
```
GlenDC, line 78: Might want to replace these 2 properties with:

```
enum CacheKind {
     MemCache {
        max_size: u64
     }
     Disabled,
}
```

With default implemented as Self::MemCache { max_size: 0 } }

Something like that. Wdyt?

rama-net/src/tls/server/config.rs
```diff
@@ -92,6 +96,12 @@ impl Default for ServerCertIssuerKind {
     }
 }
 
+impl ServerCertIssuerKind {
+    pub fn from_dynamic_issuer<T: DynamicCertIssuer + Send + Sync + 'static>(issuer: T) -> Self {
```
GlenDC, line 100: Probably best to instead just set the SS'S trait bounds enforced on the DynamicCertIssuer already, similar to Service.

I imagine we can globally be more clever about this in a year or two, but for now this is what it is

rama-net/src/tls/server/config.rs
```diff
@@ -118,6 +128,69 @@ pub struct ServerAuthData {
...
+        server_name: Option<Host>,
+    ) -> Result<ServerAuthData, OpaqueError> {
+        self.issuer.issue_cert(client_hello, server_name).await
+    }
+}
+
+impl std::fmt::Debug for DynamicIssuer {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DynamicIssuer").finish()
+    }
```
GlenDC, line 157: Is there a reason to not dynamically impl it if and only if dyn-issuer == Debug?

rama-net/src/tls/server/config.rs
```diff
@@ -118,6 +128,69 @@ pub struct ServerAuthData {
...
+pub trait DynamicCertIssuer {
+    fn issue_cert(
+        &self,
+        client_hello: ClientHello,
+        server_name: Option<Host>,
+    ) -> impl Future<Output = Result<ServerAuthData, OpaqueError>> + Send + Sync + '_;
+}
+
+/// Internal trait to support dynamic dispatch of trait with async fn.
+/// See trait DynService<S, Request> for more info about this pattern.
```
GlenDC, line 171: Can you link to it using markdown links, you can do so using rust mod syntax etc

rama-net/src/tls/server/config.rs
```diff
@@ -118,6 +128,69 @@ pub struct ServerAuthData {
...
+    ) -> Pin<Box<dyn Future<Output = Result<ServerAuthData, OpaqueError>> + Send + Sync + '_>>;
+}
+
+impl<T> DynDynamicCertIssuer for T
+where
+    T: DynamicCertIssuer,
+{
+    fn issue_cert(
+        &self,
+
```
GlenDC, line 186:
```suggestion
```

?

rama-tls/src/boring/server/service.rs
```diff
@@ -99,10 +99,14 @@ where
             })
             .or_else(|| ctx.get::<RequestContext>().map(|ctx| ctx.authority.host()));
 
+        let mut maybe_client_hello = self
+            .store_client_hello
+            .then_some(Arc::new(Mutex::new(None)));
```
GlenDC, line 104: Why did you go for an arc-mutex instead of a oneshot channel? Is it ever expected then to be set more then once? You can only do one successful handshake for a single conn, or am I missing something?

GlenDC (changes requested):

GlenDC:
> Not ready yet still needs some cleanup all over the place

Is this still the case, or out of date PR description @soundofspace ?

rama-net/src/tls/server/config.rs
soundofspace: Yea will be cleaner to use, doesn't it in that case also make more sense to set max_size to 8096, so the 0 is not treated as a default, but instead we set the default directly

GlenDC: I was thinking about that as well, but unless you use something like an u64 but that cannot ever be 0, you would anyway do that default size fallback later as well, in which case you have duplicate logic that can get out of sync

GlenDC: Eg https://doc.rust-lang.org/std/num/type.NonZeroU64.html

soundofspace: Yea was thinking about using a non zero variant to indeed force correct usage

GlenDC: Amazing, aligned we are.

rama-tls/src/boring/server/acceptor_data.rs
```diff
@@ -83,13 +89,11 @@ impl TlsCertSource {
         self,
         mut builder: SslAcceptorBuilder,
         server_name: Option<Host>,
+        maybe_client_hello: &Option<Arc<Mutex<Option<RamaClientHello>>>>,
```
GlenDC, line 92: This would be `maybe_client_hello: Option<oneshot::Sender<RamaClientHello>>`

rama-tls/src/boring/server/service.rs
soundofspace: Mainly because this logic was ported from the old approach, which used arc for this. Will check if a oneshot channel makes more sense, but will need to confirm this as some of these callbacks can be called multiple times (but pretty sure that the ones we currently use are only called once)

rama-net/src/tls/server/config.rs
soundofspace: What exactly do you mean with this?

rama-tls/src/boring/server/acceptor_data.rs
```diff
@@ -328,6 +345,66 @@ impl TryFrom<rama_net::tls::server::ServerConfig> for TlsAcceptorData {
...
+                OpaqueError::from_display("sni parse failed")
+            })? // from client (e.g. only possibility for SNI proxy)
+        }
+        (_, Some(host)) => {
+            tracing::trace!(%host, "boring: server_name not in sni: using context");
+            host.clone() // from context (lower prio)
+        }
+        (None, None) => {
+            tracing::warn!(
+                "boring: no host found in server_name or ctx: defaulting to 'localhost'"
```
GlenDC, line 363: Not your design choice, I know I already did this but I wonder if this is ever expected or desired... Seems like a footgun now that I think about it. I mean if you do want localhost seems like not an issue to let them explicitly choose that?

rama-net/src/tls/server/config.rs
GlenDC: Sorry my bad, didn't make any sense what i wanted in the first place. Let's keep it like you have for now.

tests/integration/examples/example_tests/tls_boring_dynamic_certs.rs
```diff
@@ -0,0 +1,199 @@
...
+use std::{str::FromStr, time::Duration};
+
+#[tokio::test]
+#[ignore]
+async fn test_tls_boring_dynamic_certs() {
+    utils::init_tracing();
+
+    let chain = DataEncoding::DerStack(
+        X509::stack_from_pem(include_bytes!(
+            "../../../../examples/assets/example.com.crt"
```
GlenDC, line 39: Is there a good reason why you have these as crt/keys instead of just using the easy to use Self-Signed logic that's already in there? If it is because we currently don't expose it to be used as a stand-alone thing, it might make sense to do so as part of this PR as I would prefer not having to check in crt/key pairs if I can avoid it.

rama-tls/src/boring/server/acceptor_data.rs
soundofspace: Not sure about this, having this default to localhost does make sense to me as long as `server_name: Option<Host>` is an option. Making this just a `Host` would solve this problem, but would move "defaulting" up one layer, or we could require setting a host all the way up, just not sure what is ideal here

tests/integration/examples/example_tests/tls_boring_dynamic_certs.rs
soundofspace: Main reason here to add them as files, is because I didn't find a way to import functions/constants from the examples logic in an easy way. Any tips here are welcome

rama-tls/src/boring/server/acceptor_data.rs
GlenDC: yeah but that doesn't work given you also allow the ssl_ref path... so not sure either. We can keep it for now like this, but something to be kept in mind.

tests/integration/examples/example_tests/tls_boring_dynamic_certs.rs
GlenDC: Ah.. right... Ok that does make sense. Fair enough, you can keep it like this. Only thing I can think of is to work via an intermediate test-utils crate but feels like a lot of effort for little gain.

soundofspace: Yea this is something I do plan to revise once we start thinking how we can make these examples smarter and more scalable, while still keeping them simple and easy to use examples.

rama-tls/src/boring/server/service.rs
soundofspace: Was indeed because callbacks can be called multiple times. Left a note and we can reconsider this approach once we merge boringssl bindings

GlenDC (approved):
Given how the current state of rama-tls is I think this is outstanding work @soundofspace .

However it also really shows that we learned a lot about the TLS interfaces from the work done for getting to this current implementation. Looking forward to seriously revisiting the tls story in rama. Happy that you want to drive that as well!

PR #368, opened 2024-12-25
feat: support vec/array impl for DnsResolver

NOOMA-42:
# Description
support vec/array impl for DnsResolver to enable chaining resolver together

# Related issues
https://github.com/plabayo/rama/issues/332

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,15 +55,15 @@ dependencies = [
 
 [[package]]
 name = "allocator-api2"
-version = "0.2.21"
+version = "0.2.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
+checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
 name = "anstream"
-version = "0.6.18"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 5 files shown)


Cargo.lock
GlenDC, line 1: Can you revert this file please as it is unrelated from this PR.

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,236 @@
+use std::net::{Ipv4Addr, Ipv6Addr};
+
+use rama_net::address::Domain;
+
+use crate::{DenyAllDns, DnsResolver, InMemoryDns};
+
+#[derive(Debug)]
+pub struct DnsChainDomainResolveErr<E: 'static> {
```
GlenDC, line 8:
```suggestion
pub struct DnsChainDomainResolveErr<E> {
```

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,236 @@
...
+where
+    R: DnsResolver<Error = E> + Send,
+    E: Send + 'static,
+{
+    type Error = DnsChainDomainResolveErr<E>;
+    dns_resolver_chain_impl!();
+}
+
+#[derive(Debug)]
+enum DnsResolverType {
```
GlenDC, line 76: Why did you add this type? You can use the `Either` wrapper type for this purpose, no?

GlenDC (changes requested):
Hi! thanks for picking up this PR where it was left. Left a couple of remarks already.

rama-dns/src/chain.rs
NOOMA-42: Doesn't E need to be static to ensure there's no case like below ?
```rust
struct NonStaticError<'a> {
    message: &'a str  
}

let message = String::from("temp error");
let err = DnsChainDomainResolveErr {
    errors: vec![NonStaticError {
        message: &message  // The error exists after message be destroyed
    }]
};
```

GlenDC: Yes more or less, but in general it's better not to add such bounds at the definition level, but add such bounds only the impl blocks where you really need them. Otherwise you are forced to add these bounds constantly. And in some cases it's even possible that such bounds are optional and can also work in some scenarios without them.

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+
+use rama_net::address::Domain;
+
+use crate::DnsResolver;
+use rama_core::combinators::Either;
+
+/// An error that occurs when a DNS resolver chain fails to resolve a domain.
+#[derive(Debug)]
+pub struct DnsChainDomainResolveErr<E: 'static> {
+    errors: Vec<E>,
```
GlenDC, line 11: We probably want to store `BoxError` in this `Vec<..>` and do not need a generic error type at all.

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+}
+
+impl<E: std::fmt::Debug> std::fmt::Display for DnsChainDomainResolveErr<E> {
```
GlenDC, line 14: When you use BoxError internally this generic bound is no longer needed and won't exist

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "domain resolver chain resulted in errors: {:?}",
+            self.errors
+        )
+    }
+}
+
+impl<E: std::error::Error + 'static> std::error::Error for DnsChainDomainResolveErr<E> {
```
GlenDC, line 24: same here, can be removed this bound and generic once BoxError is used

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+            }
+            Err(DnsChainDomainResolveErr { errors })
+        }
+    };
+}
+
+impl<R, E> DnsResolver for Vec<R>
+where
+    R: DnsResolver<Error = E> + Send,
+    E: Send + 'static,
```
GlenDC, line 61: this bound will change into something like `Into<BoxError>`

rama-dns/src/chain.rs
```diff
@@ -0,0 +1,192 @@
...
+{
+    type Error = DnsChainDomainResolveErr<E>;
+
+    dns_resolver_chain_impl!();
+}
+
+impl<R, E, const N: usize> DnsResolver for [R; N]
+where
+    R: DnsResolver<Error = E> + Send,
+    E: Send + 'static,
```
GlenDC, line 71: this bound will change into something like `Into<BoxError>`

rama-dns/src/variant.rs
```diff
@@ -0,0 +1,118 @@
+use crate::DnsResolver;
+use rama_net::address::Domain;
+use std::net::{Ipv4Addr, Ipv6Addr};
+macro_rules! impl_dns_resolver_either_either {
```
GlenDC, line 4:
```suggestion

macro_rules! impl_dns_resolver_either_either {
```

GlenDC (changes requested):
Different DnsResolvers probably have their own error types. So in that case, and given this is a generic feature, best to collect the errors as `BoxError` so that we can collect the errors regardless of the types.

rama-dns/src/chain.rs
NOOMA-42: I added Send + Sync bounds to ensure thread safe

NOOMA-42:
> Different DnsResolvers probably have their own error types. So in that case, and given this is a generic feature, best to collect the errors as `BoxError` so that we can collect the errors regardless of the types.

I have a more general question which might outside of the scope of this pull request. When do we have to implement a specific error type and when to implement only box? We could discuss in Discord if this is inappropriate to discuss here.

rama-dns/src/in_memory.rs
```diff
@@ -59,6 +64,16 @@ impl InMemoryDns {
         self
     }
 
+    /// Convenience method to insert an IPv4 address for a domain.
+    pub fn insert_ipv4(&mut self, name: impl Into<Domain>, addr: Ipv4Addr) -> &mut Self {
```
GlenDC, line 68:
```suggestion
    pub fn insert_ipv4(&mut self, name: Domain, addr: Ipv4Addr) -> &mut Self {
```

There is not a whole lot that can be turned directly into a domain as an infallible process. So might as well just expect the concrete type itself.

rama-dns/src/in_memory.rs
```diff
@@ -59,6 +64,16 @@ impl InMemoryDns {
...
+        self.insert(name.into(), vec![IpAddr::V4(addr)])
+    }
+
+    /// Convenience method to insert an IPv6 address for a domain.
+    pub fn insert_ipv6(&mut self, name: impl Into<Domain>, addr: Ipv6Addr) -> &mut Self {
```
GlenDC, line 73:
```suggestion
    pub fn insert_ipv6(&mut self, name: Domain, addr: Ipv6Addr) -> &mut Self {
```

There is not a whole lot that can be turned directly into a domain as an infallible process. So might as well just expect the concrete type itself.

There is however some stuff that can be turned infallible into an Ipv4/Ipv6 addr (e.g. [u8; 4)). So might be nice addition to turn the addr part into an impl parameter for both your insert methods.

rama-dns/src/lib.rs
```diff
@@ -129,3 +90,7 @@ pub use in_memory::{DnsOverwrite, DomainNotMappedErr, InMemoryDns};
 mod deny_all;
 #[doc(inline)]
 pub use deny_all::{DenyAllDns, DnsDeniedError};
+
+pub mod chain;
+
+pub mod variant;
```
GlenDC, line 96:
```suggestion
mod variant;
```

AFAIK no reason to have this public. It's just about implementing something.

GlenDC (changes requested):
Minor changes still to go, other than that it seems this PR is about to be ready to be merged into the main branch! thx a lot!!!

rama-dns/src/in_memory.rs
NOOMA-42: I made it a general method. let me know if this is okay

GlenDC (approved):

NOOMA-42:
Thanks for this detailed explanation :) @GlenDC

PR #373, opened 2024-12-29
Add Domain Address DataType to abstract Domains and Port

Hrushi20:
## New File
domain_address.rs 
* File abstract Domains and Port

## Modified 
authority.rs
* Add Type conversion from DomainAddress to Authority
* Export DomainAddress in mod.rs

## Todo
* Add tests


GlenDC:
Looks like you're off to a great start. Don't forget to run `cargo` fmt, clippy and the like. Other than that, all good. Agreed with your TODO as well.

rama-net/src/address/domain_address.rs
```diff
@@ -0,0 +1,123 @@
...
+        address.serialize(serializer)
+    }
+}
+
+impl<'de> serde::Deserialize<'de> for DomainAddress {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let s = String::deserialize(deserializer)?;
```
GlenDC, line 107: you can avoid / delay a string allocation here btw. I know you copied this from a bad pattern I had started in this codebase, but just now I started rectifying this: https://github.com/plabayo/rama/commit/1e2acaecb4dd868492ea076426112103eee817a5

The gist is that you can instead of 

```
let s = String::deserialize(deserializer)?;
s.try_into().map_err(serde::de::Error::custom)
```

do 

```
let s = <&'de str>::deserialize(deserializer)?;
s.parse().map_err(serde::de::Error::custom)
```

In this case you anyway still need a heap allocation once we create the domain,
but for the cases where the parsing would fail (e.g. because str is invalid) you would have avoided an heap allocation, making at the very least failures cheaper.

rama-net/src/address/domain_address.rs
```diff
@@ -0,0 +1,198 @@
...
+                s.as_bytes().to_vec().try_into().expect(&msg),
+                expected_domain,
+                expected_port,
+            );
+        }
+    }
+
+    #[test]
+    fn test_parse_invalid() {
+        for s in [
```
GlenDC, line 153: Can you also add some tests to ensure that valid addresses that use an ip address are still invalid for here. E.g. 127.0.0.1:8080, as well as an ipv6 version

Hrushi20: Hey! I see `Domain.rs` parses 127.0.0.1 as a valid domain. Will fix this and update code.

rama-net/src/address/domain_address.rs
```diff
@@ -0,0 +1,212 @@
...
+        Self::try_from(s)
+    }
+}
+
+impl TryFrom<&str> for DomainAddress {
+    type Error = OpaqueError;
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        let (domain, port) = parse_utils::split_port_from_str(s)?;
+        eprintln!("Domain: {}", domain);
```
GlenDC, line 66: This needs to be removed

rama-net/src/address/domain_address.rs
GlenDC: 127.0.0.1 is a valid domain though, odd as it may seem. Might have to just accept that reality.

Hrushi20:
```
127.0.0.1 is a valid domain though, odd as it may seem. Might have to just accept that reality.
```

I've added 127.0.0.1 as valid input in `domain.rs` test. Done the same in `domain_address.rs`

rama-net/src/address/domain_address.rs
Hrushi20: Do I update Domain.rs to not accept 127.0.0.1 as a valid domain? I'm confused here.

GlenDC: Nope 127.0.0.1 is valid sadly, I forgot that when creating this ticket. Ipv6 addresses are however not valid.

The good news is that domain 127.0.0.1 will still fail at runtime as that wont dns resolve.

GlenDC: You still can add the it fails ti parse [ipv6]:port

rama-net/src/address/domain_address.rs
```diff
@@ -0,0 +1,212 @@
+use crate::address::{parse_utils, Domain};
+use rama_core::error::{ErrorContext, OpaqueError};
+use std::fmt;
+use std::str::FromStr;
+
+/// A [`Domain`] with an associated port
+
```
GlenDC, line 7:
```suggestion
```

GlenDC (approved):
All good, thx!

GlenDC:
Thanks for the work @Hrushi20.
Squashed and merged it pretty much as is into main: https://github.com/plabayo/rama/commit/dd4fcbd214377060114d80d69abf89590be01854

Hrushi20:
Thanks for the help. Will remember to do that next time.

Issue #387, opened 2025-01-10
Decouple TransportContext from ProxyDB


GlenDC:
Today ProxyDB from rama-proxy takes in a TransportContext in order to get the TransportProtocol defined in there (tcp/udp).

This can be seen here: https://docs.rs/rama/latest/rama/proxy/trait.ProxyDB.html

This however makes it awkward to use this functionality where a TransportContext makes no sense (eg because you don't have an Authority to look up a proxy).

Instead we could provide a new strucy 'ProxyContext' in https://docs.rs/rama/latest/rama/proxy/index.html with just TransportProtocol, e.g. something like:

```
#[derive(Debug, Clone, Default)]
pub struct ProxyContext { transport: Option<TransportProtocol> }

```

and can `impl From<TransportContext> for ProxyContext { ... }`, as that will make it also easy to maintain this relationship and use that transformer in the proxy (service) layer

That is we can than adept https://docs.rs/rama/latest/rama/proxy/struct.ProxyDBService.html to turn the TransportContext there into a ProxyContext prior to looking up a Proxy.

YHM404:
Hi, I’m interested in this issue, and if it’s possible, I’d like to give it a try.

GlenDC:
Hi, it's all yours. I'll assign it to you. Is all clear? Do you need additional guidance? Anything you want to talk through. Feel also free to open a Draft PR at any stage.

GlenDC:
Hi @YHM404 it seems I was a bit too early in approving your PR after my initial review on the phone. Found a couple of improvements:

- minor doc fix
- move ProxyContext to `rama-proxy` as it is nothing to do with `rama-net`
- no need to clone an entire transport context just to copy out a very cheap enum variant

you can find the commit for educational reference at <https://github.com/plabayo/rama/commit/967acfd6685d7648952f085d0eddaff1ec1594bb>, sorry for not catching it during my review prior to merging it. I do always do an extra review after merge just in case I missed something anyway.

PR #392, opened 2025-01-12
refactor(proxy): introduce ProxyContext for transport layer

YHM404:
Resolve https://github.com/plabayo/rama/issues/387

Simplify transport context handling by introducing a dedicated ProxyContext struct that only contains the essential transport protocol information needed for proxy selection.

- Add new ProxyContext struct in rama-net/transport.rs
- Replace TransportContext with ProxyContext in ProxyDB trait and implementations
- Remove unused app_protocol, http_version and authority fields from proxy selection logic
- Update tests to use simplified ProxyContext

This change reduces coupling between the proxy and transport layers by only exposing the minimal required information for proxy selection.

### DIFF
--- a/rama-net/src/transport.rs
+++ b/rama-net/src/transport.rs
@@ -26,6 +26,23 @@ pub struct TransportContext {
     pub authority: Authority,
 }
 
+/// The context as relevant to the proxy layer.
+///
+/// This is used to pass proxy related information to the transport layer.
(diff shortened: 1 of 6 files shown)


GlenDC (approved):

GlenDC:
Thx looks exactly as expected! Quick work on this one!

PR #393, opened 2025-01-12
Issue 382/ support X-Robots-Tag as a typed http header XRobotsTag

### LINKED ISSUE #382: support X-Robots-Tag as a typed http header XRobotsTag
It should support the rules as documented in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Robots-Tag

Additionally these rules can also be added:

```
X-Robots-Tag: noai, noimageai, SPC
```

You can have one or multiple active, some rules are just a value others are a `name: value` pair.

By default it can just comma-separate the values, but feel free to make it opt-in to have 1 rule per header (and thus mutliple headers.

hafihaf123:
## Initial Implementation of `X-Robots-Tag` Header  

This pull request introduces an initial implementation of the `X-Robots-Tag` header for the `rama-http` crate and closes #382

### Summary of Changes:  

1. **Header Registration:**  
   - Added the `x-robots-tag` header to the static headers list in `rama-http-types/src/lib.rs`.  

2. **Implementation of the Header:**  
   - Created a `XRobotsTag` struct that represents the header and implements the `rama_http::headers::Header` trait.  
   - Defined a `Rule` enum to represent indexing rules such as `noindex`, `nofollow`, and `max-snippet`. Complex rules like `max-image-preview` and `unavailable_after` are also included.  
   - Designed an `Element` struct to represent a rule optionally associated with a bot name.  
   - Added an iterator for `XRobotsTag` to iterate over its elements.  

3. **File Structure:**  
   - The implementation resides in a new module at `rama-http/src/headers/x_robots_tag/`, which includes the following files:  
     - `rule.rs`: Defines the `Rule` enum and parsing logic.  
     - `element.rs`: Implements the `Element` struct and its parsing/formatting logic.  
     - `iterator.rs`: Provides an iterator for `XRobotsTag`.  
     - `mod.rs`: Combines and exposes the module’s functionality.  

4. **Encoding and Decoding:**  
   - Encoding and decoding are implemented using the `Header` trait, supporting CSV-style comma-separated values.  

### Questions and Feedback Requested:  

1. **Code Structure:**  
   - Is the location of the new module (`x_robots_tag`) appropriate?  
   - Does the filename structure and organization align with project conventions?  

2. **Implementation Design:**  
   - Are there any improvements or alternative approaches you’d suggest for the `Rule` and `Element` structs?  
   - Is the use of `Vec<Element>` for `XRobotsTag` suitable, or are there optimizations to consider?  

3. **Edge Cases and Standards:**
   - Are there additional rules, formats, or edge cases that should be covered?  

### Testing:

- For now, the module is not tested, I will add the tests after the code structure becomes stable.

I look forward to your feedback and suggestions for improvement. Thank you!


rama-http-types/src/lib.rs
```diff
@@ -131,6 +131,9 @@ pub mod header {
         "x-real-ip",
     ];
 
+    // non-std web-crawler info headers
+    static_header!["x-robots-tag",];
```
GlenDC, line 135:
```suggestion
    //
    // More infornation at
    // <https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Robots-Tag>.
    static_header!["x-robots-tag"];
```

rama-http/src/headers/x_robots_tag/element.rs
```diff
@@ -0,0 +1,46 @@
+use crate::headers::x_robots_tag::rule::Rule;
+use rama_core::error::OpaqueError;
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Element {
+    bot_name: Option<String>, // or `rama_ua::UserAgent` ???
```
GlenDC, line 8: You can copy https://github.com/hyperium/headers/blob/master/src/util/value_string.rs to https://github.com/plabayo/rama/tree/main/rama-http/src/headers/util and use that one here, that's also what the UserAgent typed header uses: https://docs.rs/headers/0.4.0/src/headers/common/user_agent.rs.html#43

rama-http/src/headers/x_robots_tag/element.rs
```diff
@@ -0,0 +1,46 @@
+use crate::headers::x_robots_tag::rule::Rule;
+use rama_core::error::OpaqueError;
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Element {
```
GlenDC, line 7: Depending how you structure it, this actually has to be either:

```
struct Element { bot_name: Option<HeaderValueString>, rules: Vec<Rule> }
```

or

```
enum Element {
    BotName(HeaderValueString),
    Rule(Rule),
}
```

Because when a botname is mentioned it applies to all rules that follow it, until another botname is mentioned

rama-http/src/headers/x_robots_tag/element.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+impl FromStr for Element {
+    type Err = OpaqueError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (bot_name, indexing_rule) = match Rule::from_str(s) {
+            Ok(rule) => (None, Ok(rule)),
+            Err(e) => match *s.split(":").map(str::trim).collect::<Vec<_>>().as_slice() {
```
GlenDC, line 27: You'll need to anyway modify the Element code, so this might dissapear, but as extra info, be careful with collecting stuff in a Vec, just for tmp reasons. Gives allocations for not much good. Instead better to split fixed into a tuple, or use the iterator directly .

rama-http/src/headers/x_robots_tag/rule.rs
```diff
@@ -0,0 +1,113 @@
...
+    NoFollow,
+    None,
+    NoSnippet,
+    IndexIfEmbedded,
+    MaxSnippet(u32),
+    MaxImagePreview(MaxImagePreviewSetting),
+    MaxVideoPreview(Option<u32>),
+    NoTranslate,
+    NoImageIndex,
+    UnavailableAfter(String), // "A date must be specified in a format such as RFC 822, RFC 850, or ISO 8601."
```
GlenDC, line 19: This best be validated and enforced in a typed manner :)

rama-http/src/headers/x_robots_tag/rule.rs
```diff
@@ -0,0 +1,113 @@
+use rama_core::error::OpaqueError;
+use std::convert::{TryFrom, TryInto};
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(super) enum Rule {
```
GlenDC, line 7: Don't forget about the custom ones: noai, noimageai, SPC

rama-http/src/headers/x_robots_tag/rule.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        s.split(":")
+            .map(str::trim)
+            .collect::<Vec<_>>()
+            .as_slice()
+            .try_into()
+    }
+}
+
+impl TryFrom<&[&str]> for Rule {
```
GlenDC, line 56: Don't think it's worth implementing a public trait from this. You can just make simple private functions from this, and I would instead of collecting it into a vec just use the iterator's next method. You can then have a private fn for 1 arg en one with 2 arg.

GlenDC (changes requested):
Not a bad start. Do take your time to complete it and add sufficient start. But great start indeed. Added a couple of remarks to help you with some more guidance.

hafihaf123:
I've attempted to implement your suggestions, but I still have a couple of uncertainties:

1. I've used the `OpaqueError` type almost everywhere an `Error` type is required. Is this acceptable, or should I consider a different approach?
2. I've implemented the `FromStr` trait for `Element`, expecting a string slice in the format: `"<bot_name:> indexing_rule, indexing_rule_with_value: value <...>"`. However, I'm unsure if this is the correct interpretation or the best approach.

Please note that this implementation is not yet complete, as the `XRobotsTag::decode()` function is not functional at this stage.

rama-http/src/headers/x_robots_tag/valid_date.rs
```diff
@@ -0,0 +1,65 @@
...
+    pub fn date(&self) -> &str {
+        &self.0
+    }
+
+    pub fn into_date(self) -> String {
+        self.0
+    }
+
+    pub fn is_valid(&self) -> bool {
+        let rfc_822 = r"^(Mon|Tue|Wed|Thu|Fri|Sat|Sun),\s(0[1-9]|[12]\d|3[01])\s(Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec)\s\d{2}\s([01]\d|2[0-4]):([0-5]\d|60):([0-5]\d|60)\s(UT|GMT|EST|EDT|CST|CDT|MST|MDT|PST|PDT|[+-]\d{4})$";
```
GlenDC, line 29: given that you are anyway storing it as a string, do we really care about validating these for now? And even if we do I wonder if this really needs a regexp. This is defintely a file that needs more work even if we want to go for this

rama-http/src/headers/x_robots_tag/valid_date.rs
```diff
@@ -0,0 +1,65 @@
...
+        let iso_8601 = r"^\d{4}-(0[1-9]|1[0-2])-(0[1-9]|[12]\d|3[01])\s([01]\d|2[0-4]):([0-5]\d|60):([0-5]\d|60).\d{3}$";
+
+        check_is_valid(rfc_822, self.date())
+            || check_is_valid(rfc_850, self.date())
+            || check_is_valid(iso_8601, self.date())
+    }
+}
+
+fn check_is_valid(re: &str, date: &str) -> bool {
+    Regex::new(re)
```
GlenDC, line 40: I use regexes as little as possible, but even if you use them, you'll want to have them as lazy statics (which you can now-adays do with the stuff that the std library offsets. In past you needed the lazy_static crate for that, but no longer.

GlenDC:
>     1. I've used the `OpaqueError` type almost everywhere an `Error` type is required. Is this acceptable, or should I consider a different approach?

Questions like this are answered by thinking about its typical usage. Usually the parsing from the raw http header to the typed header will happen in the background as part of layer or web extractor or the like. So while you could try to return something like a custom error with exposure of the kind of error that happened I don't see it all that being useful here, even if you manually decode. Because even if you know what specific error happened, the result is either way the same, it's an invalid header so either you are okay with it or you are not.

TLDR: opaque Error is fine here. If there is ever a strong reason to expose certain error variants, it can be requested by opening a new issue about it. But that's a discussion for then.

>     2. I've implemented the `FromStr` trait for `Element`, expecting a string slice in the format: `"<bot_name:> indexing_rule, indexing_rule_with_value: value <...>"`. However, I'm unsure if this is the correct interpretation or the best approach.

The answer is that you'll want to ensure you support this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Robots-Tag#syntax ; So purely looking at whether it's a valid implementation you would do by checking against that reference syntax. After that the question will be whether it's an optimal enough implementation, but let's first focussing on correctly implementing this.

rama-http/src/headers/x_robots_tag/valid_date.rs
hafihaf123: Is the ValidDate struct even relevant if it does not validate the String? I'm wondering if it wouldn't then make more sense to just revert it to a String in the `UnavailableAfter` enum variant.

GlenDC:
You might want to read up a bit on how other folks have done it, even in other languages. E.g. look at https://github.com/nielsole/xrobotstag/blob/4cd7d8885a3e26fda9dd1a4075663cd3b80617f0/parser.go#L97. There's a lot to like about that implementation, but of course not to be copied as-is, given there is also plenty not to like about it. But what I personally take away from an implementation like this is:

- I like the idea of just having a single `RobotTag`, which would look something like:

```
struct RobotsTag {
     bot_name: Option<HeaderValueString>,
     all: bool,
     no_follow: bool,
     unavailable_after: Option<chrono::DateTime<chrono::Utc>>, 
     // ...
     no_ai: bool,
     // ...
     custom_rules: Option<Vec<CustomRule>>,
}
```

CustomRule would be a private construct which can be as simple as

```
struct CustomRule {
      key: HeaderValueString,
      value: Option<HeaderValueString>,
}

- you are already doing pretty well on parsing, but really there is no need for regexes here, at least not us directly
- you can make use of the `chrono` crate and try the formats that are specified in the documentation, there are three possibilities: [RFC 822](https://datatracker.ietf.org/doc/html/rfc822), [RFC 850](https://datatracker.ietf.org/doc/html/rfc850), or ISO 8601. Using chrono makes that trivial.

As such the only things we really need to expose here are the `XRobotsTag` typed header and the `RobotsTag` as a single "element". So 2 structs.

Not sure if you are a fan of the builder pattern but I like it a lot for cases like this, as for each rule you can expose this method to `RobotsTag`

```
fn new() == Default::default
```

The default one would be basically an empty tag, not rendered at all. Which is okay as the `Headers` trait allows you to not encode anything in the encode step, which you would do in case all your tags would be empty.

```
fn all(&self) -> bool { self.all }
fn with_all(mut self, all: bool) -> Self { self.all = all; self }
fn set_all(&mut self, all: bool) ->  &mut Self { self.all = all; self }
```

This allows them to be constructed in any way. Feel free to define a private `macro_rules!` to easily set create this triplet for the bool variants, saves you from typing, and easy enough to do if you combine it with the `paste` crate already used in other parts of rama.

The custom rules can be constructed without exposing the `CustomRule` struct. And to get them you can just return an  opaque iterator with a tuple &str or w/e.

I think this is pretty elegant and does everything you wanted to do without even having to expose our internal representation of it. Keeps the structure also a lot more flat without really losing much, given the `RobotsTag` should be able to be represented pretty small. It's in the end just a bunch of booleans and options, with the value parts of the option being on the heap, so pretty great for this purpose.

---

TLDR: great job on your work so far. Seems you are getting pretty comfortable with the codebase and defintely know your way around rust. I hope my feedback and guidance above helps you out in finishing it. Feel free to push back on something you do not agree with or propose alternatives.

Goal is mostly to keep this thing as simple and minimal as possible, while still allowing people to also do custom stuff, but without exposing too much of our internal API as to give the flexibility to change stuff internally without breaking stuff. Not that I mind making breaking changes where needed, but even better if you can upgrade without needing to break.

**Comment by hafihaf123** [comment]
> This allows them to be constructed in any way. Feel free to define a private `macro_rules!` to easily set create this triplet for the bool variants, saves you from typing, and easy enough to do if you combine it with the `paste` crate already used in other parts of rama.

I was wondering to maybe add some sort of `#[derive(Builder)]` functionality to the builder to generate the boilerplate code but I am wondering whether it is even possible as I am quite new to rust macros. From what I've found, I would probably have needed to define it as a procedural macro and kept it in a separate crate. I found `rama-macros` in the code base, but in its documentation, it states 

> There are no more macros for Rama.

**Comment by GlenDC** [comment]
> I was wondering to maybe add some sort of `#[derive(Builder)]` functionality to the builder to generate the boilerplate code but

Out of scope for this PR and overkill for what you need here. A macro_rules has plenty to fill our needs here

If one day you want to learn to write and read proc macros you might enjoy https://www.manning.com/books/write-powerful-rust-macros

Could use some help in my venndb crate for a future release

**Comment by hafihaf123** [comment]
another thing, do you prefer a separate builder struct (example: `let robots_tag = RobotsTag::builder().noindex().max_video_preview(5).build()`) or building using setters (example: `let robots_tag = RobotsTag::new(); robots_tag.set_noindex(true).set_max_video_preview(5)` or `let robots_tag = RobotsTag::new().with_noindex(true).with_max_video_preview(5)`)

**Comment by GlenDC** [comment]
> another thing, do you prefer a separate builder struct (example: `let robots_tag = RobotsTag::builder().noindex().max_video_preview(5).build()`) or building using setters (example: `let robots_tag = RobotsTag::new(); robots_tag.set_noindex(true).set_max_video_preview(5)` or `let robots_tag = RobotsTag::new().with_noindex(true).with_max_video_preview(5)`)

The reason to introduce a separate builder struct, e.g. `RobotsTagBuilder` is because it allows you to when done right to at compile time make an impossible state. For example it would allow you if done right to make it that this does not compile:

```rust
RobotsTag::builder().build()
```

As at that point there is not yet anything set. In my proposal above I said as a shortcut you can just ignore empty robotstag and not write those. But if you do want to go this extra mile we can indeed prevent it at compile time, which is better so I would argue.

<https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=82444bb713185c3f09ceeb917cc4fd66> here is an exmaple playground for you to show what i mean with that

Reason why you would need a builder for that is because:

1. the most important reason: you avoid that you are polluting your otherwise generic-free struct with generics
2. less important is that it allows you to get around the name colission trick, as we otherwise have 3 methods `a` for the same struct, which is not allowed in rust

I'm totally okay with you building that out, good proposal of you, so why not.
Even if you do that I would still add setters to the regular struct as well as it allows, e.g. in a proxy setting, to still modify the struct itself. Another option is to be able to turn it back into a builder `e.g. `impl From` and perhaps also add `into_builder`. That works fine as well.

Btw note also in my example that I add both the "consume" variant (`mut self`) and the setter variant (`&mut self`) to the builder. Please do this as well if you make a builder. Because in some scenarios you want to consume while in another one it is not as easy.

in cases where you want to pass it immediately to a function or parent structure it is easier to just consume:

```
foo(Foo::builder().a(true).b(42).build())
```

but in cases where you work with conditionals it's more conveniant to have the setter _also_ available, e.g.:

```
let mut foo = Foo::builder().a(true);
if bar {
    foo.set_b(true);
}
```

Of course in the case of a bool this seems a bit silly, as you could just assign `bar` to set_b. But for cases where the property type is a bit more complex, e.g. an enum it becomes less easy so. As such, I usually provide both options.

GlenDC:
Hope you are doing @hafihaf123 , just checking in here with you, no pressure, take your time, there's no hurry. However in case you are stuck, have some questions or want to talk something through do let me know :) I'm here for you. But also fine if you are ok as-is and continue at your own pace on your own. All good, just want you to know this.

hafihaf123:
Sorry for the delay, lately I didn't have much time to code. Anyways, I have completely reworked the API based on your suggestions. Please let me know where it needs some improvements and what should I focus on next.

rama-http/Cargo.toml
```diff
@@ -30,6 +30,7 @@ async-compression = { workspace = true, features = [
 base64 = { workspace = true }
 bitflags = { workspace = true }
 bytes = { workspace = true }
+chrono = "0.4.39"
```
GlenDC, line 33: You need to add this to `./Cargo.toml` (the root file) and than use here `chrome = { workspace = true }`

rama-http/src/headers/x_robots_tag/mod.rs
```diff
@@ -0,0 +1,11 @@
+pub mod robots_tag;
```
GlenDC, line 1: normally you would make the modules private (e.g. `mod robots_tag`), unless you do not publicly expose the stuff here already. Breaking up in files like this is mostly just for the developer sanity that maintains x_robots_tag, not so much for the user as for them it probably looks like 1 API surface.

rama-http/src/headers/x_robots_tag/robots_tag.rs
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            pub fn [<set_ $field>](&mut self, [<$field>]: $type) -> &mut Self {
+                self.[<$field>] = Some([<$field>]);
+                self
+            }
+        }
+    };
+
+    ($field:ident, $type:ty, vec) => {
```
GlenDC, line 33: I don't think you need this, as pretty sure you only need this ever for `custom_rules`, so you can take it out of the macro_rules

rama-http/src/headers/x_robots_tag/robots_tag.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::headers::x_robots_tag::max_image_preview_setting::MaxImagePreviewSetting;
+use crate::headers::x_robots_tag::robots_tag_builder::RobotsTagBuilder;
+
+macro_rules! getter_setter {
+    ($field:ident, $type:ty) => {
+        paste::paste! {
+            pub fn [<$field>](&self) -> $type {
+                self.[<$field>]
+            }
+
```
GlenDC, line 12: can you also provide the `(mut self, ...) -> Self` variants? as an additional way to set the vaues

rama-http/src/headers/x_robots_tag/robots_tag.rs
```diff
@@ -0,0 +1,102 @@
...
+            bot_name,
+            ..Default::default()
+        }
+    }
+
+    pub fn builder() -> RobotsTagBuilder {
+        RobotsTagBuilder::new()
+    }
+
+    getter_setter!(bot_name, HeaderValueString, optional);
```
GlenDC, line 85: Why are we having a builder as well as a setter in the struct? At that point what additional guarantees or reasons do we have to even have the builder?

rama-http/src/headers/x_robots_tag/robots_tag_builder.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::headers::x_robots_tag::robots_tag::RobotsTag;
+
+macro_rules! builder_field {
+    ($field:ident, $type:ty) => {
+        paste::paste! {
+            pub fn [<$field>](mut self, [<$field>]: $type) -> Self {
+                self.0.[<set_ $field>]([<$field>]);
+                self
+            }
+
```
GlenDC, line 13: same here, you also might want to have the `(mut self, ...) -> Self` variant as an extra

GlenDC (changes requested):
Thx starts to look close to being closed. Besides my remarks on the code, it seems we are still missing the actual implementation of https://docs.rs/rama/latest/rama/http/headers/trait.Header.html, right?

hafihaf123:
Hello,

I would appreciate a review before proceeding with the implementation of the actual encoding and decoding functionalities. Additionally, I would welcome any guidance on the best approach to implement these.

My current idea is to follow a structure similar to the Via header, implementing the FromStr trait along with a custom CSV parsing function to handle bot names effectively. I would love to hear your thoughts on this approach.

Thank you!

Cargo.toml
```diff
@@ -254,7 +255,7 @@ rama-tls = { version = "0.2.0-alpha.7", path = "rama-tls", optional = true }
 rama-ua = { version = "0.2.0-alpha.7", path = "rama-ua", optional = true }
 rama-utils = { version = "0.2.0-alpha.7", path = "rama-utils" }
 serde_html_form = { workspace = true, optional = true }
-serde_json = { workspace = true, optional = true }
+    serde_json = { workspace = true, optional = true }
```
GlenDC, line 258:
```suggestion
serde_json = { workspace = true, optional = true }
```

rama-http/src/headers/x_robots_tag_components/max_image_preview_setting.rs
```diff
@@ -0,0 +1,35 @@
...
+            MaxImagePreviewSetting::Large => write!(f, "large"),
+        }
+    }
+}
+
+impl FromStr for MaxImagePreviewSetting {
+    type Err = OpaqueError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().trim() {
```
GlenDC, line 26: This (`.to_lowercase()`) allocates a new string on the heap for nothing.
What I usually do is this:

```rust
let s = s.trim();

if s.eq_ignore_ascii_case("none") { ... } else if s.eq_ignore_ascii_case("..") { ... } ... else { Err(...) }
```

bit longer to type but no allocations required.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,104 @@
...
+    unavailable_after: Option<ValidDate>, // "A date must be specified in a format such as RFC 822, RFC 850, or ISO 8601."
+    // custom rules
+    no_ai: bool,
+    no_image_ai: bool,
+    spc: bool,
+    custom_rules: Vec<CustomRule>,
+}
+
+impl RobotsTag {
+    pub(super) fn new() -> Self {
```
GlenDC, line 68: If you go for the builder pattern I would not go for a default here as there is anyway in principle not a sane default. This thing is only really valid once you have some tag. Even `new_with_bot_name` is not yet valid enough to build as you still do not have a tag.

rama-http/src/headers/x_robots_tag_components/robots_tag_builder.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl RobotsTagBuilder<()> {
+    pub(super) fn new() -> Self {
+        RobotsTagBuilder(())
+    }
+
+    pub(super) fn bot_name(
+        self,
+        bot_name: Option<HeaderValueString>,
+    ) -> RobotsTagBuilder<RobotsTag> {
```
GlenDC, line 39: this is still not enough to allow it to be build though as no tag is defined yet and AFAIK a robots-tag with just a bot name is non-sensical.

rama-http/src/headers/x_robots_tag_components/valid_date.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+}
+
+impl From<ValidDate> for DateTime<Utc> {
+    fn from(value: ValidDate) -> Self {
+        value.0
+    }
+}
+
+impl TryFrom<DateTime<Utc>> for ValidDate {
```
GlenDC, line 29: 1. if you have this, why still a constructor? Trait impls are always public (if the trait is public), so also makes the `pub(super)` restriction a bit silly
2. why is this fallible? Am I missing something? Can this not be just a normal `impl From<DateTime<Utc>>`?

GlenDC (changes requested):
> I would appreciate a review before proceeding with the implementation of the actual encoding and decoding functionalities.

Looks pretty good, a couple of small remarks but overall pretty good.

> Additionally, I would welcome any guidance on the best approach to implement these.
>
> My current idea is to follow a structure similar to the Via header, implementing the FromStr trait along with a custom CSV parsing function to handle bot names effectively. I would love to hear your thoughts on this approach.

Sounds good, my 2 cents if you wish is as follows:

* I think you can use the regular csv delim logic that's already there
* for each delimited value you could check if you could try to use <https://doc.rust-lang.org/std/primitive.str.html#method.split_once> to see whether or not a new bot name was defined, based on the key-value `:` separator

Mind though that you do want to add some sanity check there to ensure that you only allow starting a new RobotsTag if the previous tag had at least one value set (e.g. cheapest would be by keeping track of a counter or bool or w/e.

Other than that sounds like a sane idea.

Encoding is a lot easier so I guess you do understand that part.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
hafihaf123: The `new_with_bot_name` method was meant to be a convenience method for the `RobotsTagBuilder` to be able to construct an instance of the `RobotsTag` struct, as it has private fields. Is that OK or can it be achieved differently?

hafihaf123:
I have implemented the initial versions of the encode and decode functionalities. For parsing, I chose an iterator-based approach, where each iteration yields a RobotsTag representing a valid portion of the header for a single bot name. The iterator continues processing until the entire string has been parsed.

The implementation is still somewhat rough and may benefit from refinements in structure, error handling, or efficiency. I would appreciate your feedback on areas where the approach could be improved.

rama-http/src/headers/x_robots_tag.rs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    fn decode<'i, I>(values: &mut I) -> Result<Self, Error>
+    where
+        Self: Sized,
+        I: Iterator<Item = &'i HeaderValue>,
+    {
+        let mut elements = Vec::new();
+        for value in values {
+            let mut parser = Parser::new(value.to_str().map_err(|_| Error::invalid())?);
```
GlenDC, line 22: It's fine to swallow errors to turn them into other errors. However can you add at least an `inspect_error` and `tracing::debug` output them with a short description?:

Something like:

```
map_err(|err| {
     tracing::debug!(?err, "x-robots-tag header element decoding failure");
     Error::Invalid()
})?
```

rama-http/src/headers/x_robots_tag.rs
```diff
@@ -0,0 +1,58 @@
...
+            while let Some(result) = parser.next() {
```
GlenDC, line 23: you can do `for result in parser {`, works for any iterator

rama-http/src/headers/x_robots_tag_components/max_image_preview_setting.rs
```diff
@@ -0,0 +1,38 @@
...
+pub(super) enum MaxImagePreviewSetting {
+    None,
+    Standard,
+    Large,
+}
+
+impl std::fmt::Display for MaxImagePreviewSetting {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        match self {
+            MaxImagePreviewSetting::None => write!(f, "none"),
```
GlenDC, line 15: personally I would either turn these strings into constants or do the entire thing with a single custom private macro. Both approaches result in the same, that you only have to write these magic strings once.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,214 @@
...
+            custom_rules: vec![],
+        }
+    }
+
+    pub(super) fn add_custom_rule(&mut self, rule: CustomRule) -> &mut Self {
+        self.custom_rules.push(rule);
+        self
+    }
+
+    pub(super) fn builder() -> Builder {
```
GlenDC, line 98: why is this not public? is a pretty common thing to have public.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,214 @@
...
+                fields = rest;
+            }
+        }
+
+        let mut builder = RobotsTag::builder().bot_name(bot_name);
+
+        for field in fields.split(',') {
+            match builder.add_field(field) {
+                Ok(_) => {}
+                Err(e) if e.to_string().contains("not a valid robots tag field") => {
```
GlenDC, line 152: never ever ever string compare errors, use ideally static typed errors for this or in worst case use dynamic checks/casting, but no string checks

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,214 @@
...
+            || field_name.eq_ignore_ascii_case("noai")
+            || field_name.eq_ignore_ascii_case("noimageai")
+            || field_name.eq_ignore_ascii_case("spc")
+    }
+
+    pub(super) fn from_str(s: &str) -> Result<Option<Self>, OpaqueError> {
+        let mut bot_name = None;
+        let mut fields = s;
+
+        if let Some((bot_name_candidate, rest)) = s.split_once(':') {
```
GlenDC, line 137: - I feel this from_str does not belong here,because you can actually result in mutliple x-robots-tags as you do not know up front how many robots name will appear in the CSV list, right?
- the `is_valid_field_name` feels wasteful. I would think that if you do split `:` that you would check if it's one of the keys of the k-v pairs, and if not, use it as a bot name?

Perhaps the wastefulness/awkwardness here stams from the fact that you perhaps try to use the builder/struct directly, where-as you might want to work with a flatter more intermediate simple structure here and build up the elements as you need to complete them. But again, I would think this lives outside RobotsTag as it makes it a lot easier to create multiple tags where needed

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,214 @@
...
+        write_field!(self.max_image_preview(), "max-image-preview", optional);
+        write_field!(self.max_video_preview(), "max-video-preview", optional);
+        write_field!(self.no_translate(), "notranslate");
+        write_field!(self.no_image_index(), "noimageindex");
+        write_field!(self.unavailable_after(), "unavailable_after", optional);
+        write_field!(self.no_ai(), "noai");
+        write_field!(self.no_image_ai(), "noimageai");
+        write_field!(self.spc(), "spc");
+
+        writeln!(f)
```
GlenDC, line 212: why the new line?

rama-http/src/headers/x_robots_tag_components/robots_tag_components/builder.rs
```diff
@@ -0,0 +1,136 @@
...
+        bot_name: Option<HeaderValueString>,
+    ) -> Builder<RobotsTag> {
+        Builder {
+            content: RobotsTag::new_with_bot_name(bot_name),
+            valid: false,
+        }
+    }
+}
+
+impl Builder<RobotsTag> {
```
GlenDC, line 51: The beautiful thing about builder pattern in Rust is that you can make a state machine with it such that you can never ever require building into a Result type for simple cases like this, as you can prevent that an invalid usage would not compile in the first place.

E.g. someting like this:

```
pub struct NoTag {
     name: Option<HeaderValueString>,
}

impl builder<NoTag> {
   // methods to set name, just sets name (as `&mut Self` and `mut Self`), these keep the builder type unchanged

   // also methods to set one of the possible tags -> these change it to `builder<RobotsTag>`
}

impl builder<RobotsTag> {
    // methods for robot name as well as tags
    // and also allow build as at this stage we can only ever get if a tag was set :)
}
```

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
```diff
@@ -0,0 +1,67 @@
...
+use std::str::FromStr;
+
+pub(in crate::headers) struct Parser<'a> {
+    remaining: Option<&'a str>,
+}
+
+impl<'a> Parser<'a> {
+    pub(in crate::headers) fn new(remaining: &'a str) -> Self {
+        Self {
+            remaining: Some(remaining.trim()),
```
GlenDC, line 13: might want to first check if empty and use `None` if empty

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
```diff
@@ -0,0 +1,67 @@
...
+            let field = field.trim();
+            if field.is_empty() {
+                continue;
+            }
+
+            match builder.add_field(field) {
+                Ok(_) => {
+                    remaining = rest.trim();
+                }
+                Err(e) if e.to_string().contains("not a valid robots tag field") => {
```
GlenDC, line 41: same same, do not ever check strings in errors...

rama-http/src/headers/x_robots_tag_components/valid_date.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+impl AsMut<DateTime<Utc>> for ValidDate {
+    fn as_mut(&mut self) -> &mut DateTime<Utc> {
+        &mut self.0
+    }
+}
+
+impl FromStr for ValidDate {
```
GlenDC, line 42: Please add unit tests for this, both for valid cases and invalid cases

GlenDC (changes requested):
Looks good. Left some comments that hopefully help you to get further.
Thanks for all the work done so far.

Can you please also go through older comments and please press "resolve/close" for any comment that you already resolved.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
hafihaf123: Thanks for pointing this out. Actually, this was the original implementation which has been moved to `Parser.next()`, but I forgot to delete the original code

rama-http/src/headers/x_robots_tag_components/valid_date.rs
hafihaf123: tests showed a problem with the parsing from rfc850, as this format uses timezone abbreviations but chrono cannot parse them into a fixed offset as they are not unambiguous (an example from chrono documentation: _CDT can mean either Central Daylight Time (North America) or China Daylight Time_)

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
hafihaf123: Actually, I have been struggling with the visibility specifiers. Since I am uncertain about what should be accessible externally, I have restricted access to everything except `XRobotsTag` as much as possible for now. I'd appreciate your input on this.

GlenDC: Builder should be a visible (public) API yes, as it's what you would wanat to use if you _send_ a request.

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
```diff
@@ -0,0 +1,87 @@
...
+                }
+                Err(e) => return Some(Err(e)),
+            }
+        }
+
+        Some(Ok(builder.build()))
+    }
+}
+
+impl Parser<'_> {
```
GlenDC, line 67: I'm still a bit puzzled why you decided to write the parser like this.
Tell me what you think about this kind of logic:

```pseudocode
next(&mut self) -> Option<Result<RobotsTag, OpaqueError>> {
     if self.buffer.is_empty() { return None; }
     Some(self.parse_next())
   }
}

parse_next(&mut self) -> Result<RobotsTag, OpaqueError> {
   let mut tag_count = 0;
   for {
       let mut tag = RobotsTag { /* invalid all default values , non-public to do this */ };
       match self.buffer.find(&[':', ',']) {
           Some(index) if self.buffer[index] == ':' => {
                   let key_buffer = &self.buffer[..index];
                   self.buffer = &self.buffer[index+1..];
                   let index = self.buffer.find(',').unwrap_or_else(|| self.buffer.len());
                   let value_buffer = &self.buffer[..index];
                   if key_buffer.eq_ignore_ascii("kv_key_1") {
                          parse_key1_value(value_buffer, &mut tag)?;
                          self.buffer = self.buffer[index+1..];
                          tag_count += 1;
                   } else if key_buffer.eq_ignore_ascii("kv_key_2") {
                          parse_key2_value(value_buffer, &mut tag)?;
                          self.buffer = self.buffer[index+1..];
                          tag_count += 1;
                   } else if ... {
                          ...
                   } else {
                          if tag_count > 0 {
                              return Ok(tag);
                          } else {
                              builder.name = Some(value_buffer.into()); // somehow
                               self.buffer = self.buffer[index+1..];
                          }
                   }
           }
           Some(index) if self.buffer[index] == ',' => {
                   let value_buffer = &self.buffer[..index];
                   let buffer = &self.buffer[index+1..];
                   parse_value(self.buffer, &mut tag)?;
                   tag_count += 1;
           }
           _ => {
               parse_value(self.buffer, &mut tag)?;
               self.buffer = "";
               tag_count += 1;
       }
       if self.buffer.is_empty() {
             if self.tag == 0 {
                  return Err(OpqueError::from_display("no tags found in robots tag"));
             } else {
                  return Ok(tag);
             }
       }
}
```

^ that's more or less how I would do it, as it's a pretty linear process, and you ever ever have to compare the same substring once given what something can be is pretty specific to that context.

rama-http/src/headers/x_robots_tag_components/valid_date.rs
GlenDC: Hmm in case they do have a specific non-ambigious meaning for rfc850 we either have to make our own custom mapping, or else just do not allow the rfc850 for now until someone requests.

GlenDC (changes requested):
Once again great progress! Thank you for keeping up with this one!

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
hafihaf123: I like how `parse_next()` is split from `next()`, making the return values more explicit and readable - especially since `None` is only returned when `self.buffer` is empty after trimming.

One thing I’m curious about is the use of `self.buffer.find()` instead of `split_once()`. Since `split_once()` already extracts key-value pairs cleanly, it might simplify some of the slicing logic. My implementation also relies on `builder.add_field()`, which avoids redundant substring comparisons.

Is there a specific reason you preferred `find()` here?

Another thing I don't understand is the `for` loop in the `parse_next()` function. Is it even possible to use `for {}` like that and does this approach have any advantages over `loop {}` ?

GlenDC: It was pseudo code, so yeah I guess in Rust it's loop. My pseudo code isn't tested or anything so I'm pretty certain I also screwed up some indices / slices. All has to be cleaner written an tested.

Problem with split once is that you wouldn't know whether it is a comma or a colon, as it doesn't return the separator, only the two parts, with `find` you can check on the index.

The problem is that you still anyway need to check in your current logic with `is_valid_field_name` which shouldn't be needed given you do the checking as part of the parsing. Builder won't howeve work for decoding, you would just directly work with an otherwise invalid RobotsTag immediately. Builder is only useful when hand-constructing a RobotsTag (e.g. as a client).

Your key-value pair parser is also decoupled from the robots-name, while really they are just the same, as you can for now just fallback as a robots name if the key part of the key-value pair is unknown. It will either way fail later in case the value wasn't a valid tag itself. So all good there even in case of someone making a mistake.

So taken that all together, that's why I proposed the above, it just keeps it all a lot more linear and simple, and only ever needing to check once if something or not.

rama-http/src/headers/x_robots_tag_components/valid_date.rs
hafihaf123: I checked the RFC 850 documentation, and it doesn't define a specific, unambiguous meaning for timezone abbreviations, leaving it to implementors. Given this, should we define custom mappings, block RFC 850 entirely, or detect it and return a custom message suggesting users open an issue if support is needed?

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
hafihaf123: Thank you for your response. Could you please also clarify the visibility specifiers for the other structures and modules related to XRobotsTag? Additionally, what general rule should be followed to determine their appropriate visibility levels?

rama-http/src/headers/x_robots_tag_components/valid_date.rs
hafihaf123: I have found this crate: [timezone-abbreviations](https://github.com/terhechte/timezone-abbreviations). I wonder whether it could be used to simplify the rfc850 implementation. If yes, can it be added as a dependency to `rama-http` or do I just copy the relevant source file content?

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
hafihaf123: What about structures such as `ValidDate`, `CustomRule` or `MaxImagePreviewSetting`? The way the builder is written now, when the user wants to set fields of their type, they need to construct a new intance of them. Should I keep them private and add another macro rule that will build them using the FromStr trait or just make them public?

rama-http/src/headers/x_robots_tag_components/valid_date.rs
GlenDC: No thank you for looking into it. For now I would really just do it with something like `DateTime::parse_from_str` using the knowledge above. If users want more fancy than that they can open an issue in rama for it.

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
GlenDC: - `ValidDate`: I would not expose that. The builder can just accept a `DateTime` object for that immediately. 
- `CustomRule` I would also not expose as a struct, but you can just ensure that the builder can create a single custom tag (`HeaderValueString` as param), or a key-value tag (two `HeaderValueString` as params).
- MaxImagePreviewSetting yes that enum you can expose as public :)

hafihaf123: Regarding `CustomRule`, what do I return from the getter of `custom_rules`? As some custom rules might consist of 2 `HeaderValueString`*s*, some only from one `HeaderValueString`, it seems to be the most appropriate to return them coupled inside the `CustomRule` structure. The what about the functions to add a custom rule? Should there still be 2 different ones accepting `HeaderValueString`s even if the CustomRule struct is exposed?

GlenDC: There's no reason to have to expose `CustomRule` for getter either though. You can just return `impl Iterator<Item = (HeaderValueString, Option<HeaderValueString>)>`

rama-http/src/headers/x_robots_tag_components/valid_date.rs
```diff
@@ -11,6 +11,220 @@ impl ValidDate {
...
+            NaiveDateTime::parse_and_remainder(s, "%A, %d-%b-%y %T")
+                .with_context(|| "failed to parse naive datetime")?;
+
+        let fixed_offset = Self::offset_from_abbreviation(remainder)?;
+
+        Ok(DateTime::from_naive_utc_and_offset(naive_date_time, fixed_offset))
+    }
+
+    fn offset_from_abbreviation(remainder: &str) -> Result<FixedOffset, OpaqueError> {
+        Ok(match remainder.trim() {
```
GlenDC, line 26: If you go this approach you might want to have a lazy hashmap instead, so we have O(1) lookup instead of this massive match statement, e.g. something like:

```
static TIMEZONE_MAP: OnceLock<HashMap<&'static str, &'static str>> = OnceLock::new();

fn get_timezone_map() -> &'static HashMap<&'static str, &'static str> {
    TIMEZONE_MAP.get_or_init(|| {
        let mut map = HashMap::new();
        map.insert("ACDT", "+1030");
        map.insert("ACST", "+0930");
        map.insert("ACT", "−0500");
        map.insert("ACWST", "+0845");
        map.insert("ADT", "−0300");
        map.insert("AEDT", "+1100");
        map.insert("AEST", "+1000");
        map.insert("AFT", "+0430");
        map.insert("AKDT", "−0800");
        map.insert("AKST", "−0900");
        // ... (continue for all time zones)
        map
    })
}
```

rama-http/src/headers/x_robots_tag_components/robots_tag_components/builder.rs
```diff
@@ -0,0 +1,219 @@
+use crate::headers::util::value_string::HeaderValueString;
+use crate::headers::x_robots_tag_components::{MaxImagePreviewSetting, RobotsTag, ValidDate};
+use chrono::{DateTime, Utc};
+use headers::Error;
+use rama_core::error::OpaqueError;
+
+macro_rules! robots_tag_builder_field {
```
hafihaf123, line 7: I was wondering whether it would make more sense to add a custom rule for `bool` fields, as it would make more sense to just be able to set them to true without having to specify (maybe then also add an `toggle_off` function or something like that). Also, it would prevent building an invalid `RobotsTag` by setting a field to `false` (`RobotsTag::builder().bot_name(None).no_snippet(false).build()`). Please let me know what you think about that.

P. S.
It would still be possible to do something like `RobotsTag::builder().bot_name(None).no_follow().toggle_off_no_follow()` to achieve the same thing, but I feel like it is more explicit.

GlenDC: I think for now, to keep it simple:

- ✅ `no_follow()` instead of `no_follow(true)`
- 🚫 let's not add the `toggle_off` versions

Also I would perhaps while you're at it not allow `bot_name(None)` but only use it with a value if you want to set it with bot name. Such that the usage would be

```
RobotsTag::builder().bot_name(MyBot).no_follow()
```

if you want to specify it for a bot or


```
RobotsTag::builder().no_follow()
```

to specify without a bot

hafihaf123: What about the setters on a constructed `RobotsTag`? Should they also be changed for `bool` values or kept as-is?

GlenDC: I think the setters should be gone from RobotsTag, gonna be easier if that's just read-only

hafihaf123: The builder was relying on the setters to modify the `RobotsTag` structure inside it, how do I change that? Do I keep the setters but restrict their visibility or do I modify the visibility of the fields or how else could it be done?

GlenDC: You can pub(super) the properties of RobotsTag

hafihaf123:
I have rewritten the parser logic, but I have kept using the `RobotsTag::builder()` in it because it already encapsulated the logic for adding a field from a string and also ensures the constructed `RobotsTag` is valid.

rama-http/src/headers/x_robots_tag_components/custom_rule.rs
```diff
@@ -0,0 +1,28 @@
+use crate::headers::util::value_string::HeaderValueString;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(super) struct CustomRule {
+    key: HeaderValueString,
+    value: Option<HeaderValueString>,
+}
+
+impl CustomRule {
+    pub(super) fn as_tuple(&self) -> (&HeaderValueString, &Option<HeaderValueString>) {
```
GlenDC, line 10:
```suggestion
    pub(super) fn as_tuple(&self) -> (&HeaderValueString, Option<&HeaderValueString>) {
```

Usually you move the reference inwards. For this to happen you'll need to change in the line below `&self.value` to `self.value.as_ref()`

rama-http/src/headers/x_robots_tag_components/max_image_preview_setting.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// The maximum size of an image preview for this page in a search results.
+/// If omitted, search engines may show an image preview of the default size.
+/// If you don't want search engines to use larger thumbnail images,
+/// specify a max-image-preview value of standard or none. [^source]
+///
+/// # Values
+///
+/// - `none`
+///     - No image preview is to be shown.
```
GlenDC, line 14: In rust you add these as doc comments above the variant, e.g.:

```
    ///  No image preview is to be shown.
   None,
```

rama-http/src/headers/x_robots_tag_components/robots_tag.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    pub fn builder() -> Builder<NoTag> {
+        Builder::new()
+    }
+
+    pub fn custom_rules(
+        &self,
+    ) -> impl Iterator<Item = (&HeaderValueString, &Option<HeaderValueString>)> {
```
GlenDC, line 78:
```suggestion
    ) -> impl Iterator<Item = (&HeaderValueString, Option<&HeaderValueString>)> {
```

rama-http/src/headers/x_robots_tag_components/valid_date.rs
```diff
@@ -0,0 +1,373 @@
...
+        .ok_or_else(|| OpaqueError::from_display(format!("invalid abbreviation: {}", remainder)))?
+        .parse()
+        .with_context(|| "failed to parse timezone abbreviation")
+}
+
+static TIMEZONE_MAP: OnceLock<HashMap<&'static str, &'static str>> = OnceLock::new();
+
+fn get_timezone_map() -> &'static HashMap<&'static str, &'static str> {
+    TIMEZONE_MAP.get_or_init(|| {
+        let mut map = HashMap::new();
```
GlenDC, line 91: code is gonna be cleaner and more efficient if you collect the hashmap from an array of tuples

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
```diff
@@ -0,0 +1,151 @@
...
+            }
+            Err(e) => Some(Err(e)),
+        }
+    }
+}
+
+impl Parser<'_> {
+    fn parse_first(remaining: &mut &str) -> Result<Builder<RobotsTag>, OpaqueError> {
+        match remaining.find(&[':', ',']) {
+            Some(index) if remaining.chars().nth(index) == Some(',') => {
```
GlenDC, line 62: you are using high level constructs here instead of `remaining[index]`, but below you anyway do `&remaining[index + 1..];`, might as well do it here already

I also think that that header is only ever expected to containing visible ASCII chars, you might wanna validate that first, somewhere prior to parsing, so that you can safely assume indexing and slicing in parse functions like this

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
```diff
@@ -0,0 +1,151 @@
...
+            Some(colon_index) if remaining.chars().nth(colon_index) == Some(':') => {
+                let (field, rest) = if let Some((before, after)) = remaining.split_once(',') {
+                    (before, after)
+                } else {
+                    (*remaining, "")
+                };
+
+                let builder = match RobotsTag::builder().add_field(field) {
+                    Ok(builder) => Ok(builder),
+                    Err(e) if e.is::<headers::Error>() => RobotsTag::builder()
```
GlenDC, line 76: Why is this error fallback needed? Can you high level explain what is happening here?

rama-http/src/headers/x_robots_tag_components/robots_tag_components/builder.rs
```diff
@@ -0,0 +1,268 @@
...
+    /// assert!(builder.add_field("max-snippet: 8").is_ok());
+    /// assert!(builder.add_field("nonexistent").is_err_and(|e| e.is::<headers::Error>()));
+    /// assert!(builder.add_field("max-video-preview: not_a_number").is_err_and(|e| e.is::<ParseIntError>()));
+    ///
+    /// let robots_tag = builder.build();
+    ///
+    /// assert_eq!(robots_tag.no_snippet(), true);
+    /// assert_eq!(robots_tag.max_snippet(), 8);
+    /// ```
+    pub fn add_field(&mut self, s: &str) -> Result<&mut Self, OpaqueError> {
```
GlenDC, line 215: why is this public? this is internal parser logic and I'm also still confused on how it like it is

GlenDC (commented):
I like how this Pr is progressing. All in all it's pretty closed to being merged. I do still have a feeling the parsing needs a bit work. Not sure if I'm making myself always clear in that are either. Would I need to take that last part over, or you rather finish that of yourself. I'm fine either way, as long as you're happy I'm good with it :)

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
hafihaf123: If your question is why am I checking whether the `Err` is of type `headers::Error`, that is because this is what the function would return when it can't parse the field value (it is an invalid field). That means that the part before the first colon is a `bot_name` specifier, and the rest should be able to be parsed into a valid field. The check is there because there could also be other errors when parsing the field, for example `ParseIntError` when parsing the number.

rama-http/src/headers/x_robots_tag_components/robots_tag_components/builder.rs
GlenDC: No best not, and I'm still also not convinced if that's the right abstraction even on a private level.

How I see it is that this is really just one big linear scan and whether something is a solo value, key-value pair or robot name is pretty clear from the context you find it in and the thing it is equal to.

Parsing I would definitely not expose outside of a TryFrom / FromStr / OtherTopLevelConstruct. We could if ever a good constructed feature requested is made for it of course, but not out of the box.

rama-http/src/headers/x_robots_tag_components/robots_tag_components/parser.rs
GlenDC: Yeah but that tells me there is a wrong abstraction here. We shouldn't have to rely on an error for such a common scenario.

GlenDC:
> I have addressed the requested changes in the code. However, I am unsure about the expected structure or behavior of the parser. Could you provide more details or an example of how it should function?

Here is more or less how I would do the parsing, let me know what you think about it. I haven't tested it so you might need to fix some bits here and there. But I hope you get the gist. I also think it can use some more improvement beauty fixes, but my goal was mostly to get the idea I have in my head across to you:

```rust

#[derive(Debug)]
pub(crate) struct Parser<'a> {
    buffer: &'a [u8],
}

impl<'a> Parser<'a> {
    pub(crate) fn new(buffer: &'a [u8]) -> Self {
        Self { buffer }
    }
}

#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
enum Delimiter {
    Colon,
    Comma,
}

#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
enum State {
    MaxSnippet,
    MaxImagePreview,
    MaxVideoPreview,
    UnavailableAfter,
}

fn find_delimiter(buffer: &[u8]) -> Option<(usize, Delimiter)> {
    for (index, &b) in buffer.iter().enumerate() {
        match b {
            b':' => return Some((index, Delimiter::Colon)),
            b',' => return Some((index, Delimiter::Comma)),
            _ => {}
        }
    }
    None
}

fn trim_space(buffer: &[u8]) -> &[u8] {
    let mut start = 0;
    let mut end = buffer.len();
    while start < end && buffer[start] == b' ' {
        start += 1;
    }
    while start < end && buffer[end - 1] == b' ' {
        end -= 1;
    }
    &buffer[start..end]
}

impl Iterator for Parser<'_> {
    type Item = Result<RobotsTag, OpaqueError>;

    fn next(&mut self) -> Option<Self::Item> {
        if self.buffer.is_empty() {
            return None;
        }
    
        let mut tag_count = 0;
        let mut tag = RobotsTag::new_with_bot_name(None);
        let mut state = None;

        loop {

            match find_delimiter(self.buffer) {
                Some((index, Delimiter::Colon)) => {
                    if state.is_some() {
                        return Some(Err(OpaqueError::from_display(
                            "unexpected key: expected a value",
                        )));
                    }

                    let key_buffer = trim_space(&self.buffer[..index]);
                    if key_buffer.eq_ignore_ascii_case(b"max-snippet") {
                        state = Some(State::MaxSnippet);
                    } else if key_buffer.eq_ignore_ascii_case(b"max-image-preview") {
                        state = Some(State::MaxImagePreview);
                    } else if key_buffer.eq_ignore_ascii_case(b"max-video-preview") {
                        state = Some(State::MaxVideoPreview);
                    } else if key_buffer.eq_ignore_ascii_case(b"unavailable_after") {
                        state = Some(State::UnavailableAfter);
                    } else {
                        if tag.bot_name.is_some() {
                            if tag_count == 0 {
                                return Some(Err(OpaqueError::from_display(
                                    "unexpected user-agent: expected a key or value",
                                )));
                            } else {
                                return Some(Ok(tag));
                            }
                        } else {
                            return Some(
                                std::str::from_utf8(key_buffer)
                                    .context("key as utf-8 UA")
                                    .and_then(|s| {
                                        s.parse::<HeaderValueString>()
                                            .map_err(OpaqueError::from_std)
                                    })
                                    .map(|ua| {
                                        tag.bot_name = Some(ua);
                                        tag
                                    }),
                            );
                        }
                    }
                    self.buffer = &self.buffer[index + 1..];
                }
                Some((index, Delimiter::Comma)) => {
                    let value = match std::str::from_utf8(trim_space(&self.buffer[..index])) {
                        Ok(value) => value,
                        Err(err) => return Some(Err(err.context("interpret value as utf-8"))),
                    };
                    if let Err(e) = parse_value(value, state, &mut tag) {
                        return Some(Err(e));
                    }
                    tag_count += 1;
                    state = None;
                    self.buffer = &self.buffer[index + 1..];
                }
                None => {
                    let value = match std::str::from_utf8(trim_space(self.buffer)) {
                        Ok(value) => value,
                        Err(err) => return Some(Err(err.context("interpret remainder value as utf-8"))),
                    };
                    if let Err(e) = parse_value(value, state, &mut tag) {
                        return Some(Err(e));
                    }
                    tag_count += 1;
                    state = None;
                    self.buffer = &self.buffer[self.buffer.len()..];
                }
            }

            if self.buffer.is_empty() {
                if tag_count == 0 {
                    return None;
                } else {
                    return Some(Ok(tag));
                }
            }
        }
    }
}

fn parse_value(value: &str, state: Option<State>, tag: &mut RobotsTag) -> Result<(), OpaqueError> {
    match state {
        Some(State::MaxSnippet) => {
            tag.max_snippet = value.parse().context("parse max-snippet value")?;
        }
        Some(State::MaxImagePreview) => {
            tag.max_image_preview = Some(value.parse().context("parse max-image-preview value")?);
        }
        Some(State::MaxVideoPreview) => {
            tag.max_video_preview = Some(value.parse().context("parse max-video-preview value")?);
        }
        Some(State::UnavailableAfter) => {
            tag.unavailable_after = Some(value.parse().context("parse unavailable_after value")?);
        }
        None => {
            if value.eq_ignore_ascii_case("all") {
                tag.all = true;
            } else if value.eq_ignore_ascii_case("noindex") {
                tag.no_index = true;
            } else if value.eq_ignore_ascii_case("nofollow") {
                tag.no_follow = true;
            } else if value.eq_ignore_ascii_case("none") {
                tag.none = true;
            } else if value.eq_ignore_ascii_case("nosnippet") {
                tag.no_snippet = true;
            } else if value.eq_ignore_ascii_case("indexifembedded") {
                tag.index_if_embedded = true;
            } else if value.eq_ignore_ascii_case("notranslate") {
                tag.no_translate = true;
            } else if value.eq_ignore_ascii_case("noimageindex") {
                tag.no_image_index = true;
            } else if value.eq_ignore_ascii_case("noai") {
                tag.no_ai = true;
            } else if value.eq_ignore_ascii_case("noimageai") {
                tag.no_image_ai = true;
            } else if value.eq_ignore_ascii_case("spc") {
                tag.spc = true;
            } else {
                return Err(OpaqueError::from_display(
                    "unexpected value: expected a key or valid value",
                ));
            }
        }
    }

    Ok(())
}
```

Parsing like this is a nice linear process where at every stage it's pretty clear to what comes next. And if you have sufficient unit tests than it is pretty easy to get working without any bugs.

GlenDC:
@hafihaf123, would you like me to take over this PR, or do you have a plan for it?

* First, I’d sync it with the latest main.
* Then, continue the work.

Given your recent inactivity and the remaining work after all the great progress you’ve already made, I understand if you'd prefer someone else to finish it. If you do want to continue, just let me know, and I’ll be happy to assist. However, if we go that route, I’d just ask that we avoid letting this drag on any longer.

GlenDC:
I'll take over this PR, today. Thank you for your contribution so far @hafihaf123 . It will be acknowledged with your work as a commit in its own right.

GlenDC:
Superseded by https://github.com/plabayo/rama/pull/707, thank you @hafihaf123 for your hard work,
your work is set in stone in your squashed commit, which will be merged as-is. Thanks!

Issue #395, opened 2025-01-14
impl IntoResponse for Either*


GlenDC:
Impl https://docs.rs/rama/latest/rama/http/trait.IntoResponse.html for all the Either variants found in https://docs.rs/rama/latest/rama/combinators/index.html

You can make use of the `impl_either!` macro for this.

Please also add a unit test, one is sufficient, E.g. `Either<&'static str, ()>`

YHM404:
Hi, I would like to take on this issue.

GlenDC:
All yours @YHM404 , thanks for your continued efforts, greatly appreciated!

PR #397, opened 2025-01-19
feat(http-types): implement IntoResponse for Either types

YHM404:
Resolve https://github.com/plabayo/rama/issues/395

Add IntoResponse implementations for Either types from rama_core::combinators. This allows Either types to be used as responses in handlers.

Added unit tests to verify the implementations work correctly for:
- Either<&str, Vec<u8>>
- Either3<&str, Vec<u8>, StatusCode>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2316,6 +2316,7 @@ dependencies = [
  "mime_guess",
  "paste",
  "pin-project-lite",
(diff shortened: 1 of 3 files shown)


GlenDC (approved):
Flawless, thanks a lot for the swift resolution!

PR #402, opened 2025-01-26
use TryInto<SocketAddress> bound for bind/listen functions in rama

### LINKED ISSUE #400: use `TryInto<SocketAddress>` bound for bind/listen functions in rama
Currently (rama 0.2.0-alpha.6) we are using the `ToSocketAddrs` (https://docs.rs/tokio/latest/tokio/net/trait.ToSocketAddrs.html) trait for these functions:

- https://docs.rs/rama/latest/rama/http/server/struct.HttpServer.html#method.listen
- https://docs.rs/rama/latest/rama/http/server/struct.HttpServer.html#method.listen_with_state
- https://docs.rs/rama/latest/rama/tcp/server/struct.TcpListenerBuilder.html#method.bind
- https://docs.rs/rama/latest/rama/tcp/server/struct.TcpListener.html#method.bind

Problem with this is that it means we cannot for example pass our own types (such as `SocketAddress` ) to those functions, which feels weird.

Also implement `From<std::net::SocketAddr>`, `From<std::net::SocketAddrV4>` and `From<std::net::SocketAddrV6>` for `SocketADdress`., as seems we do not do that yet. Change also one or two examples (`in /examples/*`) to make use of a `SocketAddress` instead of a string, to show it is possible.

fa993:
Closes #400

### DIFF
--- a/rama-tcp/src/server/listener.rs
+++ b/rama-tcp/src/server/listener.rs
@@ -2,12 +2,13 @@ use rama_core::graceful::ShutdownGuard;
 use rama_core::rt::Executor;
 use rama_core::Context;
 use rama_core::Service;
+use rama_net::address::SocketAddress;
 use rama_net::stream::SocketInfo;
 use std::fmt;
 use std::pin::pin;
 use std::sync::Arc;
 use std::{io, net::SocketAddr};
-use tokio::net::{TcpListener as TokioTcpListener, TcpStream, ToSocketAddrs};
+use tokio::net::{TcpListener as TokioTcpListener, TcpStream};
 
 /// Builder for `TcpListener`.
 pub struct TcpListenerBuilder<S> {
@@ -93,8 +94,11 @@ where
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
-    pub async fn bind<A: ToSocketAddrs>(self, addr: A) -> io::Result<TcpListener<S>> {
-        let inner = TokioTcpListener::bind(addr).await?;
+    pub async fn bind<A: TryInto<SocketAddress, Error = io::Error>>(
+        self,
+        addr: A,
+    ) -> io::Result<TcpListener<S>> {
+        let inner = TokioTcpListener::bind(Into::<SocketAddr>::into(addr.try_into()?)).await?;
 
         if let Some(ttl) = self.ttl {
             inner.set_ttl(ttl)?;
@@ -149,7 +153,7 @@ impl TcpListener<()> {
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
-    pub async fn bind<A: ToSocketAddrs>(addr: A) -> io::Result<Self> {
+    pub async fn bind<A: TryInto<SocketAddress, Error = io::Error>>(addr: A) -> io::Result<Self> {
         TcpListenerBuilder::default().bind(addr).await
     }
 }
--- a/rama-tcp/src/server/listener.rs
+++ b/rama-tcp/src/server/listener.rs
@@ -1,3 +1,4 @@
+use rama_core::error::BoxError;
 use rama_core::graceful::ShutdownGuard;
 use rama_core::rt::Executor;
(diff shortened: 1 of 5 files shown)


fa993:
I actually wanted to ask a bit about the error types, I've marked them as io::Error in the bounds since the functions originally returned a result type with an io::Error but I can see that this is causing issues with the callers as not all TryInto are implemented with that.

rama-tcp/src/server/listener.rs
```diff
@@ -93,8 +94,11 @@ where
     /// Binding with a port number of 0 will request that the OS assigns a port
     /// to this listener. The port allocated can be queried via the `local_addr`
     /// method.
-    pub async fn bind<A: ToSocketAddrs>(self, addr: A) -> io::Result<TcpListener<S>> {
-        let inner = TokioTcpListener::bind(addr).await?;
+    pub async fn bind<A: TryInto<SocketAddress, Error = io::Error>>(
```
GlenDC, line 97:
```suggestion
    pub async fn bind<A: TryInto<SocketAddress, Error: Into<BoxError>>>(
```

Probably you want something like this,
while at the same time adapting `io::Result<TcpListener<S>>` to `Result<TcpListener<S>, BoxError>`.

It will also mean that you need to chain `.map_err(Into::into) to `.bind(addr).await` as to make it type-match the err.

rama-tcp/src/server/listener.rs
```diff
@@ -93,8 +94,11 @@ where
...
+        self,
+        addr: A,
+    ) -> io::Result<TcpListener<S>> {
+        let inner = TokioTcpListener::bind(Into::<SocketAddr>::into(addr.try_into()?)).await?;
```
GlenDC, line 101: For clarity I would rewrite this line as:

```rust
let socket_addr = addr.try_into()?;
let tokio_socket_addr: SocketAddr = socket_addr.into();
let inner = TokioTcpListener::bind(tokio_socket_addr).await?;
```

It's longer to write, but I think a lot more clear what's going on.

But feel free to do something even different, whatever the case I think the use of `Into::<SocketAddr>::into` is here adding unnecesary complexity.

GlenDC:
> I actually wanted to ask a bit about the error types, I've marked them as io::Error in the bounds since the functions originally returned a result type with an io::Error but I can see that this is causing issues with the callers as not all TryInto are implemented with that.

So in this case we do not control all the different possible types that can be turned into a `SocketAddress`, given we use the non-sealed `TryInto` (std) trait. As such it can also be really any error. If you wanted to go very statically typed you could make an enum like `enum Error<E> { AddrError(E), BindError(io::Error) }`, but only because you expect people to handle it separately. However, there's no reason for that given people can also downcast a `BoxError` into an `io::Error` in case they want to catch something specific there.

As such I would just turn these errors into a `BoxError` and get it over with. This makes it also a lot easier to add more underlying errors in future. Another approach could be in future to split up the address injection from the actual network socket binding. But for now this is probably ok as-is.

Hope this helps you out together with my comments on your code.

Thank you for the swift resolution by te way, greatly appreciated.

rama-tcp/src/server/listener.rs
fa993: Resolved in newest commit

rama-tcp/src/server/listener.rs
fa993: Resolved in newest commit

fa993:
LMK if anything else is required

GlenDC:
> LMK if anything else is required

Is it possible that you forget to push that "last" commit mentioned in the comments? I still see `io::Error` being used?

fa993:
> > LMK if anything else is required
> 
> Is it possible that you forget to push that "last" commit mentioned in the comments? I still see `io::Error` being used?

Yeah, sorry about that 😅

GlenDC (approved):
Great job, seems flawless. Thank you very much for this swift and great first contribution!

GlenDC (changes requested):
Once the tests pass regarding `&String -> SocketAddress` I'll re-evaluate if we can merge, which I think should be possible.

GlenDC (approved):
Thanks a lot for this extra mile. Great work. Take care!

Issue #414, opened 2025-02-13
Incorrect JA4H_c and JA4H_d when provided cookies have unsorted values


kriztalz:
The current JA4H calculation misses the case where a user sends multiple cookies and their order is not in a sorted order:

Original request:
```
curl --http1.1 -H "User-Agent: curl/7.81.0" \
     -H "Accept: */*" \
     -H "Cookie: sierra=echo;alpha=bravo;zulu=tango" \
     https://echo.ramaproxy.org
```

echo ja4h - `ge11cn030000_fe444ad14866_dca6ec9510fc_fd9084384323`
echo ja4h raw - `ge11cn030000_Host,User-Agent,Accept_sierra_sierra=echo;alpha=bravo;zulu=tango`

correct ja4h - `ge11cn030000_fe444ad14866_7efd8825dc5a_f0c5f5a36bc1`
correct ja4h raw - `ge11cn030000_Host,User-Agent,Accept_alpha,sierra,zulu_alpha=bravo,sierra=echo,zulu=tango`

GlenDC:
Thank you for reporting this bug @kriztalz and also for picking up to resolve it. Very much appreciated!

kriztalz:
My pleasure! It was a great way to experience rust development for the first time!

GlenDC:
Closed by @kriztalz in https://github.com/plabayo/rama/pull/415
Thanks a lot for the work, and very fast.

If this is your first rust work, great job especially, outstanding work.

PR #415, opened 2025-02-13
Fix cookie sorting for JA4H and ALPN for JA4

### LINKED ISSUE #413: Incorrect ALPN parsing for JA4 fingerprint
Currently, the ALPN value for the JA4 fingerprint is obtained by:

[https://github.com/plabayo/rama/blob/7cfff072c17ca45f9af5e25a27c1dbefbb45b976/rama-net/src/fingerprint/ja4/tls.rs#L138](url)
```
let alpn_0 = alpn_it.next().unwrap_or('0');
let alpn_1 = alpn_it.next().unwrap_or('0');
```

Resulting in the following ALPNs results:

- HTTP/1.1 => ht
- HTTP/2 => h2
- HTTP/3 => h3

This is incorrect, as noted in FoxIO's blog:

> JA4 also clearly shows the ALPN (Application-Layer Protocol Negotiation). This represents the protocol that the application wants to communicate in after the TLS negotiation is complete. “h2” = HTTP/2, “h1” = HTTP/1.1, “dt” = DNS-over-TLS, etc. A full list of possible ALPNs can be found [here](https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml). A “00” here denotes the lack of ALPN. Note that the presence of ALPN “h2” does not indicate a browser as many IoT devices communicate over HTTP/2. However, the lack of an ALPN may indicate that the client is not a web browser.

We need to take the first and last character of the ALPN to get the correct result.

kriztalz:
This PR fixes #413 and #414 

- Use first and last character for JA4 ALPNs
- Trim the cookies for correct sorting
- Additional tests / edits of the existing ones to make sure the functionality is well implemented.


GlenDC (approved):
Looks great. Thank you, amazing work as a first contributor!

GlenDC:
FYI @kriztalz you might want to fix your global `git` config on your developer machine. It seems the details you use to make this commit did not match your info on github, so github sadly doesn't recognize you as a contributor now, which is a shame as this is all yours.

anyway you do have my gratitude, if that counts for something!

PR #421, opened 2025-02-14
Added hclient hello header key in the echo api response

kriztalz:
Added a header key in the API response of the echo server. The aim of the original issue is sadly not achieved as the order of the JSON keys are alphabetically ordered but this additions does segment nonetheless.


GlenDC (approved):
Flawless, thanks!

PR #423, opened 2025-02-16
Rama web router

nagxsan:


GlenDC (changes requested):
Please make use of `matchit` and adapt the code accordingly.

rama-http/src/service/web/router.rs
```diff
@@ -100,7 +49,125 @@ where
...
+
+    async fn serve(
+        &self,
+        mut ctx: Context<State>,
+        req: Request<>,
+    ) -> Result<Self::Response, Self::Error> {
+        let uri_string = req.uri().to_string();
+        match &self.routes.at(uri_string.as_str()) {
+            Ok(matched) => {
+                let params: HashMap<String, String> = matched.params.clone().iter().map(|(k, v)| (k.to_string(), v.to_string())).collect();
```
GlenDC, line 160: This will need to be collected as https://docs.rs/rama/latest/rama/http/matcher/struct.UriParams.html to be compatible with the existing matcher object. Feel free to expand the logic of UriParams if you miss something.

E.g. might make code cleaner here if you implement `FromIterator` for `UriParams`, but that's more of an optional thing.

GlenDC: Also in general there should be no reason to double allocate.

1. first allocation is .clone()
2. than you allocate again by turning the pairs into owned pairs

Probably that first clone isn't needed

rama-http/src/service/web/router.rs
```diff
@@ -100,7 +49,125 @@ where
     /// create a new web router
     pub(crate) fn new() -> Self {
         Self {
-            routes: TrieNode::new(),
+            routes: MatchitRouter::new(),
+            not_found: Arc::new(
```
GlenDC, line 53: this can probably be made optional so you don't need to assign it if not set, and instead just fallback to NOT_FOUND anyway but on the spot

rama-http/src/service/web/router.rs
```diff
@@ -1,72 +1,20 @@
...
-            wildcard_child: self.wildcard_child.as_ref().map(|child| child.clone()),
-            handlers: self.handlers.clone(),
-            param_name: self.param_name.clone(),
-        }
-    }
-}
-
 pub struct Router<State> {
-    routes: TrieNode<State>
+    routes: MatchitRouter<HashMap<Method, Arc<BoxService<State, Request, Response, Infallible>>>>,
```
GlenDC, line 16: We also need to be able to nest and merge btw, so I think this approach is a bit too simplistic

Issue #420, opened 2025-02-14
Should rama support legacy extensions / sigalgs


kriztalz:
`Ubuntu 22.04.5 LTS` ships with the following curl version in the apt repo:
`curl 7.81.0 (x86_64-pc-linux-gnu) libcurl/7.81.0 OpenSSL/3.0.2 zlib/1.2.11 brotli/1.0.9 zstd/1.4.8 libidn2/2.3.2 libpsl/0.21.0 (+libidn2/2.3.2) libssh/0.9.6/openssl/zlib nghttp2/1.43.0 librtmp/2.3 OpenLDAP/2.5.18`

This version uses a legacy experimental TLS extension called NPN (Next Protocol Negociation), which has now been replaced by ALPN. It also uses several non-typical signature algorithms such as: `SHA512 DSA`, `SHA384 DSA`, `SHA224 ECDSA`.

Such usage (only when TLS1.3 is **not** the minimum version) results in the following results:

```
        "data": [
          "ECDSA_NISTP256_SHA256 (0x0403)",
          "ECDSA_NISTP384_SHA384 (0x0503)",
          "ECDSA_NISTP521_SHA512 (0x0603)",
          "ED25519 (0x0807)",
          "ED448 (0x0808)",
          "RSA_PSS_PSS_SHA256 (0x0809)",
          "RSA_PSS_PSS_SHA384 (0x080a)",
          "RSA_PSS_PSS_SHA512 (0x080b)",
          "RSA_PSS_SHA256 (0x0804)",
          "RSA_PSS_SHA384 (0x0805)",
          "RSA_PSS_SHA512 (0x0806)",
          "RSA_PKCS1_SHA256 (0x0401)",
          "RSA_PKCS1_SHA384 (0x0501)",
          "RSA_PKCS1_SHA512 (0x0601)",
          "Unknown (0x0303)",
          "Unknown (0x0301)",
          "Unknown (0x0302)",
          "Unknown (0x0402)",
          "Unknown (0x0502)",
          "Unknown (0x0602)"
        ],
        "id": "SIGNATURE_ALGORITHMS (0x000d)"
```
Note the "Unknown"

and
```
      {
        "data": "0x",
        "id": "Unknown (0x3374)"
      },
```

I'd be of the opinion of adding those for completeness, although usage of NPN has been deprecated a while ago and has been deprecated in curl Deprecated since 7.86.0 (released October 26 2022 )

GlenDC:
[...]

PR #425, opened 2025-02-18
Adding legacy signature algorithms and NPN extension support

kriztalz:
Added the Next Protocol Negotiation TLS extension (always len 0 value)
Added signature algorithms that are not supported by BoringSSL (SHA224 DSA etc)

Closing #420


GlenDC (approved):

GlenDC:
Pretty obscure values, but they seem to be correct. Thank you for this contribution!

Issue #352, opened 2024-11-11
suport multiple basic/bearer credentials for 'Authorization' server support


GlenDC:
We already have support for this header for a single `Basic` / `Bearer` credential as can be found in <https://github.com/plabayo/rama/blob/main/rama-http/src/layer/auth/require_authorization.rs>.

Goal of this issue is to:

* provide also support for `[Basic; N]`, `Vec<Basic>`, `[Bearer; N]` and `Vec<Bearer>`;
* but also to support `Arc<...>` support

In order to support this all it is probably easiest to refactor the code a bit by:

* providing a "sealed" (private) trait "Authorizer" which is implemented for `Basic`, `Bearer`, `Vec<impl Authorizer>`, `[impl Authorizer; N]`, `Arc<Authorizer>`.

This way you only need a single implementation of the actual authorize/anonymouse logic as the actual authorization validation can be done by the `Authorizer` trait.

This issue is low priority, but due to a contribution from someone recently I noticed that we don't have support for multiple credentials here yet, while we do have that support for Proxy-Authorization. Mentorship is available for this one.

npalladium:
Hi @GlenDC, I'd like to take this one up. 

I'm new to the codebase and fairly new to Rust though.

GlenDC:
hi @npalladium . All yours.
As you are new to Rust you might want extra mentorship and help,
so please let me know how I can best guide you.

- Do you need a more verbose description?
- Do you need me to provide you with some starter code that you can then fill in and learn from like that?
- You have enough information as is?
- Something else?

Next to that, feel free to open a PR whenever you are stuck, need help, guidance or anything else. Ask here also questions or provide feedback as you see fit.

Welcome to the rama repo and I hope to see you around! :)

npalladium:
Hi @GlenDC. I think the requirements are clear. 

I'm looking at the adjacent code right now. I hope to have a draft PR ready sometime tomorrow. I'll let you know in the PR or on this issue if I need any help.

GlenDC:
Sounds great. Do feel free to open the PR even if it doesn't compile, in case you have questions, feedback or in need of help in general. Thanks for your valued time and contribution!

GlenDC:
Hi @npalladium I wanted to check in with you to see how you are doing.
Are you still interested in resolving this ticket or do you prefer to hand it over?
Either way it's okay, and I hope you're doing ok.

If there's guidance you need or have questions, do feel free to ask.

npalladium:
[...]

PR #432, opened 2025-02-25
support multiple basic/bearer credentials for 'Authorization' server support

NOOMA-42:
Issue related: #352


rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -304,12 +353,43 @@ pub struct AuthorizeContext<C> {
 }
 
 impl<C> AuthorizeContext<C> {
-    pub(crate) fn new(credential: C) -> Self {
```
GlenDC, line 307: Why do you make this `pub`? Seems you only use it in tests. Can this be reverted to `pub(crate)` again?

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -267,34 +268,82 @@ impl<ResBody> fmt::Debug for Basic<ResBody> {
     }
 }
 
-impl<S, B, ResBody> ValidateRequest<S, B> for AuthorizeContext<Basic<ResBody>>
+/// Trait for authorizing requests.
+pub trait Authorizer: Send + Sync + 'static {
+    type ResBody: Default + Send + 'static;
```
GlenDC, line 273: I don't think this associated type is required or desired? Can you try to leave it out of here?

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -267,34 +268,82 @@ impl<ResBody> fmt::Debug for Basic<ResBody> {
     }
 }
 
-impl<S, B, ResBody> ValidateRequest<S, B> for AuthorizeContext<Basic<ResBody>>
+/// Trait for authorizing requests.
+pub trait Authorizer: Send + Sync + 'static {
```
GlenDC, line 272: Can we also please make this sealed by doing the actual logic in a private trait. e.g. something like:

```
pub trait Authorizer: sealed::AuthSeal {}

impl ...

mod sealed {
      pub(super) trait AuthSeal: Send + Sync + 'static {
             ... the actual methods
      }

      impl ...
}
```

This way nobody else can call its logic or impl it.

GlenDC: As this is a pretty internal solution that is not really made for others to build upon. for anything more advanced you probably want to make your own Rama layer.

rama-http/src/layer/auth/require_authorization.rs
NOOMA-42: do you mean keep ResBody in this ValidateRequest function and still implement Basic<ResBody> instead of Authorizer? 
```rust
impl<S, B, ResBody> ValidateRequest<S, B> for AuthorizeContext<Basic<ResBody>>
```
otherwise this implementation without an associated type in Authorizer will cause an error: the type parameter ResBody is not constrained by the impl trait, self type, or predicates unconstrained type parameter

```rust
impl<S, B, C, ResBody> ValidateRequest<S, B> for AuthorizeContext<C>
where
    C: Authorizer,
    B: Send + 'static,
    S: Clone + Send + Sync + 'static,
    ResBody: Default + Send + 'static,
{
```

GlenDC: You are correct. It's because that entire validate logic needs some rethinking... Can you add a comment above 

```
impl<S, B, C, ResBody> ValidateRequest<S, B> for AuthorizeContext<C>
```

Something like:

```
// TODO: revisit ValidateRequest and related types so we do not require
// the associated Response types for all these traits. E.g. by forcing
// downstream users that their response bodies can be turned into the standard `rama::http::Body`
```

For the time being however @NOOMA-42 keep your associated type in `AuthorizeContext`, I realize now why you had to do that, and it brought back memories of writing this code in the past. So all good. No need to change it for now as that would be out of scope for this PR's intent.

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -267,43 +278,108 @@ impl<ResBody> fmt::Debug for Basic<ResBody> {
...
+    impl<T, const N: usize> AuthorizerSeal for [T; N]
+    where
+        T: AuthorizerSeal,
+    {
+        fn is_valid(&self, header_value: &HeaderValue) -> bool {
+            self.iter().any(|auth| auth.is_valid(header_value))
+        }
+
+        fn www_authenticate_header() -> Option<HeaderValue> {
+            None
```
GlenDC, line 323:
```suggestion
            T::www_authenticate_header
```

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -267,43 +278,108 @@ impl<ResBody> fmt::Debug for Basic<ResBody> {
...
+    impl<T> AuthorizerSeal for Vec<T>
+    where
+        T: AuthorizerSeal,
+    {
+        fn is_valid(&self, header_value: &HeaderValue) -> bool {
+            self.iter().any(|auth| auth.is_valid(header_value))
+        }
+
+        fn www_authenticate_header() -> Option<HeaderValue> {
+            None
```
GlenDC, line 336:
```suggestion
            T::www_authenticate_header()
```

rama-http/src/layer/auth/require_authorization.rs
```diff
@@ -468,8 +544,116 @@ mod tests {
...
+                format!("Basic {}", BASE64.encode("user:pass")),
+            )
+            .body(Body::default())
+            .unwrap();
+        let response = service.serve(Context::default(), request).await.unwrap();
+        assert_eq!(StatusCode::OK, response.status());
+    }
+
+    #[tokio::test]
+    async fn conversion_methods() {
```
GlenDC, line 627: what's the point of this `conversion_methods` test? Seems to me just to test if something like a array or vec is correct? Which I am sure the standard library already tests?

NOOMA-42: The original intention is to test array, vec, arc all work well. I think this is duplicate and remove it

GlenDC (approved):
LGTM, thanks a lot for this contribution!

Issue #434, opened 2025-02-26
Decouple network stack from types and implementation


soundofspace:
While working on #107 I noticed that `EstablishedClientConnection` includes the socket used for the connection. While in most cases a connection will indeed use a socket, it also doesn't in a lot of other cases. Some examples are:
- Tests specific layers/services that never go over the network
- Unix pipes
- Files
- Generic connections eg component to each other...

I also remember this being a problem in some other places of the codebase, but will mention these later here when I come across them again.
I believe that types and services should know/need as little as possible from other layers. This makes them very easy to test and this also aligns with how the internet was designed (mostly): layers stacked on top of each other and that can be swapped however you want.

For addr here specifically I see two options currently without doing more research:
- Put it in context as extension similar to negotiated tls params
- Put in in connections wrapper type for layers that actually make this socket (eg `TcpConnector`) and deref everything to the nested connection by default to make it work transparently
- (make addr optional, but this feels like a bandaid solution)

Right now I prefer the ctx approach, but I might be missing some crucial details here

GlenDC:
Hi @soundofspace , thank you for raising this issue. You bring home some valid points.
I've reviewed the rama codebase and seems like we do not really care about this socketAddress, if I get you right what you would want is that `EstablishClientConnection` becomes like this:

```rust
/// The established connection to a server returned for the http client to be used.
pub struct EstablishedClientConnection<S, State, Request> {
    /// The [`Context`] of the `Request` for which a connection was established.
    pub ctx: Context<State>,
    /// The `Request` for which a connection was established.
    pub req: Request,
    /// The established connection stream/service/... to the server.
    pub conn: S,
}
```

This seems to be possible with no friction, and 95% of the time we are just passing this to the next layer.
The only place where I could find it being used was in `rama-haproxy/src/client/layer.rs`

However that is no friction either as you can quite simply, as far as I can see, to change such haprproxy impl signatures like:

```
impl<S, P, State, Request> Service<State, Request> for HaProxyService<S, P, version::One>
where
    S: ConnectorService<State, Request, Connection: Stream + Unpin, Error: Into<BoxError>>,
    P: Send + 'static,
    State: Clone + Send + Sync + 'static,
    Request: Send + 'static,
{
```

Instead of `Connection: Stream + Unpin` to define it like `Connection: Stream + Socket + Unpin`

Where Socket is <https://docs.rs/rama/latest/rama/net/stream/trait.Socket.html>

[...]

Issue #107, opened 2024-03-19
implement connection pool


GlenDC:
Can be used in the static state, from start or in a wrapper state,
and is used to request or create connections based on:

- host
- http version
- `Option<UA>`
- secure (bool)

Internal implementation doesn't have to be the best, especially given these pools are short lived.

GlenDC:
First step is to probably study the one of hyper 0.x and see what lessons we can learn from it.

GlenDC:
Moving this to 0.3, as I don't think I really need it for now. We can think about it in 0.3

GlenDC:
Moving this back to 0.2. Will be combined in the client layer. Pooling is done by checking if conn/dispatch is ready, instead of keeping a connection within a the layer, it will request a ready one.

GlenDC:
`ChillFish8` at tokio's discord on the `#hyper` channel offered a basic design for such a pool:

```
/// A dispatcher that gives active upstream connections to clients
/// that need it.
///
/// Connections are provided on a [QueueMode] priority.
pub struct ConnectionPoolDispatcher<C> {
    mode: QueueMode,
    notify: Arc<Notify>,
    available: flume::Receiver<C>,
    returner: flume::Sender<C>,
}

impl<C> Clone for ConnectionPoolDispatcher<C> {
    fn clone(&self) -> Self {
        Self {
            mode: self.mode,
            notify: self.notify.clone(),
            available: self.available.clone(),
            returner: self.returner.clone(),
        }
    }
}

impl<C: Connection> ConnectionPoolDispatcher<C> {
    /// Create a new pool dispatcher with the given [QueueMode].
    pub fn new(mode: QueueMode) -> Self {
        let (tx, rx) = flume::unbounded();

        Self {
            mode,
            notify: Arc::new(Notify::new()),
            available: rx,
            returner: tx,
        }
    }

    /// Acquire a connection from the pool.
    ///
    /// If no connection is immediately available, this will yield
    /// until a new connection is available.
    pub async fn acquire(&self) -> C {
        loop {
            while let Ok(conn) = self.available.try_recv() {
                if !conn.is_closed() {
                    return conn;
                }
            }

            self.notify.notified().await;

            let conn = self.available.recv_async().await.unwrap();

            if !conn.is_closed() {
                return conn;
            }
        }
    }

    /// Release the held connection back to the pool.
    pub fn release(&self, conn: C) {
        if self.returner.try_send(conn).is_ok() {
            match self.mode {
[...]

PR #435, opened 2025-02-27
Decouple network stack from types: remove addr from EstablishConnection type

soundofspace:


GlenDC:
@soundofspace can you descripte what the next steps/part will be.

I'll review later today.

soundofspace:
I have seen other places that are also tightly coupled to the network stack. In the near future I will go through the codebase and update issue #434  with other such places and then also refactor them in a followup pr. This pr just fixes this exact issue as it is a blocker for #107

GlenDC:
> I have seen other places that are also tightly coupled to the network stack. In the near future I will go through the codebase and update issue #434 with other such places and then also refactor them in a followup pr. This pr just fixes this exact issue as it is a blocker for #107

Unless it's directly to the same thing i would honestly make a new issue for it. It's not like we pay per unique issue or so.

GlenDC (approved):

PR #441, opened 2025-03-04
Add connection pool and connector svc

soundofspace:


rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+    }
+}
+
+impl<C: AsyncWrite + Unpin> AsyncWrite for LeasedConnection<C> {
+    fn poll_write(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &[u8],
+    ) -> std::task::Poll<Result<usize, std::io::Error>> {
+        Pin::new(&mut self).poll_write(cx, buf)
```
GlenDC, line 128: this isn't what you want, instead use ` pin_project_lite::pin_project!` where you declare `LeasedConnection`, so you `#[pin]` the inner connection.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+    fn connections(&self) -> Result<MutexGuard<'_, VecDeque<PooledConnection<C>>>, OpaqueError> {
+        self.connections
+            .lock()
+            .map_err(|_| OpaqueError::from_display("failed to lock connections"))
+    }
+}
+
+impl<C> Debug for PoolInner<C> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let connection_hashes: Vec<String> = self
```
GlenDC, line 213: Why do we need to go through all the effort of collecting stuff in a debug fmt call?

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+use crate::stream::Socket;
+use rama_core::error::{BoxError, ErrorContext, OpaqueError};
+use rama_core::{Context, Service};
+use std::collections::VecDeque;
+use std::fmt::Debug;
+use std::num::NonZeroU16;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
+use std::sync::{Arc, Weak};
+use std::sync::{Mutex, MutexGuard};
```
GlenDC, line 11: you can use parking_lot::Mutex instead, it doesn't have the poisoning concept, so lock/unlock doesn't fail.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+        Ok(EstablishedClientConnection {
+            ctx,
+            req,
+            conn: leased_conn,
+        })
+    }
+}
+
+#[cfg(test)]
+
```
GlenDC, line 429:
```suggestion
```

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+        let active_slot = ActiveSlot(active_permit);
+        let pool = Arc::<PoolInner<C>>::downgrade(&self.inner);
+
+        // Check if we can reuse stored connection
+        let pooled_conn = {
+            let mut connections = self.inner.connections()?;
+            connections
+                .iter()
+                .position(|stored| stored.hash == hash)
```
GlenDC, line 319: In the end the only thing you really require it seems is that hash is `PartialEq` from what I can see. Perhaps also some other stuff for debugging e.g. fmt::Debug.

However might awnt to make that a generic so it doesn't have to be a string, as one might as well just use a number, some rich ID, a tuple or anything else. As long as it can be compared :)

GlenDC: I would btw not call it hash but instead `ConnectionId` or `ConnectionKey`, as the user isn't forced to make a hash here.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+impl<C> AsMut<C> for LeasedConnection<C> {
+    fn as_mut(&mut self) -> &mut C {
+        self
+    }
+}
+
+impl<C> Drop for LeasedConnection<C> {
+    fn drop(&mut self) {
+        if let (Some(pool), Some(pooled_conn)) = (self.pool.upgrade(), self.pooled_conn.take()) {
+            match pool.return_pooled_conn(pooled_conn) {
```
GlenDC, line 99: if let Err(err) = pool.return_pooled_conn(pooled_conn) {

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+    use rama_core::{Context, Service};
+    use std::{
+        convert::Infallible,
+        ops::Add,
+        sync::{Arc, Mutex},
+    };
+    use tokio_test::assert_err;
+
+    struct TestService {
+        pub created_connection: Arc<Mutex<u32>>,
```
GlenDC, line 442: Why not just use an AtomicU32?

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+    impl ReqToConnHasher<String> for StringRequestHasher {
+        fn hash(&self, req: String) -> (String, String) {
+            let count = req.chars().count().to_string();
+            return (req, count);
+        }
+    }
+
+    #[tokio::test]
+    async fn test_should_reuse_connections() {
+        let pool = Pool::new(NonZeroU16::new(1).unwrap(), NonZeroU16::new(1).unwrap()).unwrap();
```
GlenDC, line 485: might want to provide somewhere in rama-utils a macro or something else to make this a bit more user-friendly to use, e.g.:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=cc2e04792fef1d5599a216c58a5a6590

Only really required for literals as in the case you deserialize it (e.g. for a config) than it is already checked when deserializing.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,577 @@
...
+        ) -> Result<Self::Response, Self::Error> {
+            let mut counter = self.created_connection.lock().expect("mutex not poisened");
+            *counter = counter.add(1);
+
+            let conn = vec![];
+            Ok(EstablishedClientConnection { ctx, req, conn })
+        }
+    }
+
+    struct StringRequestHasher;
```
GlenDC, line 474: name can be better, e.g. to indicate that this is about CharCount

GlenDC (changes requested):
Excellent start, added my initial review for you. Based on how that progresses we might be ready for a final review already.

rama-net/src/client/pool.rs
soundofspace: Pin project doesn't work with a custom drop implementation but using this instead should also be fine?
```
Pin::new(&mut self.deref_mut()).poll_write(cx, buf)

Pin::new(self.deref_mut().as_mut()).poll_write(cx, buf)

```

rama-net/src/client/pool.rs
soundofspace: We don't but this is really really useful to debug why a pool is doing what it is doing, any other ways you propose?

rama-net/src/client/pool.rs
soundofspace: Honestly because I didn't but a lot of thought in optimizing unit tests

rama-net/src/client/pool.rs
soundofspace: Nice, good to know

rama-net/src/client/pool.rs
GlenDC: Sure. Makes sense. Thanks for the clarification.

rama-net/src/client/pool.rs
soundofspace: Yea will improve it, kinda same reason as `Arc<Mutex<u32>>,` in tests. Wanted to get this PR ready for review and went over tests fast

rama-net/src/client/pool.rs
GlenDC: Hmmm, guess it's fine if it's useful.

still:

1. you can probably just go by reference here instead of needing to clone the hash
2. you could implement formatter for something that wraps an iterator, so you do not need to collect it first in a vec

a bit of effort but that would probably make this as cheap as is expected from a debug call

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,592 @@
...
+struct ActiveSlot(OwnedSemaphorePermit);
+
+#[expect(dead_code)]
+#[derive(Debug)]
+/// Pool slot is needed to add a connection to the pool. Poolslots have
+/// a one to one mapping to connections inside the pool, and are used
+/// to track the 'total' connections inside the pool
+struct PoolSlot(OwnedSemaphorePermit);
+
+#[derive(Debug)]
```
GlenDC, line 31: derive Debug manually please., impl only if all generics are debug. As this derive enforced that they have to be debug, which is a needless restriction.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,592 @@
...
+struct PooledConnection<C, ConnId> {
+    /// Actual raw connection that is stored in pool
+    conn: C,
+    /// ID is not unique but is used to group connections that can be used for the same request
+    id: ConnId,
+    /// Slot this connection takes up in the pool
+    slot: PoolSlot,
+}
+
+#[derive(Debug)]
```
GlenDC, line 44: impl Debug for structs with generics manually please., impl only if all generics are debug. As this derive enforced that they have to be debug, which is a needless restriction.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,592 @@
...
+
+/// [`ReqToConnId`] is used to convert a `Request` to a connection ID. These IDs are
+/// not unique and multiple connections can have the same ID. IDs are used to filter
+/// which connections can be used for a specific Request in a way that is indepent of
+/// what a Request is.
+pub trait ReqToConnId<Request>: Sized + Send + Sync + 'static
+where
+    Request: Send + 'static,
+{
+    type ConnId: Send + Sync + PartialEq + Clone;
```
GlenDC, line 347: You porbably want to require `'static` as well, at least in the current version of rama I think it will help

GlenDC (commented):
Looks good. Can you also as part of this PR try to remove deadpool in the rama-cli fp code. Or what is your opinion on that? Seems to me like an excellent practical use case that also is a first good demonstration of this code in a real setting?

rama-net/src/client/pool.rs
soundofspace: ah yea forgot about that

soundofspace:
> Looks good. Can you also as part of this PR try to remove deadpool in the rama-cli fp code. Or what is your opinion on that? Seems to me like an excellent practical use case that also is a first good demonstration of this code in a real setting?

Not sure if this belongs in this PR, removing deadpool is one thing, but rama-cli also uses deadpool_postgres. These two combined actually do quite a lot of logic

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+struct ActiveSlot(OwnedSemaphorePermit);
+
+#[expect(dead_code)]
+#[derive(Debug)]
+/// Pool slot is needed to add a connection to the pool. Poolslots have
+/// a one to one mapping to connections inside the pool, and are used
+/// to track the 'total' connections inside the pool
+struct PoolSlot(OwnedSemaphorePermit);
+
+/// A connection which is stored in a pool. A hash is used to determine
```
GlenDC, line 32: check comments, some still seem to mention "hash"

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+            .field("pool", &self.pool)
+            .field("_slot", &self._slot)
+            .finish()
+    }
+}
+
+impl<C, ConnId> LeasedConnection<C, ConnId> {
+    /// Take ownership of the internal connection. This will remove it from the pool.
+    pub fn take(mut self) -> C {
+        let conn = self.pooled_conn.take().expect("only None after drop").conn;
```
GlenDC, line 79: it is pretty unconventional to assign it to a local variable to immediately return it. I also do not see it simplifying any logic here

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+    }
+}
+
+impl<C: AsyncWrite + Unpin, ConnId: Unpin> AsyncWrite for LeasedConnection<C, ConnId> {
+    fn poll_write(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &[u8],
+    ) -> std::task::Poll<Result<usize, std::io::Error>> {
+        // let x = self.deref_mut().deref_mut();
```
GlenDC, line 145: to be removed?

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+        Pin::new(self.deref_mut().as_mut()).poll_flush(cx)
+    }
+
+    fn poll_shutdown(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), std::io::Error>> {
+        Pin::new(self.deref_mut().as_mut()).poll_shutdown(cx)
+    }
+}
```
GlenDC, line 162: you forgot provided methods:

- [poll_write_vectored](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html#method.poll_write_vectored)(
- [is_write_vectored](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWrite.html#method.is_write_vectored)(&self) -> [bool](https://doc.rust-lang.org/nightly/std/primitive.bool.html) { ... }

This is important to not forget as otherwise the provided version will be used which is not desired in case the inner connection actually provided its own version.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+    fn poll_read(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        buf: &mut tokio::io::ReadBuf<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        Pin::new(self.deref_mut().as_mut()).poll_read(cx, buf)
+    }
+}
+
+impl<State, Request, C: Service<State, Request>, ConnId: Send + Sync + 'static>
```
GlenDC, line 174: move these trait bounds to a `where` clause, it becomes unredable otherwise (IMHO) if you make the `impl<...>` part this long. For a simple (single) trait bound it's fine, otherwise better to a where clause, like is the case here.

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+
+impl<C, ConnId: Debug> Debug for ConnectionStore<C, ConnId> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_list()
+            .entries(self.iter().map(|item| &item.id))
+            .finish()
+    }
+}
+
+#[derive(Clone)]
```
GlenDC, line 233: impl manual please

GlenDC: Especially in this case :P you `Arc` it, but thanks to this derive you anyway do force `C` and `ConnId` to be clone, while that is clearly not a property ever used here

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,618 @@
...
+    ) -> PooledConnector<S, C, Request, R> {
+        PooledConnector {
+            inner,
+            pool,
+            req_to_conn_id,
+            wait_for_pool_timeout: None,
+        }
+    }
+
+    pub fn with_wait_for_pool_timeout(mut self, timeout: Option<Duration>) -> Self {
```
GlenDC, line 416: I tend to call these kind of things `maybe_with_x` and also provide a version `with_x` without the Option wrpaper, and even a `set_x` version for `&mut self` (without the option wrapper).

GlenDC (commented):
Mostly is looking good. I do agree that removing postgres-deadpool is probably out of scope for this PR.

Still, might be nice to add an example in `/examples` that is also tested.

rama-net/src/client/pool.rs
soundofspace: aha didnt know that

rama-net/src/client/pool.rs
soundofspace: Yea was working already on adding actual use case and noticed this. Something I tend to forget, same for default impl's

rama-http-backend/src/client/mod.rs
```diff
@@ -170,53 +180,187 @@ impl<I1, I2> EasyHttpWebClient<I1, I2> {
...
+            tls_config: self.tls_config,
+            proxy_tls_config: self.proxy_tls_config,
+            http_req_inspector_jit: self.http_req_inspector_jit,
+            http_req_inspector_svc: self.http_req_inspector_svc,
+            connection_pool: pool,
+        }
+    }
+
+    #[cfg(any(feature = "rustls", feature = "boring"))]
+    pub fn without_connection_pool(self) -> EasyHttpWebClient<I1, I2, ()> {
```
GlenDC, line 243: Given that is the default anyway, why make it a setter?

rama-http-backend/src/client/mod.rs
```diff
@@ -323,21 +467,22 @@ where
         // set the runtime http req inspector
         let connector = connector.with_svc_req_inspector(self.http_req_inspector_svc.clone());
 
-        // NOTE: stack might change request version based on connector data,
```
GlenDC, line 326: Did you remove this comment by accident?

rama-http-backend/src/client/mod.rs
```diff
@@ -170,53 +180,187 @@ impl<I1, I2> EasyHttpWebClient<I1, I2> {
...
+        connector: T,
+        ctx: Context<State>,
+        req: Request<Body>,
+    ) -> Result<Connection<C, State, Body>, OpaqueError>
+    where
+        T: ConnectorService<State, Request<Body>, Connection = C>,
+        T::Error: Send + 'static,
+    {
+        let pool = self.connection_pool.clone();
+        let connector = PooledConnector::new(connector, pool, BasicHttpConnId {});
```
GlenDC, line 351:
```suggestion
        let connector = PooledConnector::new(connector, pool, BasicHttpConnId);
```

rama-http-backend/src/client/mod.rs
```diff
@@ -170,53 +180,187 @@ impl<I1, I2> EasyHttpWebClient<I1, I2> {
...
+            http_req_inspector_jit: self.http_req_inspector_jit,
+            http_req_inspector_svc: self.http_req_inspector_svc,
+            connection_pool: (),
         }
     }
 }
 
-impl<State, BodyIn, BodyOut, I1, I2> Service<State, Request<BodyIn>> for EasyHttpWebClient<I1, I2>
+/// Map http request to unique connection id so we can use a connection pool
+struct BasicHttpConnId;
```
GlenDC, line 273:
```suggestion
#[derive(Debug)]
#[non_exhaustive]
struct BasicHttpConnId;
```

GlenDC (changes requested):
Couple of remarks but looks good in general.
One thing I do miss, what about the Context-based storage? It's fine if you do not want to implement that yet, but I would still at the very least like to hear a solid plan for this one?

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    async fn serve(
+        &self,
+        ctx: Context<State>,
+        req: Request,
+    ) -> Result<Self::Response, Self::Error> {
+        let conn_id = self.req_to_conn_id.id(&ctx, &req)?;
+
+        let pool = match ctx.get::<Pool<Storage>>() {
+            Some(pool) => &pool.clone(),
```
GlenDC, line 542: why do you clone this btw?

rama-net/src/client/pool.rs
```diff
@@ -0,0 +1,727 @@
...
+
+    fn id(&self, ctx: &Context<State>, request: &Request) -> Result<Self::ConnID, OpaqueError> {
+        self(ctx, request)
+    }
+}
+
+/// [`PooledConnector`] is a connector that will keep connections around in a local pool
+/// so they can be reused later. If no connections are available for a specifc `id`
+/// it will create a new one.
+pub struct PooledConnector<S, Storage, R> {
```
GlenDC, line 499: You also need to make a Layer struct for this one btw

examples/http_pooled_client.rs
```diff
@@ -0,0 +1,152 @@
...
+    setup_tracing();
+    tokio::spawn(async move {
+        run_server(ADDRESS).await;
+    });
+
+    // Give server time to start, we do this instead of retrying as we want all
+    // errors to be given back to us and never retried internally.
+    sleep(Duration::from_millis(10)).await;
+
+    let client = EasyHttpWebClient::default().with_connection_pool(Pool::default());
```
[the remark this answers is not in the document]
soundofspace: For now I did indeed, as this is a pretty big change and was worried how big the impact might be. Wanted to test this pool with real life workloads before making this a default

rama-http-backend/src/client/mod.rs
soundofspace: If in the future we change the default to use a connection pool or you get the client from somewhere else, you might want to disable this pool for some specific reason, this makes that possible.

rama-http-backend/src/client/mod.rs
soundofspace: yea

rama-net/src/client/pool.rs
soundofspace: Because later we move ctx when calling inner.serve but we still have it borrowed. Clone here should be cheap as it is an arc. Also tried this, same problem. Any alternatives without clone?
image

GlenDC: No you're right. However the more natural thing to write would be `ctx.get::<Pool<Storage>>().clone()`

soundofspace: That doesnt work here clone needs to be inside match

GlenDC (approved):

GlenDC:
Looks good @soundofspace once you resolve the conflicts we can merge this PR. I think it's worth revisiting this as part of 0.3, but for now it's more than good enough.

soundofspace:
> Looks good @soundofspace once you resolve the conflicts we can merge this PR. I think it's worth revisiting this as part of 0.3, but for now it's more than good enough.

Agreed, we should have a much better understanding of how we will use this pool by then, and what shortcomings the current design has.

Issue #440, opened 2025-03-03
make body limit for echo util service configurable instead of hardcoded 1 MiB


GlenDC:
https://github.com/plabayo/rama/blob/0abc1495acf91bd6a1eda72bf1662eee424a9db3/src/cli/service/echo.rs#L267

Useful for custom users that require larger payloads.

ikorason:
Hi @GlenDC I would like to work on this and learn at the same time 😃

GlenDC:
Closed in https://github.com/plabayo/rama/pull/444 by @ikorason . Thanks for the swift resolution. Have a good one!

PR #444, opened 2025-03-04
make body limit for echo util service configurable

ikorason:
Address #440


ikorason:
@GlenDC plz let me know if I'm doing this correctly?

GlenDC (approved):
Looks good, thx!

PR #463, opened 2025-03-18
add release action

### LINKED ISSUE #412: make cross-platform compilation (build.sh) work for rama-cli
`build.sh` was added in #410

It does however not work out of the box on macos.

Would be great it if can be fixed such that we can easily build for all these platforms and get the binaries for upload into a (github) release.
If needed it can be done via docker, but I believe with tooling like _zig_ it should be possible to do so without the need for docker.

The difficulty for now however is Cloudflare's boring-sys...

https://github.com/plabayo/rama/blob/main/rama-cli/scripts/build.sh

airycanon:
Closes #412

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ rustls = { version = "0.23", default-features = false, features = [
     "tls12",
     "aws_lc_rs",
 ] }
+aws-lc-sys = { version = "0.27", features = ["bindgen"] }
 derive_more = "2.0.1"
 rustls-native-certs = "0.8.0"
 rustls-pemfile = "2.1"
--- a/rama-net/Cargo.toml
+++ b/rama-net/Cargo.toml
@@ -17,11 +17,12 @@ workspace = true
 default = []
 http = ["dep:rama-http-types", "dep:sha2", "dep:itertools", "dep:hex"]
 tls = ["dep:hex", "dep:md5", "dep:sha2", "dep:itertools"]
-rustls = ["tls", "dep:rustls"]
+rustls = ["tls", "dep:rustls", "dep:aws-lc-sys"]
 boring = ["tls", "dep:boring", "dep:nom"]
 telemetry = ["rama-core/telemetry"]
 
 [dependencies]
+aws-lc-sys = { workspace = true, optional = true, features = ["bindgen"] }
 base64 = { workspace = true }
 boring = { workspace = true, optional = true }
 bytes = { workspace = true }
--- a/rama-net/Cargo.toml
+++ b/rama-net/Cargo.toml
@@ -22,7 +22,7 @@ boring = ["tls", "dep:boring", "dep:nom"]
 telemetry = ["rama-core/telemetry"]
 
 [dependencies]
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -264,6 +264,7 @@ serde_html_form = { workspace = true, optional = true }
 serde_json = { workspace = true, optional = true }
 tokio = { workspace = true, features = ["macros", "io-std"], optional = true }
 tracing = { workspace = true, optional = true }
+aws-lc-sys = { version = "0.27", features = ["bindgen"] }
```
GlenDC, line 267: I suppose you added it because we need the bindgen feature.

That said, this doesn't belong here.

1. add that dependency to the workspace dependencies instead
2. add it as a dependency (`workspace = true, features = ["bindgen"]`) to `rama-net` instead, and only when "rustls" is enabled, as it is not used for boring or if tls is not enabled at all

GlenDC (changes requested):
One minor request, other than that looks good for me.

GlenDC:
@soundofspace can you please provide feedback to here?

Cargo.toml
GlenDC: Once done you'll also have to sort dependencies btw:

```
cargo sort --workspace --grouped
```

If you do not have that cmd yet, you'll need to install it: https://crates.io/crates/cargo-sort

rama-net/Cargo.toml
```diff
@@ -18,10 +18,11 @@ default = []
 http = ["dep:rama-http-types", "dep:sha2", "dep:itertools", "dep:hex"]
 tls = ["dep:hex", "dep:md5", "dep:sha2", "dep:itertools"]
 rustls = ["tls", "dep:rustls"]
-boring = ["tls", "dep:boring", "dep:nom"]
+boring = ["tls", "dep:boring", "dep:nom", "dep:aws-lc-sys"]
```
GlenDC, line 21: Not for boring, but for `rustls`, boring doesn't use this crypto library it has its own crypto

rama-net/Cargo.toml
```diff
@@ -17,11 +17,12 @@ workspace = true
 default = []
 http = ["dep:rama-http-types", "dep:sha2", "dep:itertools", "dep:hex"]
 tls = ["dep:hex", "dep:md5", "dep:sha2", "dep:itertools"]
-rustls = ["tls", "dep:rustls"]
+rustls = ["tls", "dep:rustls", "dep:aws-lc-sys"]
 boring = ["tls", "dep:boring", "dep:nom"]
 telemetry = ["rama-core/telemetry"]
 
 [dependencies]
+aws-lc-sys = { workspace = true, optional = true, features = ["bindgen"] }
```
GlenDC, line 25:
```suggestion
aws-lc-sys = { workspace = true, optional = true }
```

feature is already enabled in workspace AFAIK

soundofspace:
Looks good, thanks for picking this up!

GlenDC (approved):

GlenDC:
Looks great @airycanon . Thank you very much for your contribution, stellar job!

PR #465, opened 2025-03-18
Issue 391/improve and refactor the crypto story in rama: rama-boring crate

soundofspace:
Goal of this PR:
- Crate for everything boring related. Right now only boring dependency but to be used for utilities and custom bindings in the future. (We might even completely replace current boring crate)
- Refactor everything to use this crate

### DIFF
--- a/rama-boring/Cargo.toml
+++ b/rama-boring/Cargo.toml
@@ -1,0 +1,29 @@
+[package]
+name = "rama-boring"
+description = "everything we need from boring is exposed here"
+version = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+repository = { workspace = true }
+keywords = ["io", "async", "boring", "network", "rama"]
+categories = ["asynchronous", "network-programming", "cryptography", "web-programming"]
+authors = { workspace = true }
+rust-version = { workspace = true }
+
+[lints]
+workspace = true
+
+[features]
+default = []
+
+[dependencies]
+boring = { workspace = true }
+
+[dev-dependencies]
+
+[package.metadata.cargo-public-api-crates]
+allowed = []
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/rama-boring/README.md
+++ b/rama-boring/README.md
@@ -1,0 +1,47 @@
+[![rama banner](../docs/img/rama_banner.jpeg)](https://ramaproxy.org/)
+
+[![Crates.io][crates-badge]][crates-url]
+[![Docs.rs][docs-badge]][docs-url]
+[![MIT License][license-mit-badge]][license-mit-url]
+[![Apache 2.0 License][license-apache-badge]][license-apache-url]
+[![rust version][rust-version-badge]][rust-version-url]
+[![Build Status][actions-badge]][actions-url]
+
+[![Discord][discord-badge]][discord-url]
+[![Buy Me A Coffee][bmac-badge]][bmac-url]
+[![GitHub Sponsors][ghs-badge]][ghs-url]
+[![Paypal Donation][paypal-badge]][paypal-url]
+
+[crates-badge]: https://img.shields.io/crates/v/rama-tls.svg
+[crates-url]: https://crates.io/crates/rama-tls
+[docs-badge]: https://img.shields.io/docsrs/rama-tls/latest
+[docs-url]: https://docs.rs/rama-tls/latest/rama_tls/index.html
+[license-mit-badge]: https://img.shields.io/badge/license-MIT-blue.svg
+[license-mit-url]: https://github.com/plabayo/rama/blob/main/LICENSE-MIT
+[license-apache-badge]: https://img.shields.io/badge/license-APACHE-blue.svg
+[license-apache-url]: https://github.com/plabayo/rama/blob/main/LICENSE-APACHE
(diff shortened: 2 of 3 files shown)


soundofspace:
@GlenDC do you agree with exporting boring crate like this, or would you export them at the module top?
```rust

pub mod dep {
    //! Dependencies for rama boring modules.
    //!
    //! Exported for your convenience (and so we centralize this).
    pub mod boring {
        //! Re-export of the [`boring`] crate.
        //!
        //! [`boring`]: https://docs.rs/boring
        #[doc(inline)]
        pub use boring::*;
    }
}
```

GlenDC:
> @GlenDC do you agree with exporting boring crate like this, or would you export them at the module top, asking before migrating everything?
> 
> ```rust
> pub mod dep {
>     //! Dependencies for rama boring modules.
>     //!
>     //! Exported for your convenience (and so we centralize this).
>     pub mod boring {
>         //! Re-export of the [`boring`] crate.
>         //!
>         //! [`boring`]: https://docs.rs/boring
>         #[doc(inline)]
>         pub use boring::*;
>     }
> }
> ```

I think it's the wrong question to ask. We wanna get eventually to a state where we own what are currently the crates `boring` and `boring-sys`. Where will these rama versions live? The answer to that question probably tells you on how you want to expose this already. Ideally you can get this reorgnised before we release 0.2, so that you can in a non-breaking fashion swap out those crates with our own version once you are ready for that work, still as part of `0.2.1` or `0.2.2` instead of having to wait for `0.3`

soundofspace:
If we already want to do this now, I will skip this step and integrate/fork broring directly in rama similar to the other forks in rama. When thinking about this more, it's probably less work then splitting it in two parts.

GlenDC:
> If we already want to do this now, I will skip this step and integrate/fork broring directly in rama similar to the other forks in rama. When thinking about this more, it's probably less work then splitting it in two parts.

It all depends on where you want to put the copied code. If both the source code of `boring` and `boring-sys` live somewhere under rama-tls-boring than this can still be a valid step. It all depend on such answers.

soundofspace:
The goal is to have everything boring related in `rama-boring`. This includes `boring` and `boring-sys`, but also the custom bindings we will provide and custom wrappers on top of `boring` APIs.

`rama-tls-boring` and `rama-crypto` will then both use this crate but will focus on business logic, respectively how do handle tls and how to expose crypto primitives. It will be a fine line to choose if something belongs in `rama-boring` or one of the more specialized crates. How I currently see it:
- Custom bindings ->  `rama-boring`
- Wrappers on top of `boring` that might be used in multiple places or just make the API a lot nicer to deal with  ->  `rama-boring`
- Tls connector / connections / clients / layers -> `rama-tls-boring`. If we miss stuff add new binding/APis in  `rama-boring` and consume them here
- Crypto primitives (in line with what we expose for other crypto providers) and nice to haves (eg converting between formats, quality of live shortcuts...) -> `rama-crypto`
- `rama-tls-boring` will use `rama-crypto` for crypto related logic instead of reinventing the wheel so it can really focus on tls itself (again fine line)

I'm still trying to figure out what the best / easiest way is to split this work in smaller chunks so this doesn't drag on forever and has to be rebased every time. Integrating `boring` and `boring-sys`  in rama directly (by copy pasting it) will be a none trivial task as it is quite a complex project, especially `boring-sys` which has all the fun stuff: cmake, submodules, and git patches.

GlenDC:
Sounds good. `tokio-boring` shouldn't be used though, let's just integrate that directly into `rama-tls-boring`. As it will be incompatible with our own dependencies.

Regarding `boring`. I think honestly that what you now call `rama-boring` will really have to be a separate repository and a direct fork of <https://github.com/cloudflare/boring>. It's a shame as it breaks the nice mono repo we have now, but it will be the sanest way. Not just speaking in regards to the rabit hole of submodules, but also the licensing of the google code is complex. And I do not want submodules in rama repo, so we'll have to just fork boring as rama-boring. Honestly might be for the best. We can just add the stuff we miss and try to stay as conservative as possible there. Anything high level of QoL can be added in a separate crate (e.g. `rama-boring-utils`) so we keep that diff as small as possible, ideally just stuff that we really just need because we need bindings or deep integrations with FFI, etc.).

I have the repo here already ready for you: https://github.com/plabayo/rama-boring, for your local tests you can just use `path imports by referencing it on disk`, I'll publish it perhaps together with v0.2.0-alpha.8 so that you can reference it for real already as well.

soundofspace:
@GlenDC goal of this PR:
- Crate for boring tls implementation: rama-tls-boring
- Move boring tls logic from rama-tls to rama-tls-boring
- Remove rama-tls as we no longer need it
- Move all boring tls types from rama-net to rama-tls-boring similar to https://github.com/plabayo/rama/pull/524
- Remove boring dependency from rama-net. This crate has now no dependencies on any specific tls implementation

What this PR doesn't do:
- Move "common" tls logic that is actually boring specific to boring tls crate and revise it how we implement and expose that
- Rethink how we stack tls configs

Those will be added in different PRs so we can iterate fast on this without creating huge merge conflicts

GlenDC (approved):
Awesome job. Thx for this splendid pr, so happy where this is going.

Couple of beauty remarks. Also do not forget to update the tls feature links in several markdowns as well as the site index page:

image

As these otherwise link to no longer used rama-tls crate.

GlenDC:
Once review is addressed you can merge it, please.

PR #466, opened 2025-03-18
Provide 'rama tls' cli cmd

### LINKED ISSUE #427: provide 'rama tls' cli cmd
Similar to the existing 'http' command in rama-cli, it would be great if we can also have a 'tls' command. In a first version would already be good enough if it would just allow one to establish a tls connection and immediately exit after that. It can be expanded further after that based on feature requests or internal needs.

Goal would mostly be so it can already be used for quick inspections where you want to see if a tls connection can be made with a remote/local tls acceptor, instead of the more traditional tools such as `openssl s_client`. Similar to the `http` command it would be able with the different log level modes to print more info or not, especially given as the underlying net/tls libraries already have tracing that will do so anyway.

This might require some extra work in the rama-tls code though so we can expose a lot more information about the server hello, etc. Client hello can probably already be printed, but also that is to be checked.

ikorason:
Closes #427

### DIFF
--- a/rama-cli/src/cmd/tls/mod.rs
+++ b/rama-cli/src/cmd/tls/mod.rs
@@ -7,7 +7,7 @@ use rama::{
         client::{ConnectorService, EstablishedClientConnection},
         tls::{
             DataEncoding,
-            client::{ClientConfig, NegotiatedTlsParameters},
+            client::{ClientConfig, NegotiatedTlsParameters, ServerVerifyMode},
         },
     },
     tcp::client::{Request, service::TcpConnector},
@@ -44,68 +44,52 @@ pub async fn run(cfg: CliCommandTls) -> Result<(), BoxError> {
         )
         .init();
 
-    // let authority = Authority::try_from(cfg.address.clone())?;
-
     let authority = match Authority::try_from(cfg.address.clone()) {
         Ok(authority) => authority,
-        Err(err) => {
-            // if missing port, we can try to add the default port
-            if err.to_string().contains("missing port") {
-                let authority = format!("{}:443", cfg.address);
-                Authority::try_from(authority)?
-            } else {
-                return Err(err.into());
-            }
+        // if missing port, we can try to add the default port
+        Err(err) if err.to_string().contains("missing port") => {
+            let authority = format!("{}:443", cfg.address);
+            Authority::try_from(authority)?
         }
+        Err(err) => return Err(err.into()),
     };
+
+    tracing::info!("Connecting to: {}", authority);
 
     let tls_client_data = TlsConnectorData::try_from(ClientConfig {
         store_server_certificate_chain: true,
+        server_verify_mode: if cfg.insecure {
+            Some(ServerVerifyMode::Disable)
+        } else {
+            Some(ServerVerifyMode::Auto)
+        },
         ..Default::default()
     })
     .expect("create tls connector data for client");
 
     let tcp_connector = TcpConnector::new();
-
     let loggin_service = LoggingLayer.layer(tcp_connector);
 
     let tls_connector = TlsConnectorLayer::secure()
         .with_connector_data(tls_client_data)
         .layer(loggin_service);
 
-    let ctx = Context::default();
-
-    let EstablishedClientConnection { ctx, req, conn } =
-        tls_connector.connect(ctx, Request::new(authority)).await?;
+    let EstablishedClientConnection {
+        ctx,
+        req: _,
+        conn: _,
+    } = tls_connector
+        .connect(Context::default(), Request::new(authority))
+        .await?;
 
     let params = ctx.get::<NegotiatedTlsParameters>().unwrap();
 
     if let Some(cert_chain) = params.peer_certificate_chain.clone() {
         match cert_chain {
-            DataEncoding::Der(raw_data) => match X509::from_der(&raw_data) {
-                Ok(cert) => {
-                    println!("Certificate: {:?}", cert);
-
-                    println!("Subject: {:?}", cert.subject_name());
-                    println!("Issuer: {:?}", cert.issuer_name());
-                }
-                Err(err) => {
-                    println!("Error decoding certificate: {:?}", err);
-                }
-            },
+            DataEncoding::Der(raw_data) => log_cert(&raw_data, 1),
             DataEncoding::DerStack(raw_data_list) => {
                 for (i, raw_data) in raw_data_list.iter().enumerate() {
-                    match X509::from_der(raw_data) {
-                        Ok(cert) => {
-                            println!("\nCertificate #{}: {:?}", i + 1, cert);
-
-                            println!("Subject: {:?}", cert.subject_name());
-                            println!("Issuer: {:?}", cert.issuer_name());
-                        }
-                        Err(err) => {
-                            println!("Error decoding certificate #{}: {:?}", i + 1, err);
-                        }
-                    }
+                    log_cert(raw_data, i + 1);
                 }
             }
             DataEncoding::Pem(raw_data) => {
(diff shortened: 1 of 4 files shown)


rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+/// rama tls support
+pub struct CliCommandTls {
+    /// The address to connect to
+    /// e.g. "example.com" or "example.com:8443"
+    /// if no port is provided, the default port 443 will be used
+    address: String,
+
+    #[arg(long, short = 'k')]
+    /// Wether to skip certificate verification
+    insecure: bool,
```
GlenDC, line 33: this value is ignored for now it seems, you'll want to use it for the `server_verify_mode` in https://docs.rs/rama/latest/rama/net/tls/client/struct.ClientConfig.html

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+    let ctx = Context::default();
+
+    let EstablishedClientConnection { ctx, req, conn } =
+        tls_connector.connect(ctx, Request::new(authority)).await?;
+
+    let params = ctx.get::<NegotiatedTlsParameters>().unwrap();
+
+    if let Some(cert_chain) = params.peer_certificate_chain.clone() {
+        match cert_chain {
+            DataEncoding::Der(raw_data) => match X509::from_der(&raw_data) {
```
GlenDC, line 85: you can simplify this code by returning the error with your context attached to it. There's no point in continuing if you have an error here already

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                Ok(cert) => {
+                    println!("Certificate: {:?}", cert);
+
+                    println!("Subject: {:?}", cert.subject_name());
+                    println!("Issuer: {:?}", cert.issuer_name());
+                }
+                Err(err) => {
+                    println!("Error decoding certificate: {:?}", err);
```
GlenDC, line 93:
```suggestion
                    eprintln!("Error decoding certificate: {:?}", err);
```

and please also keep track that erorrs have occurred such that you can return an error at the end as well to make sure we exit with a non-success status code due to errors that have happened

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,158 @@
...
+                for (i, raw_data) in raw_data_list.iter().enumerate() {
+                    match X509::from_der(raw_data) {
+                        Ok(cert) => {
+                            println!("\nCertificate #{}: {:?}", i + 1, cert);
+
+                            println!("Subject: {:?}", cert.subject_name());
+                            println!("Issuer: {:?}", cert.issuer_name());
+                        }
+                        Err(err) => {
+                            println!("Error decoding certificate #{}: {:?}", i + 1, err);
```
GlenDC, line 106:
```suggestion
                            eprintln!("Error decoding certificate #{}: {:?}", i + 1, err);
```

GlenDC (changes requested):
Thank you for the first draft of this PR, it looks great. Have a couple of remarks for you to resolve, once that is ok, can you please paste an output of you using that tool on both `example.com:80` as well as `example.com:443`?

ikorason:
@GlenDC since my last commit, after I ran these 2 commands `example.com:443` and `example.com:80`, the latter will exit with error.

image

Also I will update the error to keep track of them as you said.

rama-cli/src/cmd/tls/mod.rs
ikorason: Yes forgot to add it, plz check last commit

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        )
+        .init();
+
+    let authority = match Authority::try_from(cfg.address.clone()) {
+        Ok(authority) => authority,
+        // if missing port, we can try to add the default port
+        Err(err) if err.to_string().contains("missing port") => {
```
GlenDC, line 50: First, never rely on error strings. If you do ever need to match on an error we would need to have a specific error type you can check for via BoxError/stdError.

However I think you can keep it a lot more simpler. Something like:

```
let address = cfg.address.trim();
let authority: Authority = if cfg.address.contains(':') {
     address.parse().context("parse config address as authority")?
} else {
     let host: Host = address.parse().context("parse config address as host")?;
     Authority::new(host, 443)
};
```

Doesn't that do exactly what you want without having to rely on some specific errors even?

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    let tcp_connector = TcpConnector::new();
+    let loggin_service = LoggingLayer.layer(tcp_connector);
+
+    let tls_connector = TlsConnectorLayer::secure()
+        .with_connector_data(tls_client_data)
+        .layer(loggin_service);
+
+    let EstablishedClientConnection {
+        ctx,
+        req: _,
```
GlenDC, line 79: you can just do:

```
EstablishedClientConnection { ctx, .. }
```

Forgot if it's 2 dots or 3 dots, but either way that does exactly what you want without having to list all the properties you do not care about.

rama-cli/src/cmd/tls/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        conn: _,
+    } = tls_connector
+        .connect(Context::default(), Request::new(authority))
+        .await?;
+
+    let params = ctx.get::<NegotiatedTlsParameters>().unwrap();
```
GlenDC, line 85:
```suggestion
    let params = ctx.get::<NegotiatedTlsParameters>().expect("NegotiatedTlsParameters to be available in connector context");
```

GlenDC (changes requested):
Looks pretty much good to go. We could ship this, but I would like to give you the opportunity to resolve some last nitpicks that you might be able to learn from. Let me know if too much though and I can apply them myself after merging. Either way good job!

GlenDC:
> @GlenDC since my last commit, after I ran these 2 commands `example.com:443` and `example.com:80`, the latter will exit with error.
> 
> image
> 
> Also I will update the error to keep track of them as you said.

Awesome job, looks exactly what I had in mind. Is this PR ready for merge or is there a reason why this PR is still in draft mode? If it is ready according to you please mark it as "ready for review". Once the last comments are resolved I can immediately merge it in. But like said before, in case you would rather have me do those minor nitpicks, also cool for me, I mostly left them there so you might be able to learn from it but I would understand if you feel it's enough as-is. Either way cool, just let me know :)

ikorason:
@GlenDC thanks for the feedback, I'll update them as soon as possible later. Also question CI clippy is complaining about println! use, should I ignore them for now?

GlenDC:
> @GlenDC thanks for the feedback, I'll update them as soon as possible later. Also question CI clippy is complaining about println! use, should I ignore them for now?

You can add:

```
#![allow(clippy::print_stdout)]
```

to the top of your `tls.rs` file :)

rama-cli/src/cmd/tls/mod.rs
ikorason: Thank you, this is much better indeed, learned something new ;)

ikorason:
@GlenDC fixed up the comments, plz check, and thanks for the feedback, learned a bunch ;)

GlenDC (approved):
Looks great. Thank you for this contribution!

Issue #391, opened 2025-01-11
improve and refactor the Crypto story in rama


GlenDC:
This story would be about refactoring the code related to crypto and prepare the path for the crypto future of rama. Follow up stories can continue the work started here. There are two aspects to this:

1. Regards to tls, we drop the attempt of finding one unified API to rule them all and instead respect that the rustls integration has a different purpose than that of boring. For all the standard common use cases Rustls is fine. And so it wil have a nice working client connector and a nice transport layer-service to accept streams on server sides. But that's pretty much where that one ends. Boring will be used in case you need a tight grip on the controls, have all kind of special needs, want to do advanced stuff etc. In this story we'll still work with the cloudflare bindings but in a future we'll want to provide our own `rama-tls-boring` crate. As part of this task you can already create this crate (dummy-style, see rama-socks5 for an example) to already reserve the name next time we publish an alpha version.
2. Secondly, we affirm that crypto is more than just tls. So far we didn't have the luxary to think about that, but now with the http(s) stack shaping up we can think about that a bit more. Data verification, encryption and hashing is a thing and used for a lot more than just tls. E.g. JWT support is an example for that. As such it would make sense to perhaps have a `rama-crypto` crate where we would also house crypto primitives and expose crypto backends as well. Or perhaps we do not need such a crate and we could just expose `jwt` in future directly from within rama master crate somewhere.
  - we would need to think on how this relates to tls though. It's confusing in the crypto space in general. E.g. libs like boringssl and openssl are really focussed on SSL/TLS, but at the same time they also come with many related cryptographical algorithms and tools. This makes me not entirely certain what to do with crypto backends if we also have a `rama-crypto` crate. E.g. do we _also_ expose boring there? Or instead use `rama-tls-boring` if the boring feature is enabled? or perhaps only have it in `rama-crypto` and have `rama-tls-backend` use that? All choices feel weird in their own way.
  -  what is however more certain that it does make sense that we should have:
      - easy support using our backends already available for our tls support for: encryption, signing, hashing and random data
         (e.g. boring in case that's anyway available)
       - jwt support as it's pretty common for a lot of projects to need that, so it might make sense to expose that. That's not for this story. But I did start work on https://github.com/Keats/jsonwebtoken to have that crate be able to work with other crypto backends, which would allow us to embed that crate and re-export it, but reusing the same crypto backend (depending on the feature) that we already use for tls. Keeping the dep tree of the rama user small.

[...]

PR #467, opened 2025-03-19
Issue 391/improve and refactor the crypto story in rama: separate crate for rustls

soundofspace:
Goal of this PR:
- Split rustls of to separate crate
- Update everything to use this new crate
- Remove old logic in rama-tls/rustls
- Use rustls types directly and don't use wrapper types in rama-net/tls

Not part of this PR
- Split of boring
- Change APIs of rama-net/tls or rama-tls. This will be done later

Part of #391


docs/book/src/deploy/rama-cli.md
```diff
@@ -31,7 +31,7 @@ Options:
 The easiest way to install `rama` is by using `cargo`:
 
 ```sh
-cargo install rama-cli@0.2.0-alpha.8
+cargo install rama-cli@0.2.0-alpha.9
```
GlenDC, line 34: revert changes in this file please, the install instructions are always of the last released version, as the current main version is always the next (not yet published) one

examples/http_mitm_proxy.rs
```diff
@@ -60,18 +60,29 @@ use rama::{
...
+use rama::{
+    net::tls::{
+        client::{ClientConfig, ClientHelloExtension, ServerVerifyMode},
+        server::{ServerAuth, ServerConfig},
+    },
+    tls::boring::server::{TlsAcceptorData, TlsAcceptorLayer},
+};
+
+#[cfg(all(feature = "rustls", not(feature = "boring")))]
+use rama::tls_rustls::{
```
GlenDC, line 81: I would expose `rama-tls-rustls` as `rama::tls::rustls`

examples/http_mitm_proxy.rs
```diff
@@ -103,8 +114,14 @@ async fn main() -> Result<(), BoxError> {
         )
         .init();
 
+    #[cfg(feature = "boring")]
+    let mitm_tls_service_data =
+        new_mitm_tls_service_data_boring().context("generate self-signed mitm tls cert")?;
```
GlenDC, line 119: why name these functions differently? You might as well name them the same, no? Removes some extra code here as well.

examples/mtls_tunnel_and_service.rs
```diff
@@ -72,34 +81,51 @@ async fn main() {
...
+    let (tls_client_data, tls_server_data) = {
+        let (client_cert_chain, client_priv_key) = self_signed_client_auth().unwrap();
+        let client_cert = client_cert_chain[0].clone();
+        let http_versions = &[ApplicationProtocol::HTTP_2, ApplicationProtocol::HTTP_11];
+
+        let tls_client_data =
+            TlsConnectorDataBuilder::new_with_client_auth(client_cert_chain, client_priv_key)
+                .expect("connector with client auth")
+                .with_no_cert_verifier()
+                .with_http_versions(http_versions)
```
GlenDC, line 93: Why is thie called http versions, even though it's application protocols? Can be more than just http.

rama-cli/src/cmd/fp/storage/postgres.rs
```diff
@@ -10,7 +10,7 @@ use rama::{
     net::{address::Host, stream::Stream},
     tls::{
         boring::client::{TlsStream, tls_connect},
-        std::dep::boring::{hash::MessageDigest, nid::Nid, ssl::SslRef},
+        boring::dep::boring::{hash::MessageDigest, nid::Nid, ssl::SslRef},
```
GlenDC, line 13: in future we probably want to rename this path from `boring::dep::boring` to `boring::core` or so

rama-net/src/tls/keylog.rs
```diff
@@ -0,0 +1,142 @@
+//! Keylog facility used by any tls implementation
+//! supported by rama, and which can be used for your owns as well.
+//!
+//! Center to thsi module is the `KeyLogger` which is a wrapper around
```
GlenDC, line 4:
```suggestion
//! Center to this module is the `KeyLogger` which is a wrapper around
```

I guess typo from myself

rama-tls-rustls/src/client/connector_data.rs
```diff
@@ -0,0 +1,256 @@
...
+    /// on providing h2 connections.
+    pub fn new_http_2() -> Result<TlsConnectorData, OpaqueError> {
+        Ok(TlsConnectorDataBuilder::new()
+            .with_env_key_logger()?
+            .with_http_versions(&[ApplicationProtocol::HTTP_2])
+            .build())
+    }
+}
+
+/// [`ClientConfigBuilder`] can be used to construct [`rustls::ClientConfig`] for most common use cases in Rama.
```
GlenDC, line 71: Try to keep the first docstring line short as markdown being markdown a new line is only an actual newline if it is separated by an empty newline. So the problem with things like this is that your summary (first line) of `TlsConnectorDataBuilder` is going to be very long (in stuff like docs.rs)

rama-tls-rustls/src/client/connector_data.rs
```diff
@@ -0,0 +1,256 @@
...
+/// If this doesn't work for your use case, no problem [`TlsConnectorData`] can be created from a raw [`rustls::ClientConfig`]
+pub struct TlsConnectorDataBuilder {
+    pub client_config: rustls::ClientConfig,
```
GlenDC, line 74: It's not entirely wrong, but it feels weird for this to be public, and especially be the only thing public. It's also pretty anti-builder pattern wise.

GlenDC (changes requested):
Awesome work @soundofspace . Mostly looks ok, have a couple of changes to request but all minor. This is going in an excellent direction. I'm happy that the tls refactoring is in good hands of you!

examples/http_mitm_proxy.rs
soundofspace: at one point both of them were active at the same, but something I removed then, forgot to clean this one up

examples/mtls_tunnel_and_service.rs
soundofspace: agree, was focussed on that for the wrong reasons

rama-cli/src/cmd/fp/storage/postgres.rs
soundofspace: Yea indeed

rama-tls-rustls/src/client/connector_data.rs
soundofspace: yea was doubting this also, so thought lets include it and see what Glen says, but you have the same remark so making this private

examples/tls_rustls_termination.rs
```diff
@@ -56,9 +56,10 @@ use rama::{
         client::service::{Forwarder, TcpConnector},
         server::TcpListener,
     },
-    tls::std::server::{TlsAcceptorData, TlsAcceptorLayer},
+    tls::rustls::server::{TlsAcceptorDataBuilder, TlsAcceptorLayer},
 };
-use rama_net::tls::server::{SelfSignedData, ServerAuth, ServerConfig};
+use rama_net::tls::server::SelfSignedData;
+// use rama_net::tls::server::{SelfSignedData, ServerAuth, ServerConfig};
```
GlenDC, line 62: Why leave this in comment?

tests/integration/examples/example_tests/utils/mod.rs
```diff
@@ -197,6 +226,10 @@ impl ExampleRunner<()> {
                 .run()
                 .unwrap()
                 .command()
+                .env(
```
GlenDC, line 229: Can you clarify the rational behind this change? Is this not hiding a potential bigger issue under the hood? Your answer probably wont impact this PR but might require me to add a follow up GH issue.

GlenDC (changes requested):
2 questions left, after that let's ship it, I think.

examples/tls_rustls_termination.rs
soundofspace: mistake

tests/integration/examples/example_tests/utils/mod.rs
soundofspace: In some places you parse output of example started, to eg check if it is ready. All these logs are always of type info. This is not used everywhere, but for consistency sake I added this env var everywhere. Pretty sure in cicd this is set everywhere in env already to trace, so no problem there. One of the places where I encountered this "problem" was in the cli tests, where this happens:
```
if line.contains("ip service ready") {
```

GlenDC: Hmmm ok, we might want to look into a cleaner solution at some point, but it is what it is for now. I mean there are easy enough solutions that I can think of but hard to make it work without over complicating the examples. Oh well it is what it is for now.

GlenDC (approved):

tests/integration/examples/example_tests/utils/mod.rs
soundofspace: I honestly think this is is pretty decent solution considering the constraints of keeping the examples small/easy. But I do think we could/should add a lot more integrations tests where we wont have this problem. We can then always tell users once you are comfortable enough we rama code you can also check these integrations tests for more advances uses cases

GlenDC:
Follow up commit with some uber minor fixes: https://github.com/plabayo/rama/commit/136c5a7f52b8fd032738ccf865993fba361dd4b7

Issue #396, opened 2025-01-15
add web Router service to rama-http


GlenDC:
There's already a Matcher approach in https://docs.rs/rama/latest/rama/http/service/web/index.html, in two ways:

- WebService: this is using boxed handlers and comes with the benefits of nice regular Rust
- `match_service!` macro: this is same-same but using tuples and thus no need to Box. It is however a bit less easy to use for first time users

Both are very flexible in that they allow you to match on pretty much anything. However for a traditional web server it is a lot nicer to work with a trie-like structure to match purely on paths. You could still mix-and-match that with the two approaches above in case you want to specialise further (e.g. method) once you are in a path. This is possible as the two above still result in a `Service` (https://docs.rs/rama/latest/rama/trait.Service.html), which is also an `EndpointService` (https://docs.rs/rama/latest/rama/http/service/web/trait.IntoEndpointService.html)

The `Router` has to implement the `Service` trait (https://docs.rs/rama/latest/rama/trait.Service.html), which will be very similar to how https://docs.rs/rama/latest/rama/http/service/web/struct.WebService.html#impl-Service%3CState,+Request%3CBody%3E%3E-for-WebService%3CState%3E does it, except that you would do it using an approach focussed on paths. These paths should support extraction of segment data and wildcards out of the box and inject them similar to how a PatchMatcher (https://docs.rs/rama/latest/rama/http/matcher/struct.PathMatcher.html) does it.

API interface could look like:

```rust
use rama::http::service::web::Router

let app = Router::new()
    .get("/", root)
    .get("/users", list_users)
    .post("/users", create_user)
    .get("/users/{id}", show_user)
    .get("/assets/{*path}", serve_asset);
```
As part of implementing this `Router` you can add  `matchit` as a dependency: https://docs.rs/matchit/latest/matchit/ 

You could also go fancier and create also some kind of `RoutingService` which could wrap the `WebService` and makes it a lot easier to chain based on common matchers such as methods etc, for same path. But that's more of an extra.

Which would allow you to make it look more like:

```rust
use rama::http::service::web::{Router, routing::get};

let app = Router::new()
    .route("/", get(root))
    .route("/users", get(list_users).post(create_user))
    .route("/users", get(create_user))
    .route("/users/{id}", get(show_user))
    .route("/assets/{*path}", get(serve_asset));
```

Might be nicer and easier to implement, something to think about.

nagxsan:
[...]

PR #474, opened 2025-03-21
Implement web `Router` service to `rama-http`

ikorason:
Refer to #396

### DIFF
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -358,6 +358,11 @@ mod test {
                 params.insert("title".to_owned(), "oxford-dictionary".to_owned());
                 params
             }),
+            TestCase::some("/book/oxford-dictionary/author", "/book/{title}/author", {
+                let mut params = UriParams::default();
+                params.insert("title".to_owned(), "oxford-dictionary".to_owned());
+                params
+            }),
             TestCase::some(
                 "/book/oxford-dictionary/author/0",
                 "/book/:title/author/:index",
@@ -365,6 +370,16 @@ mod test {
                     let mut params = UriParams::default();
                     params.insert("title".to_owned(), "oxford-dictionary".to_owned());
                     params.insert("index".to_owned(), "0".to_owned());
+                    params
+                },
+            ),
+            TestCase::some(
+                "/book/oxford-dictionary/author/1",
+                "/book/{title}/author/{index}",
+                {
+                    let mut params = UriParams::default();
+                    params.insert("title".to_owned(), "oxford-dictionary".to_owned());
+                    params.insert("index".to_owned(), "1".to_owned());
                     params
                 },
             ),
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -13,7 +13,7 @@ pub struct UriParams {
 }
 
 impl UriParams {
-    pub fn insert(&mut self, name: String, value: String) {
+    fn insert(&mut self, name: String, value: String) {
         self.params
             .get_or_insert_with(HashMap::new)
             .insert(name, value);
@@ -60,6 +60,20 @@ impl UriParams {
             None => Err(de::PathDeserializationError::new(de::ErrorKind::NoParams)),
         }
         .map_err(UriParamsDeserializeError)
+    }
+
+    /// Extend the [`UriParams`] with the given iterator.
+    pub fn extend<I, K, V>(&mut self, iter: I) -> &mut Self
+    where
+        I: IntoIterator<Item = (K, V)>,
+        K: Into<String>,
+        V: Into<String>,
+    {
+        let params = self.params.get_or_insert_with(HashMap::new);
+        for (k, v) in iter {
+            params.insert(k.into(), v.into());
+        }
+        self
     }
 
     pub fn iter(&self) -> impl Iterator<Item = (&str, &str)> {
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -155,7 +155,7 @@ impl PathMatcher {
         let path = path.as_ref();
         let path = path.trim().trim_matches('/');
 
-        if !path.contains([':', '*']) {
+        if !path.contains([':', '*', '{', '}']) {
             return Self {
                 kind: PathMatcherKind::Literal(path.to_lowercase()),
             };
@@ -180,6 +180,9 @@ impl PathMatcher {
                     Some(PathFragment::Param(
                         s.trim_start_matches(':').to_lowercase(),
                     ))
+                } else if s.starts_with("{") && s.ends_with("}") {
+                    let param_name = s[1..s.len() - 1].to_lowercase();
+                    Some(PathFragment::Param(param_name))
                 } else if s == "*" && index == fragment_length - 1 {
                     Some(PathFragment::Glob)
                 } else {
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -63,6 +63,16 @@ impl UriParams {
     }
 }
 
+impl<'a> FromIterator<(&'a str, &'a str)> for UriParams {
+    fn from_iter<T: IntoIterator<Item = (&'a str, &'a str)>>(iter: T) -> Self {
+        let mut params = UriParams::default();
+        for (k, v) in iter {
+            params.insert(k.to_owned(), v.to_owned());
+        }
+        params
+    }
+}
+
 #[derive(Debug)]
 /// Error that can occur during the deserialization of the [`UriParams`].
 ///
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -13,7 +13,7 @@ pub struct UriParams {
 }
 
 impl UriParams {
-    fn insert(&mut self, name: String, value: String) {
+    pub fn insert(&mut self, name: String, value: String) {
         self.params
             .get_or_insert_with(HashMap::new)
             .insert(name, value);
@@ -60,6 +60,14 @@ impl UriParams {
             None => Err(de::PathDeserializationError::new(de::ErrorKind::NoParams)),
         }
         .map_err(UriParamsDeserializeError)
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&str, &str)> {
+        self.params
+            .as_ref()
+            .map(|params| params.iter().map(|(k, v)| (k.as_str(), v.as_str())))
+            .into_iter()
+            .flatten()
     }
 }
 
--- a/rama-http/src/matcher/path/mod.rs
+++ b/rama-http/src/matcher/path/mod.rs
@@ -180,7 +180,7 @@ impl PathMatcher {
                     Some(PathFragment::Param(
                         s.trim_start_matches(':').to_lowercase(),
                     ))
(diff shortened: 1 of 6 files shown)


ikorason:
@GlenDC not sure if this is the correct approach, but anyways I decided to give it a try, if I am in the right direction and this is the correct-ish approach I would like to work on this if thats possible?

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,151 @@
...
+
+use matchit::Router as MatchitRouter;
+use rama_core::{
+    Context,
+    service::{BoxService, Service},
+};
+use rama_http_types::Body;
+
+pub struct Router<State> {
+    routes: HashMap<Method, MatchitRouter<Arc<BoxService<State, Request, Response, Infallible>>>>,
```
GlenDC, line 13:
```suggestion
    routes: MatchitRouter<Vec<(HttpMatcher, BoxService<State, Request, Response, Infallible>)>>,
```

I would suggest to try this approach. Where you first match based on path and when found use the first service whose HttpMatcher matches the input.

Cool thing about the http matcher is that it allows not only a method, but also multiple methods, and other more complex matching cases.

GlenDC (changes requested):
Thx for taking this over. Can you please also comment on the issue, I'll assign it to you.

I had planned this for 0.2 which is planned to be release April, so please do put some priority on the ticket if you indeed want to do it.

rama-http/src/service/web/router.rs
ikorason: interesting, I will take a look at this!

ikorason:
Hey @GlenDC I think it's time I would like some feedback with this, so currently the basic routing is good I think like it supports chaining and params

Ex. Router::new().post("user", create_user_service).get("user/{id}", get_user_service)

But now I feel like my current implementation need some adjustment for sub routing, similar to nesting in WebService, but something like

Ex. Router::new().sub("/api", Router::new().get("/user/{id}", user_service)

Which resolves to "/api/user/{id}"

I don't know if I should change my data structure for 'routes' to make this work or currently I'm kind of mimicking how nested web service is implemented but so far no luck, What you think about this?

rama-http/src/matcher/path/mod.rs
```diff
@@ -170,6 +180,9 @@ impl PathMatcher {
                     Some(PathFragment::Param(
                         s.trim_start_matches(':').to_lowercase(),
                     ))
+                } else if s.starts_with("{") && s.ends_with("}") && s.len() > 2 {
```
GlenDC, line 183:
```suggestion
                } else if s.starts_with('{') && s.ends_with('}') && s.len() > 2 {
```

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+impl<State> Router<State>
+where
+    State: Send + Sync + 'static,
+{
+    pub fn new() -> Self {
+        Self {
+            routes: MatchitRouter::new(),
+            not_found: Arc::new(
```
GlenDC, line 42: I would just make this optional, and return a `404` if none.

but just for educational purposes thanks to <https://docs.rs/rama/latest/rama/http/service/web/trait.IntoEndpointService.html> you can just do `StatusCode::NOT_FOUND.into_endpoint_service()`... But again I would just Option it given you anyway box it already so easy enough to do.

GlenDC:
> Hey @GlenDC I think it's time I would like some feedback with this, so currently the basic routing is good I think like it supports chaining and params
> 
> Ex.
> 
> ```
> Router::new().sub("/api", Router::new().get("/user/{id}", user_service)
> ```
> 
> Which resolves to "/api/user/{id}"
> 
> I don't know if I should change my data structure for 'routes' to make this work or currently I'm kind of mimicking how nested web service is implemented but so far no luck, What you think about this?
> 
> Edit: Upon reading and researching I think delegating to another service is better approach only if the top level structure should be kept same, similar to how nest is handled in WebService, but for router something like
> 
> ```
> struct NestedRouterService<State> {
>     prefix: String,
>     nested: Arc<Router<State>>,
> }
> ```


Yeah I think `NestedRouterService` in combination with your `sub` api makes a lot of sense. Souns like a clean solution to me.

Again I wouldn't `Arc` all these things though. Because even though you arc all these services it's still expensive to clone all these data structures. So all one would really just need is an arc at the most outer layer.

Other than that your code looks clean. Only thing in your nested proposal is how to store the nested service. I would declare it as

```
struct NestedRouterService<State> {
    prefix: String,
    nested:  BoxService<State, Request, Response, Infallible>,
}
```

So that your nested can be _any_ service, including a Router which is also a service that matches this.

ikorason:
@GlenDC I think it's ready for review unless I miss something, only thing is I didn't implement RoutingService, but if times allow I can give it a try tho.

```rust
use rama::http::service::web::{Router, routing::get};

let app = Router::new()
    .route("/", get(root))
    .route("/users", get(list_users).post(create_user))
    .route("/users", get(create_user))
    .route("/users/{id}", get(show_user))
    .route("/assets/{*path}", get(serve_asset));
```

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+
+    /// add a GET route to the router.
+    /// the path can contain parameters, e.g. `/users/{id}`.
+    /// the path can also contain a glob matcher, e.g. `/assets/{*path}`.
+    pub fn get<I, T>(self, path: &str, service: I) -> Self
+    where
+        I: IntoEndpointService<State, T>,
+    {
+        // keep everything before the "{*" part and add a glob matcher
```
GlenDC, line 59: Can you explain why get has this special case and why it is required and why only here?

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,440 @@
...
+        let matcher_path = if let Some(start) = path.find("{*") {
+            format!("{}*", &path[..start])
+        } else {
+            path.to_owned()
+        };
+
+        let matcher = HttpMatcher::get(matcher_path);
```
GlenDC, line 66: I get why you did it, but it's a bit silly to also match on path in the matcher given matchit already will ensure we match correctly on path, including nested wise.

So please change these matchers to <https://docs.rs/rama/latest/rama/http/matcher/struct.HttpMatcher.html#method.method> instead.

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,440 @@
...
+
+    /// register a nested router under a prefix.
+    ///
+    /// The prefix is used to match the request path and strip it from the request URI.
+    pub fn sub<I, T>(self, prefix: &str, service: I) -> Self
+    where
+        I: IntoEndpointService<State, T>,
+    {
+        let path = format!("{}/{}", prefix.trim_end_matches(['/']), "{*nest}");
+        let matcher = HttpMatcher::path("*");
```
GlenDC, line 141: you can perhaps use `HttpMatcher::custom(true.into())`

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,440 @@
...
+        &self,
+        mut ctx: Context<State>,
+        req: Request,
+    ) -> Result<Self::Response, Self::Error> {
+        let mut ext = Extensions::new();
+
+        if let Ok(matched) = self.routes.at(req.uri().path()) {
+            for (matcher, service) in matched.value.iter() {
+                if matcher.matches(Some(&mut ext), &ctx, &req) {
+                    let uri_params = matched.params.iter().collect::<UriParams>();
```
GlenDC, line 243: You might want to invent a new private type `ParentUriParams(UriParams)` that you do not expose but that you inject into the context from within `NestedService`. This way you can merge it here into the uri_params, in case it exists.

GlenDC (changes requested):
I love it. Absolutely love it. You nailed it.
Left a couple of remarks, once those are resolved I think it's ready to be merged!!! Super exciting, thanks.

rama-http/src/service/web/router.rs
GlenDC: Also make sure to add a test case for this, so it is validated to work, because while it is probably a niche use case I bet someone will eventually have parameters in their prefix :P

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,442 @@
...
+        let nested_router_service = NestedRouterService {
+            prefix: prefix.to_owned(),
+            nested: service.into_endpoint_service().boxed(),
+        };
+
+        self.add_route(&path, matcher, nested_router_service)
+    }
+
+    /// add a route to the router with it's matcher and service.
+    fn add_route<I, T>(mut self, path: &str, matcher: HttpMatcher<State, Body>, service: I) -> Self
```
GlenDC, line 154: can you rename this to `match` and make it public. This allows one to use a custom matcher :)

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    async fn serve(
+        &self,
+        mut ctx: Context<State>,
+        mut req: Request,
+    ) -> Result<Self::Response, Self::Error> {
+        if let Some(param) = ctx.get::<UriParams>() {
+            let nested_params = param.clone();
+
+            if let Some(nest) = nested_params.get("nest") {
```
GlenDC, line 232: it should be an internal server error in case it is not available as you are the one defining this param, so how could it ever not be in there?

rama-http/src/service/web/router.rs
```diff
@@ -0,0 +1,512 @@
...
+    pub fn sub<I, T>(self, prefix: &str, service: I) -> Self
+    where
+        I: IntoEndpointService<State, T>,
+    {
+        let path = format!("{}/{}", prefix.trim_end_matches(['/']), "{*nest}");
+        let matcher = HttpMatcher::custom(true);
+
+        let nested_router_service = NestedRouterService {
+            prefix: prefix.to_owned(),
+            nested: service.into_endpoint_service().boxed(),
```
GlenDC, line 147: I just realized, sorry for not seeing that before, that because we allow _any_ service, the concept of `ParentUriParams` is not sound. I think we need to remove `ParentUriParams` again and instead you need to make sure that _anywhere_ (so also in existing code from before this PR) where you make+inject a `UriParams` you need to have the following kind of logic:

```
let params: UriParams = ctx.remove::<UriParams>() {
      Some(params) => params.extend(new_params),
      None => new_params.collect(),
}
ctx.insert(params);
```

hope that's clear. That also means you want to some kind of `extend` feature that works with an iterator, similar to how `fn extend` is implemented for stuff such as `Vec`.

GlenDC (changes requested):
Some last feedback, once again, my apologies. Hope you're ok with it, if not do let mek now and I'll take it over, as depending on you as a person you might want this to be done already, and I would get that. If not and you do want to apply this last feedback as well, big thanks, but otherwise just as grateful as well.

Ideally the last changes you'll make now require no changes to your tests, as those do look fine afaik as-is.

ikorason:
> Some last feedback, once again, my apologies. Hope you're ok with it, if not do let mek now and I'll take it over, as depending on you as a person you might want this to be done already, and I would get that. If not and you do want to apply this last feedback as well, big thanks, but otherwise just as grateful as well.
> 
> Ideally the last changes you'll make now require no changes to your tests, as those do look fine afaik as-is.

No need to apologies, I am still learning and you are providing with very good mentorship, I will take a stab at it and let you know!

ikorason:
@GlenDC plz check again, I removed the `ParentUriParams` instead use the approach you suggested, it works great, but let me know if there is anything else not looking good I can take a look again.

Issue #456, opened 2025-03-13
allow to overwrite the repeat collector for HeaderFromStrConfigLayer


GlenDC:
By default it will be `Vec`, but would be nice if one can overwrite it :)
Not need to not let it, especially given structs can have defaults.

YHM404:
Hi, could you please introduce more details about this? If possible, I would like to address this.

GlenDC:
Basically for https://docs.rs/rama/latest/rama/http/layer/header_from_str_config/struct.HeaderFromStrConfigLayer.html and (similar for its service), change the signature for that layer to:

```
pub struct HeaderFromStrConfigLayer<T, C = Vec<T>>
```

For this you'll need to also change the phantom data internally from `_marker: PhantomData<fn() -> T>,` to `_marker: PhantomData<fn() -> (T, C)>,

This way you can than enforce the `FromIterator` bound on `C` for the actual service and collect just like we do already, but have more freedom than being forced in `Vec`.

GlenDC:
Let me know if that's sufficiently clear or in case you need more guidance or have follow up questions.

GlenDC:
@YHM404 were the above instructions clear? Need any further guidance? Either way hope you're doing Ok and just want to know I'm here for you in case there's something I can help with. Take care!

YHM404:
> were the above instructions clear? Need any further guidance? Either way hope you're doing Ok and just want to know I'm here for you in case there's something I can help with. Take care!

Sorry I've been quite busy lately, I'll try to resolve this issue in the next couple of days.

YHM404:
Hi @GlenDC. Here is a little issue. For non-Vec types, there is no way to tell if it is empty. So if I replace `Vec` to `C`, [L169](https://github.com/plabayo/rama/blob/main/rama-http/src/layer/header_from_str_config.rs#L169) would lose a convenient way to check if C is empty.

GlenDC:
> Hi [@GlenDC](https://github.com/GlenDC). Here is a little issue. For non-Vec types, there is no way to tell if it is empty. So if I replace `Vec` to `C`, [L169](https://github.com/plabayo/rama/blob/main/rama-http/src/layer/header_from_str_config.rs#L169) would lose a convenient way to check if C is empty.

[...]

PR #475, opened 2025-03-22
Refactor HeaderFromStrConfigService and HeaderFromStrConfigLayer

YHM404:
Resolve https://github.com/plabayo/rama/issues/456

### DIFF
--- a/rama-http/src/layer/header_from_str_config.rs
+++ b/rama-http/src/layer/header_from_str_config.rs
@@ -44,6 +44,7 @@ use crate::{
 };
 use rama_core::{Context, Layer, Service, error::BoxError};
 use rama_utils::macros::define_inner_service_accessors;
+use std::iter::FromIterator;
 use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


rama-http/src/layer/header_from_str_config.rs
```diff
@@ -164,9 +170,12 @@ where
                             .map(|x| x.parse::<T>().map_err(Into::into))
                     })
                 })
-                .collect();
-            let values = result?;
-            if values.is_empty() {
+                .peekable();
+
+            let is_empty = parsed_values.peek().is_none();
+            let values = parsed_values.collect::<Result<C, _>>()?;
```
GlenDC, line 176: Only collect it right before you insert it into the context. Not needed nor desired to do this upfront.

GlenDC (changes requested):
Looks great. One comment to resolve and once done we can merge it. Thanks for a great job!

PR #482, opened 2025-03-27
replace cow static str with smol-str

### LINKED ISSUE #477: integrate smol-str in types such as Domain and other small str types
Currently these use cow<'static, str>, but smol-str is probably a better fit for these.

https://github.com/rust-analyzer/smol_str

airycanon:
close #477

### DIFF
--- a/rama-net/src/address/domain.rs
+++ b/rama-net/src/address/domain.rs
@@ -127,7 +127,7 @@ impl Domain {
         self.as_ref()
     }
 
-    /// Returns the domain name inner Cow value.
+    /// Returns the domain name inner value.
     ///
     /// Should not be exposed in the public rama API.
     pub(crate) fn into_inner(self) -> SmolStr {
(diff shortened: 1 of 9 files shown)


rama-net/src/address/domain.rs
```diff
@@ -129,14 +130,14 @@ impl Domain {
     /// Returns the domain name inner Cow value.
```
GlenDC, line 130:
```suggestion
    /// Returns the domain name inner value.
```

rama-net/src/address/domain.rs
```diff
@@ -182,7 +183,7 @@ impl<'a> TryFrom<&'a [u8]> for Domain {
 
     fn try_from(name: &'a [u8]) -> Result<Self, Self::Error> {
         if is_valid_name(name) {
-            Ok(Self(Cow::Owned(
+            Ok(Self(SmolStr::new(
                 String::from_utf8(name.to_vec()).context("convert domain bytes to utf-8 string")?,
```
GlenDC, line 187:
```suggestion
                std::str::from_utf8(name).context("convert domain bytes to utf-8 string")?,
```

to avoid a heap allocation just to pass it as reference to `SmolStr` :)

GlenDC (changes requested):
Thanks, great start, 2 minor comments, other than that looks good!

GlenDC (approved):
With requested changes applied looks good now.
Thank you for this swift contribution!

airycanon:
@GlenDC thanks for the changes.

PR #501, opened 2025-04-07
Either connector

soundofspace:
`EitherConn` which is a variant of `Either` but that works with Connectors. This should make it very easy to combine connectors as needed

To do this I reused as much as possible from `Either` so we don't re-invent this concept and can use what already works. `EitherConn` is needed because different connectors have different return types (different connections) which doesn't play nicely with `Either`. Connectors also have two layers of services: the connector and then the actual connection, to also make the connection work properly `EitherConnConnected` is also needed to implement all needed traits (mainly AsyncRead/Write)

### DIFF
--- a/rama-net/src/client/either_conn.rs
+++ b/rama-net/src/client/either_conn.rs
@@ -70,7 +70,7 @@ macro_rules! impl_service_either_conn {
 
 impl_either_conn!(impl_service_either_conn);
 
-/// `EitherConnConnected` is created when `EitherConn` has been connected and we now have an actuall
+/// `EitherConnConnected` is created when `EitherConn` has been connected and we now have an actual
 /// connection instead of a connector
 macro_rules! impl_either_conn_connected {
     ($macro:ident) => {


soundofspace:
Example of how to use it, will be used heavily in EasyHttpClient:

```rust
#[tokio::test]
async fn test_any_connector_stack() {
    let req = Request::get("https://127.0.0.1:63801")
        .body(Body::empty())
        .unwrap();
    let ctx = Context::default();

    let with_tls = true;

    let connector = TcpConnector::new();

    let tls_config = ClientConfig {
        server_verify_mode: Some(ServerVerifyMode::Disable),
        extensions: Some(vec![
            ClientHelloExtension::ApplicationLayerProtocolNegotiation(vec![
                ApplicationProtocol::HTTP_2,
                ApplicationProtocol::HTTP_11,
            ]),
        ]),
        ..Default::default()
    };

    let connector = match with_tls {
        true => EitherConn::A(
            TlsConnector::auto(connector).with_connector_data(tls_config.try_into().unwrap()),
        ),
        false => EitherConn::B(connector),
    };

    let connector = HttpConnector::new(connector);

    let EstablishedClientConnection { conn, req, ctx } =
    connector.serve(ctx.clone(), req).await.unwrap();

    let response = conn.serve(ctx, req).await.unwrap();
}
```

rama-net/src/client/either_conn.rs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                }
+            }
+        }
+    };
+}
+
+impl_either_conn!(impl_service_either_conn);
+
+/// `EitherConnConnected` is created when `EitherConn` has been connected and we now have an actuall
```
GlenDC, line 73:
```suggestion
/// `EitherConnConnected` is created when `EitherConn` has been connected and we now have an actual
```

GlenDC (approved):

PR #506, opened 2025-04-15
strip default ports from authority header in h2

soundofspace:
Browsers by default strip port from authority header if port is the default port for that specific protocol.
Some reverse proxies (eg nginx with some configs) do exact matches on authority header and return 404, if it doesn't match. Meaning authority header `example.com:443` will be seen as invalid and only `example.com` will be accepted.

### DIFF
--- a/rama-http-backend/src/client/svc.rs
+++ b/rama-http-backend/src/client/svc.rs
@@ -244,7 +244,7 @@ fn sanitize_client_req_header<S, B>(
                 // as some reverse proxies such as nginx respond 404 if authority is not an exact match
                 let authority = match request_ctx.is_authority_default_port() {
                     true => request_ctx.authority.host().to_string(),
-                    false => request_ctx.authority.host().to_string(),
+                    false => request_ctx.authority.to_string(),
                 };
 
                 uri_parts.authority = Some(
--- a/rama-http-backend/src/client/svc.rs
+++ b/rama-http-backend/src/client/svc.rs
@@ -10,7 +10,7 @@ use rama_http_types::{
     header::{CONNECTION, HOST, KEEP_ALIVE, PROXY_CONNECTION, TRANSFER_ENCODING, UPGRADE},
     headers::HeaderMapExt,
 };
-use rama_net::{address::ProxyAddress, http::RequestContext};
+use rama_net::{Protocol, address::ProxyAddress, http::RequestContext};
 use std::fmt;
 use tokio::sync::Mutex;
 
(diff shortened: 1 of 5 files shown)


rama-http-backend/src/client/svc.rs
```diff
@@ -239,10 +239,17 @@ fn sanitize_client_req_header<S, B>(
                 // NOTE: in a green future we might not need to stringify
                 // this entire thing first... maybe something someone at some
                 // point can take a look at this mess
+
+                // Default port is stripped in browsers. It's important that we also do this
+                // as some reverse proxies such as nginx respond 404 if authority is not an exact match
+                let authority = match (&request_ctx.protocol, request_ctx.authority.port()) {
```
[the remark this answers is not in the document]
soundofspace: Authority doesn't know which protocol is being used, but RequestCtx does, so moved `is_default_port` fn to there

rama-ua/src/emulate/service.rs
```diff
@@ -789,9 +789,12 @@ fn compute_sec_fetch_site_value(
                     let referer_protocol =
                         Protocol::maybe_from_uri_scheme_str_and_method(uri.scheme(), method);
 
-                    let default_port = uri
-                        .port_u16()
-                        .or_else(|| referer_protocol.as_ref().map(|p| p.default_port()));
+                    let default_port = uri.port_u16().or_else(|| {
+                        referer_protocol
+                            .as_ref()
+                            .map(|p| p.default_port())
```
GlenDC, line 795:
```suggestion
                            .and_then(|p| p.default_port())
```

^ so you do not need to flatten

rama-net/src/address/proxy.rs
```diff
@@ -62,7 +66,11 @@ impl TryFrom<&str> for ProxyAddress {
                 Host::try_from(slice).map(|h| {
                     (
                         h,
-                        protocol.as_ref().unwrap_or(&Protocol::HTTP).default_port(),
+                        protocol
+                            .as_ref()
+                            .map(|proto| proto.default_port())
```
GlenDC, line 71:
```suggestion
                            .and_then(|proto| proto.default_port())
```

rama-net/src/address/proxy.rs
```diff
@@ -41,7 +41,11 @@ impl TryFrom<&str> for ProxyAddress {
                         Host::try_from(&slice[i + 1..]).map(|h| {
                             (
                                 h,
-                                protocol.as_ref().unwrap_or(&Protocol::HTTP).default_port(),
+                                protocol
+                                    .as_ref()
+                                    .map(|proto| proto.default_port())
```
GlenDC, line 46: and_then

GlenDC (changes requested):
Better, some minor changes requested, and once those are resolved and we also have CI passing, I think it's good to go

GlenDC (approved):

PR #511, opened 2025-04-16
cli: add serve command

### LINKED ISSUE #508: add `serve` command to `rama-cli`
It's basically a static file web service.

It should be build very similar to "echo service", in the sense that:

- you'll want to implement the core logic of it all in https://github.com/plabayo/rama/tree/main/src/cli/service as `serve.rs` or w/e, feel free to look at `echo.rs` and copy paste from there to get you started
- then you'll add the cmd and actual command file to rama-cli: https://github.com/plabayo/rama/tree/main/rama-cli/src/cmd <- there as `serve/mod.rs`, which again will be pretty similar to echo, so feel free to take a look at that one

Similar to echo service the serve command should by default run it as a plain http service, but that you can also add it as a https service instead, either with a self-signed cert, or with the cert provided (again same as echo service).

There are 3 modes of operation to get us started, based on whether or not the (one and only) position argument is provided and if so what it contains:

- `rama serve`: doesn't have the position argument, it can just serve https://github.com/plabayo/rama/blob/main/docs/index.html, so it's still useful at least. This can be done by doing `Html(include_str!("../../path/bla/bla"))`
- `rama serve index.html`: if the positional argument is provided we always assume it is a path. However a path can still be pointing to either a file or a directory. If it's a file it should just be served as the only resource. TO serve this path you can use https://docs.rs/rama/latest/rama/http/service/fs/struct.ServeFile.html
- `rama serve content` if the positional argument is a path to a directory you'll serve the directory using https://docs.rs/rama/latest/rama/http/service/fs/struct.ServeDir.html

And that's about it. Should be pretty simple, yet another nice additional tool for `rama-cli` as well as anyone building a binary with rama that can make use of this `ServeService` out of the box (similar to `EchoService`).

Kornelijus:
Closes #508 

- [x] serve docs page by default
- [x] serve single file
- [x] serve directory

extra:
- [x] update docs for `rama-cli` usage
- [x] error when path is invalid
- [x] replace cli args port `-p` and interface `-i` with `--bind` using `rama::net::SocketAddress`
  - [x] update cli usage in fly.io deployments
  - [x] update cli usage in integration tests

### DIFF
--- a/rama-cli/src/cmd/serve/mod.rs
+++ b/rama-cli/src/cmd/serve/mod.rs
@@ -4,7 +4,7 @@ use clap::Args;
 use rama::{
     Service,
     cli::{ForwardKind, service::serve::ServeServiceBuilder},
-    error::BoxError,
+    error::{BoxError, ErrorContext, OpaqueError},
     http::{IntoResponse, Request, Response, matcher::HttpMatcher},
     layer::HijackLayer,
     net::{
@@ -145,11 +145,16 @@ pub async fn run(cfg: CliCommandServe) -> Result<(), BoxError> {
         .maybe_tls_server_config(maybe_tls_server_config)
         .http_layer(maybe_acme_service)
         .maybe_content_path(cfg.path)
-        .build(Executor::graceful(graceful.guard()))?;
+        .build(Executor::graceful(graceful.guard()))
+        .map_err(OpaqueError::from_boxed)
+        .context("build serve service")?;
 
     tracing::info!("starting serve service on: {}", cfg.bind);
-
-    let tcp_listener = TcpListener::build().bind(cfg.bind).await?;
+    let tcp_listener = TcpListener::build()
+        .bind(cfg.bind)
+        .await
+        .map_err(OpaqueError::from_boxed)
+        .context("bind serve service")?;
 
     graceful.spawn_task_fn(async move |guard| {
         tracing::info!("serve service ready");
--- a/rama-cli/src/cmd/serve/mod.rs
+++ b/rama-cli/src/cmd/serve/mod.rs
@@ -1,4 +1,4 @@
-//! Serve service that serves a file, directory or placeholder page
+//! Serve service that serves a file, directory or placeholder page.
 
 use clap::Args;
 use rama::{
--- a/rama-cli/src/cmd/serve/mod.rs
+++ b/rama-cli/src/cmd/serve/mod.rs
@@ -1,0 +1,179 @@
+//! Serve service that serves a file, directory or placeholder page
+
+use clap::Args;
+use rama::{
+    Service,
+    cli::{ForwardKind, service::serve::ServeServiceBuilder},
+    error::BoxError,
+    http::{IntoResponse, Request, Response, matcher::HttpMatcher},
+    layer::HijackLayer,
+    net::tls::{
+        ApplicationProtocol, DataEncoding,
+        server::{SelfSignedData, ServerAuth, ServerAuthData, ServerConfig},
+    },
+    rt::Executor,
+    tcp::server::TcpListener,
+};
+
+use base64::Engine;
+use base64::engine::general_purpose::STANDARD as ENGINE;
+
+use std::{convert::Infallible, time::Duration};
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::{EnvFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt};
+
+#[derive(Debug, Args)]
+/// rama serve service (serves a file, directory or placeholder page)
+pub struct CliCommandServe {
+    #[arg(short = 'p', long, default_value_t = 8080)]
+    /// the port to listen on
+    port: u16,
+
+    #[arg(short = 'i', long, default_value = "127.0.0.1")]
+    /// the interface to listen on
+    interface: String,
+
+    #[arg(short = 'c', long, default_value_t = 0)]
+    /// the number of concurrent connections to allow
+    ///
+    /// (0 = no limit)
+    concurrent: usize,
+
+    #[arg(short = 't', long, default_value_t = 8)]
+    /// the timeout in seconds for each connection
+    ///
+    /// (0 = no timeout)
+    timeout: u64,
+
+    #[arg(long, short = 'f')]
+    /// enable support for one of the following "forward" headers or protocols
+    ///
+    /// Supported headers:
+    ///
+    /// Forwarded ("for="), X-Forwarded-For
+    ///
+    /// X-Client-IP Client-IP, X-Real-IP
+    ///
+    /// CF-Connecting-IP, True-Client-IP
+    ///
+    /// Or using HaProxy protocol.
+    forward: Option<ForwardKind>,
+
+    #[arg(long, short = 's')]
+    /// run serve service in secure mode (enable TLS)
+    secure: bool,
+}
+
+/// run the rama serve service
+pub async fn run(cfg: CliCommandServe) -> Result<(), BoxError> {
+    tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        )
+        .init();
+
+    let maybe_tls_server_config = cfg.secure.then(|| {
+        let tls_key_pem_raw = match std::env::var("RAMA_TLS_KEY") {
+            Ok(raw) => raw,
+            Err(_) => {
+                return ServerConfig {
+                    application_layer_protocol_negotiation: Some(vec![
+                        ApplicationProtocol::HTTP_2,
+                        ApplicationProtocol::HTTP_11,
+                    ]),
+                    ..ServerConfig::new(ServerAuth::SelfSigned(SelfSignedData::default()))
+                };
+            }
+        };
+        let tls_key_pem_raw = std::str::from_utf8(
+            &ENGINE
+                .decode(tls_key_pem_raw)
+                .expect("base64 decode RAMA_TLS_KEY")[..],
+        )
+        .expect("base64-decoded RAMA_TLS_KEY valid utf-8")
+        .try_into()
+        .expect("tls_key_pem_raw => NonEmptyStr (RAMA_TLS_KEY)");
+        let tls_crt_pem_raw = std::env::var("RAMA_TLS_CRT").expect("RAMA_TLS_CRT");
+        let tls_crt_pem_raw = std::str::from_utf8(
+            &ENGINE
+                .decode(tls_crt_pem_raw)
+                .expect("base64 decode RAMA_TLS_CRT")[..],
+        )
+        .expect("base64-decoded RAMA_TLS_CRT valid utf-8")
+        .try_into()
+        .expect("tls_crt_pem_raw => NonEmptyStr (RAMA_TLS_CRT)");
+        ServerConfig {
+            application_layer_protocol_negotiation: Some(vec![
+                ApplicationProtocol::HTTP_2,
+                ApplicationProtocol::HTTP_11,
+            ]),
+            ..ServerConfig::new(ServerAuth::Single(ServerAuthData {
+                private_key: DataEncoding::Pem(tls_key_pem_raw),
+                cert_chain: DataEncoding::Pem(tls_crt_pem_raw),
+                ocsp: None,
+            }))
+        }
+    });
+
+    let maybe_acme_service = std::env::var("RAMA_ACME_DATA")
+        .map(|data| {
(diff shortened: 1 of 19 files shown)


src/cli/service/serve.rs
```diff
@@ -0,0 +1,350 @@
...
+            http_forwarded_layer,
+        )
+            .into_layer(self.http_service_builder.layer(ServeService {}))
+    }
+}
+
+#[derive(Debug, Clone)]
+#[non_exhaustive]
+/// The inner serve-service used by the [`ServeServiceBuilder`].
+pub struct ServeService {}
```
GlenDC, line 321:
```suggestion
pub struct ServeService;
```

FYI :)

src/cli/service/serve.rs
```diff
@@ -0,0 +1,350 @@
...
+    async fn serve(
+        &self,
+        mut ctx: Context<()>,
+        req: Request,
+    ) -> Result<Self::Response, Self::Error> {
+        Ok(Html(include_str!("../../../docs/index.html")).into_response())
+    }
+}
+
+pub mod mode {
```
GlenDC, line 336: FYI There is probably no need for any of this mode. You. can just have 3 different http services internally:

- Html<'static str>
- ServeFile
- ServieDir

Which you wrap with `Either3`

Should be sufficient for what you need without the need to introduce a lot of boilerplate code around it.

GlenDC (changes requested):
I know it's still early work, but thought it might help to leave some feedback already.

src/cli/service/serve.rs
Kornelijus: Gotcha, looked at this `mode` pattern from `ip`, but I now see it's used because the user explicitly specifies the mode, while here we infer the type of path.

I assume there's no need to wrap `ServeFile` or `ServeDir` at this point, so updated to represent the service like
```rs
type ServeService = Either3<ServeStaticHtml, ServeFile, ServeDir>;
```

src/cli/service/serve.rs
```diff
@@ -0,0 +1,370 @@
...
+        };
+
+        Ok(tcp_service_builder.into_layer(http_transport_service))
+    }
+
+    fn build_serve(&self) -> ServeService {
+        match self.content_path {
+            None => Either3::A(ServeStaticHtml(Html(include_str!(
+                "../../../docs/index.html"
+            )))),
```
GlenDC, line 306: You can just do `Html(include_str!("../../../docs/index.html")).into_endpoint_service()` ;)
(makes use of <https://ramaproxy.org/docs/rama/http/service/web/trait.IntoEndpointService.html>)

src/cli/service/serve.rs
```diff
@@ -0,0 +1,370 @@
...
+
+            http_version: self.http_version,
+
+            http_service_builder: (self.http_service_builder, layer),
+
+            content_path: self.content_path,
+        }
+    }
+
+    pub fn content_path(mut self, path: PathBuf) -> Self {
```
GlenDC, line 230: You can use `impl Into<PathBuf>` to also allow other stuff, makes it easier, you can do this also for `set_content_path`. However `maybe_content_path` better is left like it is as otherwise `None` will not know what type to use.

GlenDC (commented):
Looks already a lot better, thx for that! already have some more feedback as well while you're at it. If you do not like this do tell me and I'll wait until you ask for review (which I usually do anyway).

src/cli/service/serve.rs
Kornelijus: Nice, replaced the wrapper with this, just had some trouble with the types where bounds for `H: Layer<...>` are specified, so it ended up as `BoxService`.

rama-cli/src/cmd/serve/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+pub struct CliCommandServe {
+    /// The path to the file or directory to serve
+    ///
+    /// If not provided, a placeholder page will be served.
+    #[arg()]
+    path: Option<PathBuf>,
+
+    /// the address to bind to
+    #[arg(long, default_value = "127.0.0.1:8080")]
+    bind: SocketAddress,
```
Kornelijus, line 39: Not sure if we want a name with a matching shorthand, since there are already args using `-a` and `-b`.
For the long name I think any of bind/addr/listen would be ok.

src/cli/service/serve.rs
```diff
@@ -0,0 +1,370 @@
...
+        self.content_path = Some(path);
+        self
+    }
+}
+
+impl<H> ServeServiceBuilder<H>
+where
+    H: Layer<ServeService, Service: Service<(), Request, Response = Response, Error = BoxError>>,
+{
+    #[allow(unused_mut)]
```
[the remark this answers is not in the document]
Kornelijus: Left over from `echo`, oops.

Kornelijus:
> Looks already a lot better, thx for that! already have some more feedback as well while you're at it. If you do not like this do tell me and I'll wait until you ask for review (which I usually do anyway).

Between the language features used, and the framework itself, this is all new to me, so the quick feedback is very appreciated! Usually, I try to iterate with many small commits and glance over the final diff before requesting a review, but that becomes tricky when you simply 'don't know what you don't know' yet :smile:

rama-cli/src/cmd/serve/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    let tcp_service = ServeServiceBuilder::new()
+        .concurrent(cfg.concurrent)
+        .timeout(Duration::from_secs(cfg.timeout))
+        .maybe_forward(cfg.forward)
+        .maybe_tls_server_config(maybe_tls_server_config)
+        .http_layer(maybe_acme_service)
+        .maybe_content_path(cfg.path)
+        .build(Executor::graceful(graceful.guard()))
+        .expect("build serve service");
```
Kornelijus, line 149: Left all `expect`s for consistency, here output for invalid path is a panic.

rama-cli/src/cmd/serve/mod.rs
GlenDC: Only long is just fine :) people can always propose a short version if it bothers them. For now let's go with your choice of just the long one, I like it.

rama-cli/src/cmd/serve/mod.rs
GlenDC:
```suggestion
        .context("build serve service")?;
```

best to return it. You say consistency so I assume you see it somewhere else as well? if so feel free to fix it there. Because a panic renders pretty ugly on a cli. Not critical, but a bit ugly. While we have a nice error catcher in main.rs

src/cli/service/serve.rs
```diff
@@ -0,0 +1,362 @@
...
+            }
+            Some(ForwardKind::CFConnectingIp) => {
+                Some(Either7::F(GetForwardedHeadersLayer::<CFConnectingIp>::new()))
+            }
+            Some(ForwardKind::TrueClientIp) => {
+                Some(Either7::G(GetForwardedHeadersLayer::<TrueClientIp>::new()))
+            }
+        };
+
+        let serve_service = match self.content_path {
```
GlenDC, line 332:
```suggestion
        let serve_service = match &self.content_path {
```

this way you do not need to use `ref path` everywhere down :)

src/cli/service/serve.rs
```diff
@@ -0,0 +1,362 @@
...
+            None => Either3::A(
+                Html(include_str!("../../../docs/index.html"))
+                    .into_endpoint_service()
+                    .boxed(),
```
GlenDC, line 336: can you make https://github.com/plabayo/rama/blob/796392e175df26a36d5e5b9ae657fa29d82c1b91/rama-http/src/service/web/endpoint/mod.rs#L45 public please and add a comment to it. The inner property can remain private, so just change it to `pub struct StaticService<R>(R);` with a comment added to the top.  And then add it to the list her: https://github.com/plabayo/rama/blob/796392e175df26a36d5e5b9ae657fa29d82c1b91/rama-http/src/service/web/mod.rs#L9

Once that is done you can replace `type ServeStaticHtml = BoxService<(), Request, Response, Infallible>;` with `type ServeStaticHtml = StaticService<Html<&'satic str>>;` which prevents a silly box. Not that i am against boxing, but besides the cost it also loses meaning, so nice if you can name things.

GlenDC (changes requested):
Only a couple of minor changes, other than that I am ok with merging this in, looks good to me :) Thx a lot for this first contribution of yours, amazing work!

rama-cli/src/cmd/serve/mod.rs
Kornelijus: Sure, in that case I'll also update to handle `AddrInUse` when binding to a port, since that's a common error.
That leaves the setup for TLS and ACME, but I guess that's not expected to fail often.

GlenDC: Unless there is a good reason not to we should handle it or propagate it.

Bit out of scope perhaps for your pr but if you have the energy for it, correcting it would be appreciated

src/cli/service/serve.rs
Kornelijus: Sorry if I'm missing something, I also added `StaticService::new` with the same bounds for `R` to make this work.

I think this doesn't work as is because the `into_endpoint_service` return type is opaque `impl Service...` so the compiler will not give guarantees about it being `StaticService` because the implementation may change.

GlenDC: Sorry, you're correct indeed. In that case you can keep the StaticService private, sorry for the trouble. Boxing it is indeed

GlenDC: Correct. You can add a public 'fn new' to that StaticService within an 'impl<R: IntoResponse + Clone + Send + Sync + 'static> StaticService {}' block.

This way you can construct it directly instead of calling 'into_endpoint_service'

src/cli/service/serve.rs
```diff
@@ -0,0 +1,359 @@
...
+            trace::TraceLayer, ua::UserAgentClassifierLayer,
+        },
+        server::HttpServer,
+    },
+    layer::{ConsumeErrLayer, LimitLayer, TimeoutLayer, limit::policy::ConcurrentPolicy},
+    net::stream::layer::http::BodyLimitLayer,
+    proxy::haproxy::server::HaProxyLayer,
+    rt::Executor,
+};
+use rama_http::{
```
GlenDC, line 22: move these in the relevant section under `use crate::{` please

GlenDC (changes requested):
Looks good to me. One final remark, once that is resolved I can merge this PR in, assuming CI checks all pass. Thx for all your harad work!

rama-cli/src/cmd/serve/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+    let graceful = rama::graceful::Shutdown::default();
+
+    let tcp_service = ServeServiceBuilder::new()
+        .concurrent(cfg.concurrent)
+        .timeout(Duration::from_secs(cfg.timeout))
+        .maybe_forward(cfg.forward)
+        .maybe_tls_server_config(maybe_tls_server_config)
+        .http_layer(maybe_acme_service)
+        .maybe_content_path(cfg.path)
+        .build(Executor::graceful(graceful.guard()))?;
```
Kornelijus, line 148: Quick fix for the panics, looks like `ErrorContext` may need changes to work with `BoxError` here.

rama-cli/src/cmd/serve/mod.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    tracing::info!("starting serve service on: {}", cfg.bind);
+
+    let tcp_listener = TcpListener::build().bind(cfg.bind).await?;
```
Kornelijus, line 152: Also fixes issue where tokio worker panics because bind failed, but program does not exit.

rama-cli/src/cmd/serve/mod.rs
GlenDC: No way to fix the. Just do map_err(OpaqueError::from_boxed) or w/e works for your case

rama-cli/src/cmd/serve/mod.rs
GlenDC: That doesn't apply here, no?

Where it does apply the easiest solution is to do the fallible operation (eg bind) prior to moving it to a task or in worst case introduce an error channel or something like it.

Kornelijus: Sorry, I mean that I also moved it out of `spawn_task_fn`, and that fixed the program hanging.  
I'll update it for the other commands in a sec.

GlenDC (approved):

GlenDC:
Looks excellent. Thank you so much for your first contribution. You went above and beyond. Great job!

PR #512, opened 2025-04-18
feat: server include dir

### LINKED ISSUE #507: add support for `include_dir` for to ServeDir and related
We have in rama-http already functionality to serve files and directories using `ServeDir`: https://docs.rs/rama/latest/rama/http/service/fs/struct.ServeDir.html

This works great and is usually what you want. However sometimes you serve only a couple of files as part of some cli or other kind of binary, and what you really want for ease of use is to serve directly from memory, while still preserving these serve_dir semantics.

Here is where https://crates.io/crates/include_dir/0.7.4 can come in handy.

Now serveDir does a lot of heavy lifting, so it would be silly to create an entire different layer/service just for this additional use-case. So instead I think it could make sense to be able to work a bit on the internals so that we have an enum or something like it, to differentiate internally between serving from disk or serving from memory. 

It will be a small refactor but probably worth it in the long run.

airycanon:
Closes #507


GlenDC:
Do you already want some feedback @airycanon or will you ping me once you want me to have a look?
or are you stuck with something, have questions or need guidance? Either way take care. just checking in :) Have a good one.

airycanon:
@GlenDC  Thanks for you attention, could you take a look at my current approach and see if I'm on the right track.

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -2,6 +2,7 @@ use crate::dep::http_body::{self, Body as HttpBody};
 use crate::layer::set_status::SetStatus;
 use crate::{Body, HeaderValue, Method, Request, Response, StatusCode, header};
 use bytes::Bytes;
+use include_dir::Dir;
```
GlenDC, line 5: Please add `pub use include_dir` to `dep` module in <https://github.com/plabayo/rama/blob/39039d33f9e3c93d1db406a233041b20a3557cc0/rama-http/src/lib.rs#L37> and use it also here from dep folder as that lib might be most stable so will be easier for us to upgrade that dep on major versions without breaking users (assuming they use our `dep` version).

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -515,7 +538,7 @@ impl<F> ServeDir<F> {
 
         let path_to_file = match self
             .variant
-            .build_and_validate_path(&self.base, req.uri().path())
+            .build_and_validate_path(self.base.clone(), req.uri().path())
```
GlenDC, line 541: why all these `self.base.clone()` are needed all of a sudden? We should still be able to take by reference, no? What am I missing?

GlenDC (commented):
Looks pretty alright to me. Added 2 comments but besides that seems that you do get the idea and approach. Also please add at least 1 unit test making use of this embedded option where you test some different cases with it (similar to the tests already there for the regular dir option).

airycanon:
@GlenDC Do we need to do encoding processing for embedded files?

GlenDC:
> @GlenDC Do we need to do encoding processing for embedded files?

Yes for sure. It would make serving such content even cheaper as you do not need to do any processing in that case, it's basically just a straight byte copy at that point.

The encoding support in `serve_dir` is just about the ability to support pre-compressed content. E.g. that you serve `file.txt.gz` in case client requests `file.txt` and has `gzip` in their accept-encoding header.

In the end `include_dir` is pretty much like dir, except that we do not have metadata about file (so the new `HtmlFileTree` mode will provide `--` for last modified data but can still support size obviously), but we do have information that's in the path, such as file extension which is where the encoding support comes in.

GlenDC:
It's by the way not the most easy feature to add btw as it does combine a couple of different topics and I have recently expanded this code a bit. So do let me know if you need help at any point or some guidance / mentoring, I'm here for you if you do.

GlenDC:
@airycanon , would you like me to take over this PR, or do you have a plan for it?

* First, I’d sync it with the latest main.
* Then, continue the work.

Given your recent inactivity and the remaining work after all the great progress you’ve already made, I understand if you'd prefer someone else to finish it. If you do want to continue, just let me know, and I’ll be happy to assist. However, if we go that route, I’d just ask that we avoid letting this drag on any longer.

airycanon:
@GlenDC Sorry for the long delay. I recently started a new job and the development team has been very busy with many tasks. Now that things have settled into a routine at work, I finally have time to address this. Could please take a look at?

GlenDC:
> @GlenDC Sorry for the long delay. I recently started a new job and the development team has been very busy with many tasks. Now that things have settled into a routine at work, I finally have time to address this. Could please take a look at?

No problem. Happy to have you here, even after all this time. Welcome back!

Cargo.toml
```diff
@@ -481,7 +482,7 @@ required-features = ["http-full", "rustls", "acme"]
 
 [[example]]
 name = "autobahn_client"
-required = ["rama-ws", "http-full"]
+required-features = ["ws", "http-full"]
```
GlenDC:
```suggestion
required-features = ["http-full"]
```

Good catch ,it's from a mistake of another PR. I also fixed it in a WIP PR from me already, but good that you also fixed it. Love it. Means you are paying attention to the details. Very good and very much appreciated.

(ws is already a feature included in http-full)

rama-http-core/Cargo.toml
```diff
@@ -48,19 +48,19 @@ want = { workspace = true }
...
 
-# Fuzzing
-quickcheck = { workspace = true }
-rand = { workspace = true }
-
 # HPACK fixtures
 hex = { workspace = true }
+
+# Fuzzing
+quickcheck = { workspace = true }
```
GlenDC: Why did these move, was that automatically or did you do it?
Can you perhaps revert changes to this file as seems you didn't add anything to it?

rama-http-core/tests/h2-support/Cargo.toml
```diff
@@ -7,12 +7,12 @@ edition = "2024"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-rama-core = { workspace = true }
-rama-http-core = { workspace = true, features = ["unstable"] }
-rama-http-types = { workspace = true }
 
 futures = { workspace = true }
 http = { workspace = true }
+rama-core = { workspace = true }
```
GlenDC: please revert this file

rama-http/src/lib.rs
```diff
@@ -44,6 +44,8 @@ pub mod dep {
 
     pub use rama_core as core;
 
+    pub use include_dir;
```
GlenDC: Do not do this. If we do want to re-export it it should be under a `dep` module similar to other locations where we do that.

However, why do we need to expose this? are there types or functions that folks need from that library to make this work? If they are limited enough it might make more sense to include them in the location of where we use it instead. But would need to know the list of symbols you need folks to use from the crate directly.

rama-http/src/service/fs/mod.rs
```diff
@@ -52,16 +52,6 @@ where
             reader: ReaderStream::with_capacity(read, capacity),
         }
     }
-
-    fn with_capacity_limited(
```
GlenDC: Why is this removed?

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -58,7 +66,21 @@ impl ServeDir<DefaultServeDirFallback> {
         base.push(path.as_ref());
 
         Self {
-            base,
+            base: DirSource::Filesystem(base),
+            buf_chunk_size: DEFAULT_CAPACITY,
```
GlenDC: might make sense to make a private `new_with_base` fn so we only need to define these defaults for all other ones once instead of twice

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -369,29 +391,48 @@ impl<F> ServeDir<F> {
...
-            req,
-            negotiated_encodings,
-            range_header,
-            buf_chunk_size,
-        )
-        .await;
+        let open_file_result = match &self.base {
+            DirSource::Filesystem(_) => {
+                let variant = self.variant.clone();
+                let range_header = req
```
GlenDC: Why not keep this up so we do not need to define this two times

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -487,37 +528,53 @@ enum ServeVariant {
...
-                        Component::CurDir => {}
-                        Component::Prefix(_) | Component::RootDir | Component::ParentDir => {
-                            return None;
-                        }
+                        Some(path)
+                    }
+                    DirSource::Embedded(_) => {
+                        // For embedded directories, return the decoded path
+                        // The actual file lookup will be handled in open_file_embedded
+                        Some(path_decoded.to_path_buf())
```
GlenDC: Do we not need similar protection for both sources? Even if it's embedded we are still dealing with same kind of file paths, no?

rama-http/src/service/fs/serve_dir/open_file.rs
```diff
@@ -41,14 +45,122 @@ pub(super) struct FileOpened {
 pub(super) enum FileRequestExtent {
     Full(File, Metadata),
     Head(Metadata),
+    Embedded(Vec<u8>),
+    EmbeddedHead(u64), // Content length for HEAD requests on embedded files
+}
+
+impl FileRequestExtent {
+    pub(super) fn reader(self) -> Option<Box<dyn AsyncRead + Send + Sync + Unpin>> {
```
GlenDC: You should be able to return `Option<impl AsyncRead + Send + Sync + Unpin>` by working here internally with `Either`. Do let me know if I'm missing something.

rama-http/src/service/fs/serve_dir/open_file.rs
```diff
@@ -41,14 +45,122 @@ pub(super) struct FileOpened {
...
+            Self::Full(file, _) => Some(Box::new(file)),
+            Self::Embedded(content) => Some(Box::new(Cursor::new(content))),
+        }
+    }
+
+    pub(super) fn range_reader_with_offset(
+        self,
+        offset: u64,
+        range_size: u64,
+    ) -> Option<Box<dyn AsyncRead + Send + Sync + Unpin>> {
```
GlenDC: You should be able to return Option<impl AsyncRead + Send + Sync + Unpin> by working here internally with Either. Do let me know if I'm missing something.

rama-http/src/service/fs/serve_dir/open_file.rs
```diff
@@ -41,14 +45,122 @@ pub(super) struct FileOpened {
...
+        .map(HeaderValue::from_static)
+        .unwrap_or_else(|| HeaderValue::from_str(mime::APPLICATION_OCTET_STREAM.as_ref()).unwrap());
+
+    let maybe_encoding = preferred_encoding(&mut path_to_file, negotiated_encodings);
+
+    let Some(file) = base.get_file(&path_to_file) else {
+        return Ok(OpenFileOutput::FileNotFound);
+    };
+
+    let last_modified = file
```
GlenDC: there is a lot of overlap between file embedded and regular file. Not only here but in general. Do try to keep a check on the duplication. There are are probably some more clever abstractions we can introduce to limit that kind of code duplication

rama-http/src/service/fs/serve_dir/tests.rs
```diff
@@ -33,6 +33,26 @@ async fn basic() {
     assert_eq!(body, contents);
 }
 
+#[tokio::test]
+async fn basic_embedded() {
```
GlenDC, line 23: same same here, try to avoid all the duplication. From a user perspective it's pretty much the same so would be good if at the very least every test is defined just once but ran for both embedded and non-embedded

GlenDC (changes requested):
the PR looks good. Now it's time to clean up and polish it all up and then we can merge it probably in. Do let me know if you need specific guidance or even if you need me to partner up on this.

rama-http/src/service/fs/serve_dir/tests.rs
airycanon: I’ve reused the shared parts of theese test cases. Also, because include_dir has performance issues when pointed at the repository root, I changed some cases to use the README.md and Cargo.toml in the rama-http directory.

airycanon:
@GlenDC PTAL

rama-http/src/service/fs/serve_dir/tests.rs
```diff
@@ -16,8 +16,17 @@ use std::io::Read;
...
 #[tokio::test]
 async fn basic() {
-    let svc = ServeDir::new("..");
+    let svc = ServeDir::new("../rama-http");
+    test_basic(svc).await;
+}
 
+#[tokio::test]
+async fn basic_embedded() {
+    let svc = ServeDir::new_embedded(include_dir::include_dir!("rama-http"));
```
GlenDC: how come relative paths do not work here similar to `include_str!`? Bit confusing if this would work different from include_str and include_bytes people are used to?

rama-http/src/service/fs/serve_dir/tests.rs
```diff
@@ -29,14 +38,24 @@ async fn basic() {
 
     let body = body_into_text(res.into_body()).await;
 
-    let contents = std::fs::read_to_string("../README.md").unwrap();
+    let contents = std::fs::read_to_string("../rama-http/README.md").unwrap();
     assert_eq!(body, contents);
 }
 
 #[tokio::test]
 async fn basic_with_index() {
```
GlenDC, line 45: If we could use relative path here as well you would be able to have a macro_rules to reduce even more code

Usage would be something like

```rust
impl_serve_dir_test!(
    basic_with_index,
    "../test-files",
    async |svc: ServeDir| {
        let req = Request::new(Body::empty());
        let res = svc.serve(Context::default(), req).await.unwrap();
    
        assert_eq!(res.status(), StatusCode::OK);
        assert_eq!(res.headers()[header::CONTENT_TYPE], "text/html");
        let body = body_into_text(res.into_body()).await;
        #[cfg(target_os = "windows")]
        assert_eq!(body, "<b>HTML!</b>\r\n");
        #[cfg(not(target_os = "windows"))]
        assert_eq!(body, "<b>HTML!</b>\n");
    },
);
```

which would generate the two tests `test_basic_with_fs` and `test_basic_with_include`,
as AFAIK that serve dir creation code is always the same for all tests?

rama-http/src/service/fs/serve_dir/future.rs
```diff
@@ -151,11 +146,11 @@ fn build_response(output: FileOpened) -> Response {
                         )))
                         .unwrap()
                 } else {
-                    let body = if let Some(file) = maybe_file {
-                        let range_size = range.end() - range.start() + 1;
+                    let range_size = range.end() - range.start() + 1;
```
GlenDC, line 154: why did you switch around these two statements? Seems like it's a diff for no good reason?

rama-http/src/service/fs/serve_dir/future.rs
```diff
@@ -119,10 +117,7 @@ where
 }
 
 fn build_response(output: FileOpened) -> Response {
-    let (maybe_file, size) = match output.extent {
-        FileRequestExtent::Full(file, meta) => (Some(file), meta.len()),
-        FileRequestExtent::Head(meta) => (None, meta.len()),
-    };
+    let size = output.extent.get_size();
```
GlenDC:
```suggestion
    let size = output.extent.file_size();
```

it's not idiomatic in Rust to use `get_` for getters. Plus the real issue is that "size" is not clear to what it is referring. As such two problems solved at once by renaming the method to `file_size`.

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -57,6 +65,15 @@ impl ServeDir<DefaultServeDirFallback> {
         let mut base = PathBuf::from(".");
         base.push(path.as_ref());
 
+        Self::new_with_base(DirSource::Filesystem(base))
+    }
+
+    #[must_use]
+    pub fn new_embedded(path: Dir<'static>) -> Self {
```
GlenDC, line 72: please add proper documentation for this method

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -57,6 +65,15 @@ impl ServeDir<DefaultServeDirFallback> {
         let mut base = PathBuf::from(".");
         base.push(path.as_ref());
 
+        Self::new_with_base(DirSource::Filesystem(base))
```
GlenDC, line 67: please also add extra docts to the regular `new` to make clear that if they want it from an embedded dir that they can do so using `new_embedded`

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -374,24 +391,34 @@ impl<F> ServeDir<F> {
...
+                    path_to_file,
+                    req,
+                    negotiated_encodings,
+                    range_header.as_deref(),
+                    buf_chunk_size,
+                )
+                .await
+            }
+            DirSource::Embedded(path) => open_file_embedded(
+                path,
```
GlenDC: can you explain why the path comes in this case from `embedded` while otherwise it comes from `variant`?

rama-http/src/service/fs/serve_dir/mod.rs
```diff
@@ -487,38 +514,53 @@ enum ServeVariant {
...
-                        Component::CurDir => {}
-                        Component::Prefix(_) | Component::RootDir | Component::ParentDir => {
-                            return None;
-                        }
+                match source {
+                    DirSource::Filesystem(base_path) => {
+                        Self::validate_path_components(base_path.clone(), path_decoded)
+                    }
+                    DirSource::Embedded(_) => {
+                        Self::validate_path_components(PathBuf::new(), path_decoded)
```
GlenDC: there is no base path here needed?

rama-http/src/service/fs/serve_dir/open_file.rs
```diff
@@ -41,26 +65,95 @@ pub(super) struct FileOpened {
...
+    mut path_to_file: PathBuf,
+    req: &Request,
+    negotiated_encodings: &[QualityValue<Encoding>],
+    range_header: Option<&str>,
+    buf_chunk_size: usize,
+) -> io::Result<OpenFileOutput> {
+    // Check if this is a directory request (empty path or directory path)
+    // If so, try to serve index.html
+    if path_to_file.as_os_str().is_empty() || path_to_file.to_str() == Some(".") {
+        path_to_file = PathBuf::from("index.html");
```
GlenDC: shouldn't this behavior depend on the serve mode? Only when append index mode is used or w/e. As right now I think you are always doing it? Check how the regular FS based approach does it.

rama-http/src/service/fs/serve_dir/open_file.rs
```diff
@@ -41,26 +65,95 @@ pub(super) struct FileOpened {
 pub(super) enum FileRequestExtent {
     Full(File, Metadata),
     Head(Metadata),
+    Embedded(Vec<u8>, u64), // Content, original file size
+    EmbeddedHead(u64),      // Content length for HEAD requests on embedded files
+}
+
+impl FileRequestExtent {
+    pub(super) fn reader(self) -> Option<impl AsyncRead + Send + Sync + Unpin> {
```
GlenDC:
```suggestion
    pub(super) fn into_reader(self) -> Option<impl AsyncRead + Send + Sync + Unpin> {
```

to make it clear we are turning `self` into something else. Because when I read `reader` alone I assume it is not consuming, while here it is.

rama-http/src/service/fs/serve_dir/tests.rs
airycanon: yeah, here is the explanation from the [official documentation](https://docs.rs/include_dir/latest/include_dir/#environment-variables)： 

> When invoking the [include_dir!()](https://docs.rs/include_dir/latest/include_dir/macro.include_dir.html) macro you should try to avoid using relative paths because rustc makes no guarantees about the current directory when it is running a procedural macro.

GlenDC: I see... That crate is no longer maintained though... I think I'm gonna fork this and add it to rama-macros... especially as this is now possible to make it relative similar to `include_str` and `include_bytes`. https://doc.rust-lang.org/stable/proc_macro/struct.Span.html the methods we need are stable in rama's MSRV so all is well... Plus that crate hasn't been touched in a year which does smell for something with so much unfinished business... Not something I want to expose our users to.

Gonna add it this week into rama and you can switch to that instead of using that crate..

In case you have any feature requests from stuff you were missing in the original code that you wanted to have as well let me know, perhaps I can quickly add it in while I do some cleaning up.

rama-http/src/service/fs/serve_dir/mod.rs
airycanon: When using include_dir, we must call Dir::get_file() with relative paths to access file contents, which means we can't append the requested path to a base_path like we do with filesystem operations.

GlenDC: Can you give a specific example why that is not possible? I just fail to grasp why there's a difference here? The path given by the url will always be relative, no?

rama-http/src/service/fs/serve_dir/tests.rs
GlenDC: @airycanon https://github.com/plabayo/rama/pull/665 is merged. You should be able to merge `main` into your branch and remove the `include_dir` dependency. Instead you can just use the macro from `rama_utils::include_dir`.

Also added an example for you found at `/examples/http_service_include_dir.rs` that you can adapt to use `ServeDir::new_embedded` instead of the web router. Should work similar to the `http_service_fs` example.

Relative paths are now supported in the `include_dir` macro so you should be able to make those examples more easily shared :)

airycanon: @GlenDC Thanks for the great work. However, I noticed that basic and basic_embedded still behave differently when I try to use them with the same directory path. 

Here's what I'm experiencing:
```rust
#[tokio::test]
async fn basic() {
    let svc = ServeDir::new("../rama-http");
    test_basic(svc).await;
}

#[tokio::test]
async fn basic_embedded() {
    let svc = ServeDir::new_embedded(include_dir!("../rama-http"));
    test_basic(svc).await;
}

async fn test_basic(svc: ServeDir) {
    println!("Current working directory: {:?}", std::env::current_dir().unwrap());
}
```

The behavioral difference stems from when and where the path resolution occurs:

### basic test (Runtime Path Resolution)

  let svc = ServeDir::new("../rama-http");

  - Base directory: The current working directory of the test process (rama/rama-http)
  - Path resolution: ../rama-http resolves to rama/rama-http
  - Result: Successfully finds files because it's reading from the actual filesystem

 ### basic_embedded test (Compile-time Path Resolution)

  let svc = ServeDir::new_embedded(include_dir!("../rama-http"));
  - Base directory: The location of the source file (rama-http/src/service/fs/serve_dir/tests.rs)
  - Path resolution: ../rama-http resolves to rama-http/src/service/fs/rama-http (which doesn't exist)
  - Result: Compilation fails because the macro cannot find the directory

rama-http/src/service/fs/serve_dir/mod.rs
airycanon: ### Filesystem Operations
```rust
  // Base path
  let base_path = PathBuf::from("../rama-http");

  // User requests "/README.md"
  let requested_path = "/README.md";

  // File content，read
  let content = std::fs::read_to_string("../rama-http/README.md")?;
```
  ### Embedded Directory Operations
```rust
  // Directory structure embedded at compile time
  const base_path: Dir = include_dir!("../rama-http");

  // User requests "/README.md"
  let requested_path = "/README.md";

  // File content
  let content = base_path.get_file("README.md")?; 
```
  Key Differences:
 - Filesystem: Base path is dynamic (runtime), so we need to concatenate base_path + requested_path
  - include_dir: Base path is static (compile-time), so the embedded directory already represents that base, and we only need the requested path relative to it

rama-http/src/service/fs/serve_dir/tests.rs
GlenDC: Can you please push all your latest commits (including this test) to your branch. I'll check it out as that would surprise me very much.

airycanon: @GlenDC Done, I've pushed all the changes.

GlenDC: Ok I get it now, it's because of the FileSystem path... Yes of course that depends on the FS. But at least the `include_dir!` now is predictable. Sadly for your tests that makes no difference but for normal include_dir use case that does make it more sane. Suffice to say I understand you now.

```
#[tokio::test]
async fn basic_embedded() {
    let svc = ServeDir::new_embedded(include_dir!("../rama-http"));
    test_basic(svc).await;
}
```

That path will have to be changed to `"../../../../"` or probably more sane would be to do `"$CARGO_MANIFEST_DIR"` as that will be the directory of `rama-http` root. For cases where you need to access `test-files` you can do for embedded: `"$CARGO_MANIFEST_DIR/../test-files"`.

You can even afaik use that same variable for the `fs` case using `env!("CARGO_MANIFEST_DIR")` and `concat!(env!("CARGO_MANIFEST_DIR"), "/../test-files")` but I suppose you might as well continue to use the existing paths in those tests for the FS cases.

As such I'm ok with your test file as it is now, just fix the embedded paths and leave the FS paths alone. That is what my suggestion is of course, you use your own judgement. TLDR: I do understand what you are saying now.

PR #520, opened 2025-04-21
Dynamic async rustls config depending on ch

soundofspace:
This adds support to dynamically set a rustls config based on the received client hello. This is really powerfully and will allow us to support a lot of different rustls use cases without having to explicitly support it in rama

One of those use cases is dynamically loading in certificates in an async context, also added example for it.


rama-tls-rustls/src/server/acceptor_data.rs
```diff
@@ -1,47 +1,113 @@
...
+
+impl<D: DynamicConfigProvider> From<Arc<D>> for TlsAcceptorData {
+    fn from(value: Arc<D>) -> Self {
+        Self {
+            server_config: ServerConfig::Async(value),
+        }
+    }
+}
+
+pub trait DynamicConfigProvider: Send + Sync + 'static {
```
GlenDC, line 73: Implement it also already for an `FnOnce + Clone` please. This will also simplify your example as you can just make an `async fn` and pass that as dynamic provider

GlenDC (changes requested):
Looks excellent, only 1 minor request, other than that good to go. Thx!

rama-tls-rustls/src/server/acceptor_data.rs
GlenDC: Nvm won't work because of `ClientHello` being a reference in Rustls

GlenDC (approved):

Issue #509, opened 2025-04-15
add `curl` module to rama-http-types


GlenDC:
It's gonna be a relatively small module with 2 methods:

1. `pub fn request_headers_to_curl_command(request: Impl IntoCurlHeadersCommand) -> Result<String, OpaqueError>`
2. `pub async fn request_to_curl_command(request impl IntoCurlCommand) -> Result<(String, Request), OpaqueError>`

(1) only deals with the headers, so not the body, while (2) deals with the actual body.

That means that (2) needs full access to the body. It doesn't however have to consume or destroy it, so it can return it. But it does make it potentially a more heavy lift process.

- `IntoCurlHeadersCommand` would be a sealed trait and can be implemented for now already for:
   - request `Parts`
   - request `&Parts`
   - `Request<B>` for any `B`
   - `&Request<B>` for any `B`

- IntoCurlCommand` would be a sealed trait and can be implemented for now already for:
   - `Request<Body>` (own it)
 
Where the latter `Body` needs to respect:

```
Body: rama::http::dep::http_body::Body<Data = bytes::Bytes, Error: Into<BoxError>>
        + Send
        + Sync
```

And that's about it.

The result should be a valid (one-line) curl command.

Example:

```
request_headers_to_curl_command(&Request::builder().uri("http://example.com").header("accept", "application/json").body(()).unwrap()).unwrap()
```

Would result in:

```
curl --http1.1 -H 'accept: application/json' http://example.com
```

Make sure to add sufficient test cases in that module.

As icing on the cake you can add a new property `curl: <cmd>` to the `EchoService` such that services such as https://echo.ramaproxy.org/ will have that info added to the `http` object:

```{
     "http": {
           "curl": "curl --http1.1 -H 'accept: application/json' http://example.com",
           ...
     }
     ...
}
```

Remarks:

- make sure to have also a look at https://docs.rs/rama/latest/rama/http/io/fn.write_http_request.html as it will show you how you can get the original header order and casing of your request, which is importnat
- pseudo header info you do not need as there is no way to pass this to curl
- tls can be ignored for now as well

GlenDC:
Hi @Taenerys thank you for your interest in the project. I'll assign the issue to you.

In case you have questions, need help, guidance or mentoring, do let me know. I'll gladly assist you. As such feel free to open a PR as soon as you want or need it, even if very early WIP. If it helps you it's alright.

GlenDC:
[...]

PR #521, opened 2025-04-22
[DRAFT/WIP PR] feat: adding curl module to rama-http-types

Taenerys:
Starting my work for #509 

**Progress:**
- Read rama documentation
- Get familiar with the codebase by running cargo build and tests
- Read up on `write_http_request` as recommended
- Review rama-http-types module
- Review sealed traits and how Rama uses it across the codebase
- Write the skeleton and add TODOs accordingly

@GlenDC here is my draft PR! i'll continue to make commits to this PR as I work on this issue, would super appreciate your thoughts/feedbacks whenever you see anything that can be improved, thank you so much Glen!


rama-http-types/src/curl/traits.rs
```diff
@@ -0,0 +1,55 @@
+use crate::{Body, Request, Parts}
+use crate::dep::http_body;
+use bytes::Bytes;
+use rama_core::error::{BoxError, OpaqueError};
+
+mod sealed {
+    pub trait Sealed {}
```
GlenDC, line 7: Good start, but better make 2 sealed traits, and call them the same as the public trait but with the `Priv` suffix. Also move the actual functions into the private trait, this way you also avoid that others can make use of the methods, giving us the flexibility to still change this.

So it would become something like:

```
trait IntoCurlHeadersCommandPriv {
    fn into_curl_headers_command(&self) -> Result<String, OpaqueError>;
}

pub trait IntoCurlHeadersCommand: IntoCurlHeadersCommandPriv {}
```

rama-http-types/src/curl/traits.rs
```diff
@@ -0,0 +1,55 @@
...
+
+    impl Sealed for crate::Parts {}
+
+    impl<T: Sealed> Sealed for &T {}
+}
+
+/// Sealed trait for converting HTTP request headers into a curl command
+
+pub trait IntoCurlHeadersCommand: sealed::Sealed {
+    fn into_curl_headers_command(&self) -> Result<String, OpaqueError>;
```
GlenDC, line 22: `into_curl_command` is indeed gonna be fallible as you'll need to consume the body, but pretty certain that `into_curl_headers_command` can be infallible, so can just return a `String`. Accessing header data cannot fail I think. Not in a way that is anyway recoverable.

rama-http-types/src/curl/traits.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+
+/// Sealed trait for converting a full HTTP request into a curl command.
+pub trait IntoCurlCommand: sealed::Sealed {
+    fn into_curl_command(self) -> Result<(String, Request), OpaqueError>;
```
GlenDC, line 28: it would be a bit silly I think (correct me if mistaken) to have to format the cmd for each impl. As such I think it's more useful if this just provides access to the relevant parts.

E.g. `IntoCurlCommandPriv` can be done with just a single method: `async fn try_into_request(self) -> Result<Request, OpaqueError>`

rama-http-types/src/curl/traits.rs
GlenDC: Also `IntoCurlHeadersCommandPriv` probably should just provide some methods to give access to the relevant parts. `Http1HeaderMap` (or however it's called), http version, method, uri, ... Where all except the h1HeaderMap are returned as reference

rama-http-types/src/curl/headers.rs
```diff
@@ -0,0 +1,13 @@
+use rama_core::error::OpaqueError;
+
+pub fn request_headers_to_curl_command<T: IntoCurlHeadersCommand>(request: T) -> Result<String, OpaqueError> {
```
GlenDC, line 3: FYI in case you have an `AcceptEncoding` header also do not forget to add the `--compressed` flag at the end

rama-http-types/src/curl/body.rs
```diff
@@ -0,0 +1,16 @@
+use crate::{Body, Request};
+use rama_core::error::OpaqueError;
+
+pub fn request_to_curl_command<T: IntoCurlCommand>(request: T) -> Result<(String, Request), OpaqueError> {
+    todo!()
```
GlenDC, line 5: Well organised. Thanks for that. FYI you probably can make use of `request_headers_to_curl_command` here (or some common internal methods at the least) to do most of the heavy lifting for you, except for the body (if there's any of course).

GlenDC (commented):
Thank you. You are off to a great start. Do let me know if you are ever stuck, require a bit of mentoring / help / etc... I'm here for you if you need me. Left a couple of remarks already but other than that I am looking forward to your further progress. Thank you for your effort and work so far and I hope you have a nice week as it is progressing! Take care.

GlenDC:
Closing this for inactivity.
Feel free to pick up another issue in future or any other contribution / feedback you see fit @Taenerys .
Take care and thank you for all. I hope you are ok and wish you the best.

PR #524, opened 2025-04-23
Remove rustls dependency from rama net

soundofspace:
Orphan rule prevents us from implementing `From` for rama-net tls types to rustls types in rama-tls-rustls. To workaround this we add our own `From` trait `RamaFrom` (and also into, try_from). This allows us to standardize this process and can also be extended to rama-tls-boring, but also works if people want to implement external tls crates

Part of: #391


rama-utils/src/macros/enums.rs
```diff
@@ -322,3 +322,47 @@ macro_rules! __enum_builder {
...
+///
+/// Orphan rule happens because we neither own the type or the trait where we want
+/// to define their actual implementation.
+///
+/// By adding these traits to crates where we have this problem
+/// we can workaround that. This will become the standard approach
+/// where the normal from/into doesn't work. Main use case right now for
+/// this trait is to support defining conversions from rama tls types (rama-net)
+/// to external types (eg rustls) in tls crates (eg rama-tls-rustls). This macro
+/// should be called from the root module, and ideally these traits are also
```
GlenDC, line 338:
```suggestion
/// should be called from the root module.
```

rama-utils/src/macros/enums.rs
```diff
@@ -322,3 +322,47 @@ macro_rules! __enum_builder {
...
+/// added to the prelude of that crate
```
GlenDC, line 339:
```suggestion
```

GlenDC (approved):
looks good, another great step towards a cleaner tls integration, thx!

GlenDC:
Please merge once threads are resolved @soundofspace and you are satisfied as well.

PR #535, opened 2025-04-29
Issue 528/make default pool usage easier for external use

soundofspace:


rama-net/src/client/pool.rs
```diff
@@ -3,124 +3,266 @@ use crate::stream::Socket;
...
+                returner: self.returner(),
+            }));
+        }
+
+        let pool_slot = match self.total_slots.clone().try_acquire_owned() {
+            Ok(permit) => PoolSlot(permit),
+            Err(_) => {
+                // By poping from back when we have no new Poolslot available we implement LRU drop policy
+                self.storage
+                    .lock()
```
GlenDC, line 205: it's not new code, but why actually the choice to drop the original lock prior to getting a slot?

GlenDC (approved):
Looks good, just couple of minor comments, once that is resolved I think you can merge this one in. Nice work!

rama-net/src/client/pool.rs
soundofspace: This is kind of new code, two functions were merged into one, hence why locking twice. Should have seen this, thanks for noticing

rama-net/src/client/pool.rs
```diff
@@ -3,124 +3,266 @@ use crate::stream::Socket;
...
-    /// can still decide if it will immedialty create storage of the given capacity,
-    /// or do it in a custom way (eg grow storage dynamically as needed)
-    fn new(capacity: NonZeroU16) -> Self;
+#[derive(Debug, Clone)]
+/// Connection pool that doesn't store connections and has no limits.
+///
+/// Basically this pool operates like there would be no connection pooling.
+/// Can be used in places where were we work with a [`PooledConnector`], but
+/// don't want connection pooling to happen.
+pub struct NoPool;
```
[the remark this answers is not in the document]
soundofspace: Good catch forgot it here

rama-net/src/client/pool.rs
```diff
@@ -585,13 +523,110 @@ impl<Storage, R> PooledConnectorLayer<Storage, R> {
...
+#[cfg(feature = "http")]
+pub mod http {
+    use std::time::Duration;
+
+    use super::{FiFoReuseLruDropPool, PooledConnector, ReqToConnID};
+    use crate::{Protocol, address::Authority, client::pool::OpaqueError, http::RequestContext};
+    use rama_core::Context;
+    use rama_http_types::Request;
+
+    #[derive(Debug, Default)]
```
GlenDC, line 549: shouldn't these be `Clone` as well otherwise `PoolConnector` etc won't be clone-able? And anyway doesn't hurt for these to be clone either way

rama-net/src/client/pool.rs
```diff
@@ -585,13 +523,110 @@ impl<Storage, R> PooledConnectorLayer<Storage, R> {
...
+    #[non_exhaustive]
+    /// [`BasicHttpConnIdentifier`] can be used together with a [`super::Pool`] to create a basic http connection pool
+    pub struct BasicHttpConnIdentifier;
+
+    pub type BasicHttpConId = (Protocol, Authority);
```
GlenDC, line 554: you can leave it as-is btw, but I keep thinking if it wouldn't be better to just always hash these so you don't need to compare the full types over and over again as well as clone them? There are plenty of fast hashing algorithms out there, and given we are dealing with a small pool the chance of collision even for collision-risk hashes are probably neglecatable. 

Again feel free to leave as-is, but just throwing it out there as every time I see this code that is the first thing that flashes through me, yet I never have written it to you.

rama-net/src/client/pool.rs
```diff
@@ -3,124 +3,279 @@ use crate::stream::Socket;
...
+    }
+}
+
+impl<C, ID> FiFoReuseLruDropPool<C, ID> {
+    pub fn new(max_active: usize, max_total: usize) -> Self {
+        let storage = Arc::new(Mutex::new(VecDeque::with_capacity(max_total)));
+        Self {
+            storage,
+            returner: OnceLock::new(),
+            total_slots: Arc::new(Semaphore::const_new(max_total)),
```
GlenDC, line 151: if max_active == 0 or max_total == 0, do we want to allow that? Should this perhaps return a result to check for that? or should we turn them into defaults if 0? Or should it for example mean that there is no max total and we can can just keep permitting and have as many active as we want?

rama-net/src/client/pool.rs
soundofspace: I have thought about this aswel but have ignored it for now until we have benchmarks for things like this. Same for `returner: OnceLock::new(),`, I'm honestly not sure if using a Oncelock here is cheaper and/or faster then just boxing and creating that Fn each time.

rama-net/src/client/pool.rs
soundofspace: For this specific pool I added errors for these cases. Transforming 0 to a default feels wrong to me, if you want the default calling Pool::default is much more explicit and clear. And having no limit here seems like a memory leak waiting to happen. This pool assumes active <= total, as it sees active as part of total, but could be interesting idea to have a pool where we have limit on stored and no limit active so we can reuse connections, but if we dont have any we can always create new ones

GlenDC: yeah but at same time feels weird to allow it than to be just 0, 0, which makes it also useless. Might better to return a result than for now and provide an alternative constructor in future or breaking change if we do ever want to try such ideas.

rama-net/src/client/pool.rs
GlenDC: yeah that oncelock does seem odd there... But we can always revisit that later

rama-net/src/client/pool.rs
soundofspace: Yea changed it to result commit wasn't pushed yet when i wrote this back

PR #539, opened 2025-05-02
Issue 534/ Implement SubdomainTrieMatcher Logic

### LINKED ISSUE #534: add a SubdomainTrieMatcher
- you can add <https://github.com/laysakura/trie-rs> as a dependency for it (we'll use it, but it will be an internal detail not exposed to the outside, this way we can still change it if we want to)
- add it under `rama-http/src/matcher/subdomain_trie.rs`
- support its usage also in `HttpMatcherKind` and `HttpMatcher`

it can be created from a list of domains that will be the source of tries (and thus it also can implement FromIterator for anything that can be turned into a valid Domain), internally you'll just insert them as`&[u8]` in the trie.

Prior to inserting it in the tree you do have to reverse it, example algo (also use tests for it):

```
 fn reverse_domain(domain: &str) -> String {
    let from = domain.strip_prefix('.').unwrap_or(domain);
    let mut domain = from.split('.').rev().collect::<Vec<&str>>().join(".");
    domain.push('.');
    domain
}
```

^ do add unit test(s) for it if you use it, or a better version of it (open to suggestions).
Either way the algorithm is only used to create the Trie matcher not each time it is trying to match.

LED-0102:
Closes #534

Added a trie-based domain matcher for efficient subdomain matching.
Ensured `sub.domain.com` returns `true` if `domain.com` is present but not vice-versa.

Please review and suggest any improvements. If there are any concerns about the logic or potential edge cases, kindly point them out! I'd appreciate any feedback as I'm still learning.

### DIFF
--- a/rama-http/Cargo.toml
+++ b/rama-http/Cargo.toml
@@ -58,7 +58,7 @@ tokio = { workspace = true, features = ["macros", "fs", "io-std"] }
 tokio-util = { workspace = true, features = ["io"] }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
-trie-rs = "0.4.2"
+radix_trie = { workspace = true }
 
 [dev-dependencies]
 brotli = { workspace = true }
--- a/rama-http/Cargo.toml
+++ b/rama-http/Cargo.toml
@@ -43,6 +43,7 @@ mime_guess = { workspace = true }
 nanoid = { workspace = true }
 percent-encoding = { workspace = true }
 pin-project-lite = { workspace = true }
+radix_trie = { workspace = true }
 rama-core = { version = "0.2.0-alpha.14", path = "../rama-core" }
 rama-http-types = { version = "0.2.0-alpha.14", path = "../rama-http-types" }
 rama-macros = { version = "0.2.0-alpha.14", path = "../rama-macros" }
@@ -58,7 +59,6 @@ tokio = { workspace = true, features = ["macros", "fs", "io-std"] }
 tokio-util = { workspace = true, features = ["io"] }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
-radix_trie = { workspace = true }
 
 [dev-dependencies]
 brotli = { workspace = true }
--- a/rama-http/Cargo.toml
+++ b/rama-http/Cargo.toml
@@ -58,6 +58,7 @@ tokio = { workspace = true, features = ["macros", "fs", "io-std"] }
 tokio-util = { workspace = true, features = ["io"] }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
+trie-rs = "0.4.2"
 
 [dev-dependencies]
 brotli = { workspace = true }
--- a/rama-http/src/matcher/mod.rs
+++ b/rama-http/src/matcher/mod.rs
@@ -615,6 +615,42 @@ impl<State, Body> HttpMatcher<State, Body> {
         self.or(Self::custom(matcher))
     }
 
+    /// Create a [`SubdomainTrieMatcher`] matcher that matches if the request domain is a subdomain of the provided domains.
+    ///
+    /// See [`SubdomainTrieMatcher`] for more information.
+    pub fn any_subdomain<I, S>(domains: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        HttpMatcher {
+            kind: HttpMatcherKind::SubdomainTrie(SubdomainTrieMatcher::new(domains)),
+            negate: false,
+        }
+    }
+
+    /// Add a [`SubdomainTrieMatcher`] matcher that matches if the request domain is a subdomain of the provided domains on top of the existing set of [`HttpMatcher`] matchers.
+    ///
+    /// See [`SubdomainTrieMatcher`] for more information.
+    pub fn and_any_subdomain<I, S>(self, domains: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        self.and(Self::any_subdomain(domains))
+    }
+
+    /// Create a [`SubdomainTrieMatcher`] matcher that matches if the request domain is a subdomain of the provided domains as an alternative to the existing set of [`HttpMatcher`] matchers.
+    ///
+    /// See [`SubdomainTrieMatcher`] for more information.
+    pub fn or_any_subdomain<I, S>(self, domains: I) -> Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<str>,
+    {
+        self.or(Self::any_subdomain(domains))
+    }
+
     /// Create a [`PathMatcher`] matcher to match for a POST request.
     pub fn post(path: impl AsRef<str>) -> Self {
         Self::method_post().and_path(path)
--- a/rama-http/src/matcher/mod.rs
+++ b/rama-http/src/matcher/mod.rs
@@ -34,6 +34,10 @@ mod header;
 #[doc(inline)]
 pub use header::HeaderMatcher;
 
+mod subdomain_trie;
+#[doc(inline)]
+pub use subdomain_trie::SubdomainTrieMatcher;
+
 /// A matcher that is used to match an http [`Request`]
 pub struct HttpMatcher<State, Body> {
     kind: HttpMatcherKind<State, Body>,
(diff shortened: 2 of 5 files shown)


LED-0102:
Hello @GlenDC,
When you have a moment, could you please take a look? Thank you, sir!

rama-http/Cargo.toml
```diff
@@ -58,6 +58,7 @@ tokio = { workspace = true, features = ["macros", "fs", "io-std"] }
 tokio-util = { workspace = true, features = ["io"] }
 tracing = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
+trie-rs = "0.4.2"
```
GlenDC, line 61: add it first as a dependency under `./Cargo.toml` please (workspace dependency, afterwards you can add it here as `trie-rs = { workspace = true }`.

The workspace dependency would be defined as:

```
trie-rs = { version = "0.4", features = ["serde"] }
```

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
+use trie_rs::TrieBuilder;
+use crate::Request;
+use rama_core::{Context, context::Extensions, matcher::Matcher};
+use rama_net::http::RequestContext;
+use rama_net::address::{Host};
+
+#[derive(Debug, Clone)]
+pub struct SubdomainTrieMatcher {
+    trie: trie_rs::Trie<u8>,
```
GlenDC, line 9:
```suggestion
    trie: Trie<u8>,
```

can just also add it to your `use trie_rs::` statement above

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+            let reversed = reverse_domain(domain.as_ref());
+            builder.push(reversed.as_bytes());
+        }
+        let trie = builder.build();
+        Self { trie }
+    }
+
+    pub fn is_match(&self, domain: &str) -> bool {
+        let reversed = reverse_domain(domain);
+        let domain_parts = reversed.split('.');
```
GlenDC, line 29: my appologies, this is on me, can you please switch to radix_tree crate and use https://docs.rs/radix_trie/0.2.1/radix_trie/struct.Trie.html for your internal trie structure. That one has a nice method called `get_ancestor()` so we can reduce this method to:

```
let reversed = reverse_domain(domain);
self.trie.get_ancestor(&reversed).is_some()
```

seems `trie-rs` is not yet optimised for this use case as it for now collects

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+        let mut builder = TrieBuilder::new();
+        for domain in domains {
+            let reversed = reverse_domain(domain.as_ref());
+            builder.push(reversed.as_bytes());
+        }
+        let trie = builder.build();
+        Self { trie }
+    }
+
+    pub fn is_match(&self, domain: &str) -> bool {
```
GlenDC, line 27:
```suggestion
    pub fn is_match(&self, domain: impl AsRef<str>) -> bool {
```

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+fn reverse_domain(domain: &str) -> String {
+    let from = domain.strip_prefix('.').unwrap_or(domain);
+    let mut domain = from.split('.').rev().collect::<Vec<&str>>().join(".");
+    domain.push('.');
+    domain
+}
+
+impl<State, Body> Matcher<State, Request<Body>> for SubdomainTrieMatcher {
```
GlenDC, line 54: Can be simplified to this, great suggestion btw to add the req-ctx to the ext if we had to compute it! This simplification avoids you having to clone host.

```rust
impl<State, Body> Matcher<State, Request<Body>> for SubdomainTrieMatcher {
    fn matches(
        &self,
        ext: Option<&mut Extensions>,
        ctx: &Context<State>,
        req: &Request<Body>,
    ) -> bool {
        let match_authority = |ctx: &RequestContext| {
            match ctx.authority.host() {
                Host::Name(domain) => {
                    let is_match = self.is_match(domain.as_str());
                    tracing::trace!(
                        %domain,
                        %is_match,
                        "SubdomainTrieMatcher: matching", domain, is_match,
                    );
                    is_match
                }
                Host::Address(address) => {
                    tracing::trace!(
                        %address,
                        "SubdomainTrieMatcher: ignoring numeric address",
                    );
                    false
                }
            }
        };

        match ctx.get() {
            Some(req_ctx) => match_authority(req_ctx),
            None => {
                let req_ctx: RequestContext = match (ctx, req).try_into() {
                    Ok(rc) => rc,
                    Err(err) => {
                        tracing::debug!(
                            error = %err,
                            "SubdomainTrieMatcher: failed to extract request context",
                        );
                        return false;
                    }
                };
                let is_match = match_authority(&req_ctx)
                if let Some(ext) = ext {
                    ext.insert(req_ctx);
                }
                is_match
            }
        }
    }
}
```

Do feel free to adapt further based on your own knowledge and insights

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+            }
+            Host::Address(_) => {
+                tracing::trace!("SubdomainTrieMatcher: ignoring numeric address");
+                false
+            }
+        }
+    }
+}
+
+impl FromIterator<String> for SubdomainTrieMatcher {
```
GlenDC, line 93: You just need to implement this once really, which is the same as your constructor, also no reason to collect:

```rust
impl<S> FromIterator<S> for SubdomainTrieMatcher
    where S: AsRef<str>,
{
   #[inline]
    fn from_iter<I: IntoIterator<Item = S>>(iter: I) -> Self {
        SubdomainTrieMatcher::new(iter)
    }
}
```

With this you can drop the `impl<'a> FromIterator<&'a str> for SubdomainTrieMatcher {` impl block as well

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+#[cfg(test)]
+mod subdomain_trie_tests {
+    use super::*;
+
+    #[test]
+    fn test_reverse_domain() {
+        assert_eq!(reverse_domain("example.com"), "com.example.");
```
GlenDC, line 115: test also cases like `"localhost"` and `""`

rama-http/src/matcher/subdomain_trie.rs
```diff
@@ -0,0 +1,173 @@
...
+        assert!(matcher.is_match("example.com"));
+        assert!(matcher.is_match(".example.com"));
+        assert!(matcher.is_match("sub.domain.org"));
+        assert!(matcher.is_match("sub.example.com"));
+        assert!(!matcher.is_match("domain.org"));
+        assert!(!matcher.is_match("other.com"));
+    }
+
+    #[test]
+    fn test_from_iterator_string() {
```
GlenDC, line 132: just need one of these as they are all the same, and you can just do it with a `["example.com", "sub.domain.org"]` kind of test

rama-http/src/matcher/subdomain_trie.rs
GlenDC, line 27: add a comment explaining what this function does please

GlenDC (changes requested):
Thank you for your first contribution. It's pretty great. I have a couple of changes to request but other than that it looks good! Thank you for your effort so far already!

rama-http/src/matcher/mod.rs
```diff
@@ -80,6 +84,8 @@ pub enum HttpMatcherKind<State, Body> {
     ///
     /// [`SocketAddr`]: std::net::SocketAddr
     Socket(SocketMatcher<State, Request<Body>>),
+    /// [`SubdomainTrieMatcher`], a matcher based on domain and subdomains using a trie structure.
+    SubdomainTrie(SubdomainTrieMatcher),
```
GlenDC, line 88: Also add methods to `impl<State, Body> HttpMatcher<State, Body>` block: `any_subdomain`, `and_any_subdomain` and `or_any_subdomain`

LED-0102:
Hello @GlenDC. Thank you for being patient with me. Could you please have a look?

GlenDC:
Looks good @LED-0102 can you please run `just qq`?

You might need to install:
- just: <https://just.systems/man/en/packages.html>
- cargo-sort: `cargo install cargo-sort`
- clippy: `rustup component add clippy`

once that is done you can commit and we can try to run CI again.
Your code looks good however so this is probably the last step prior to merging

GlenDC (approved):

GlenDC:
Looks good now! Thx for this great and swift contribution! And welcome to the rama community!

PR #541, opened 2025-05-07
Issue 408/re enable windows in ci / cross.toml and add it back to supported platforms

soundofspace:
- Add windows tests in ci
- Implement missing udp socket traits for windows
- Always use jemalloc for linux and mimalloc for windows
- Refactor fuzz testing and disable it on windows
- Add windows build target back to cross bulding


rama-cli/src/main.rs
```diff
@@ -12,11 +12,11 @@ use cmd::{echo, fp, http, ip, proxy, serve, tls};
 
 pub mod error;
 
-#[cfg(all(not(feature = "mimalloc"), feature = "jemalloc"))]
+#[cfg(unix)]
```
GlenDC, line 15: should be linux, not unix, I think the macos one is fine last time I checked

rama-cli/Cargo.toml
```diff
@@ -46,6 +41,12 @@ tokio-postgres = { workspace = true, features = ["with-serde_json-1", "with-chro
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
+[target.'cfg(unix)'.dependencies]
```
GlenDC, line 44:
```suggestion
[target.'cfg(linux)'.dependencies]
```

.github/workflows/RamaCLIRelease.yaml
```diff
@@ -24,19 +24,12 @@ jobs:
       matrix:
         target:
           - name: x86_64-unknown-linux-gnu
```
GlenDC, line 26: You still need to add `build-release-windows` here

GlenDC (approved):
LGTM, left a couple of things to fix, but once those are done you can merge and close the linked issue. Thx a lot for this!

.github/workflows/RamaCLIRelease.yaml
soundofspace: had that commit still local, thanks for noticing

rama-cli/src/main.rs
soundofspace: Macos one is indeed fine, but actually did this on purpose. Working with jemalloc is actually really useful to debug memory leaks, and would streamline linux/macos development. Do you prefer native for macos, or can we leave it to jemalloc. I'm also planning to add jemalloc OTLP metrics, which would then also work on macos. There are also other options to track heap usage so open to both

.github/workflows/RamaCLIRelease.yaml
```diff
@@ -131,3 +124,46 @@ jobs:
             target/${{ matrix.target }}/release/*.sha256
           prerelease: ${{ contains(github.ref_name, '-') }}
           tag_name: ${{ inputs.tag || github.ref_name }}
+
+  build-release-windows:
+    runs-on: ubuntu-latest
```
GlenDC, line 129: You can build windows on ubuntu?? This doesn't need to be windows?

CONTRIBUTING.md
```diff
@@ -54,4 +54,4 @@ Before you can do this you do require the following to be installed:
 
 Once this is all done you should be able to run `just qa`.
 When all these pass you can be pretty certain that all tests in the GitHub CI step
-will also succeed. The difference still though is that GitHub Action will also run some of these tests on the MSRV and the supported platforms (for now just Linux and MacOS).
+will also succeed. The difference still though is that GitHub Action will also run some of these tests on the MSRV and three platforms in total: MacOS, Linux and Windows.
```
GlenDC, line 57: I seem to remember there are 2 other locations where this is mentioned, but I might be wrong

GlenDC (approved):

.github/workflows/RamaCLIRelease.yaml
soundofspace: We use cross to cross compile, internally it uses docker to do all the magic. I actually didn't create logic but got it from [this reversed commit](https://github.com/plabayo/rama/commit/a12310a08c04bf87d602609e756fa53d887508fb)

CONTRIBUTING.md
soundofspace: tried finding other but this is the only place i found that only mentions linux and macos

PR #546, opened 2025-05-12
Re export bytes

soundofspace:
As discussed in private re-export bytes crate since it's used all over the place and an important crate.

Export in rama-core so we can use it everywhere, also updated all imports to use this one instead


rama-socks5/src/client/udp.rs
```diff
@@ -0,0 +1,192 @@
+use rama_core::bytes::BytesMut;
```
GlenDC, line 1: you were on the socks5 feature branch it seems. Remove these files again please.

rama-socks5/src/proto/client.rs
```diff
@@ -0,0 +1,617 @@
+//! Client implementation of the SOCKS5 Protocol [RFC 1928]
```
GlenDC, line 1: you were on the socks5 feature branch it seems. Remove these files again please.

rama-socks5/src/server/udp/inspect.rs
```diff
@@ -0,0 +1,384 @@
+use std::fmt;
```
GlenDC, line 1: you were on the socks5 feature branch it seems. Remove these files again please.

rama-socks5/src/server/udp/relay.rs
```diff
@@ -0,0 +1,487 @@
+use std::io::ErrorKind;
+
```
GlenDC, line 2: you were on the socks5 feature branch it seems. Remove these files again please.

GlenDC (approved):
Looks good, Once you resolved my only remarks you can merge this into main.

PR #548, opened 2025-05-13
Macro to help reduce boilerplate for builders

soundofspace:


rama-utils/src/macros/mod.rs
```diff
@@ -233,6 +233,37 @@ macro_rules! __generate_field_setters {
...
+            $vis:vis fn $fn_name:ident(&mut $self_token:ident, $param_name:ident: $param_ty:ty) -> &mut Self {
+                $($body:tt)*
+            }
+        )*
+    ) => {
+        rama_utils::macros::paste! {
+            $(
+                $(#[$outer_doc])*
+                ///
+                /// This is a variant of $fn_name that consumes Self
```
soundofspace, line 251: We could add anything we want to here explain that this is an alternative to set_ but cant really link to that function

rama-utils/src/macros/mod.rs
```diff
@@ -233,6 +233,37 @@ macro_rules! __generate_field_setters {
 
 pub use crate::__generate_field_setters as generate_field_setters;
 
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __generate_set_and_with {
+    (
+        $(
+            $(#[$outer_doc:meta])*
+            $vis:vis fn $fn_name:ident(&mut $self_token:ident, $param_name:ident: $param_ty:ty) -> &mut Self {
```
soundofspace, line 242: default implemention now is set, but we could also do with_ if that is more readable?

rama-net/src/client/pool.rs
```diff
@@ -464,7 +464,16 @@ impl<S, P, R> PooledConnector<S, P, R> {
         }
     }
 
-    generate_field_setters!(wait_for_pool_timeout, Duration);
+    generate_set_and_with!(
+        /// Set timeout after which requesting a connection from the pool will timeout
+        ///
+        /// If no timeout is specified there will be no limit, this could be dangerous
+        /// depending on how many users are waiting for a connection
+        pub fn wait_for_pool_timeout(&mut self, timeout: impl Into<Option<Duration>>) -> &mut Self {
```
soundofspace, line 472: we could a prefix here eg `pub fn set_and_with wait_for_pool_timeout` but dont really think that adds a lot of benefit? `pub fn set_wait_for_pool_timeout` would be nice, but not really possible with macro rules, so that would require proc macro and dont really think it adds a lot of benefit?

GlenDC: agreed

GlenDC, line 472: pub should probably be changed to a vis specifier instead

rama-utils/src/macros/mod.rs
GlenDC: same for me, the `with` case is less to write though :P

rama-utils/src/macros/mod.rs
GlenDC: I don't think you expanded this as I'm pretty certain this will not really put in `$fn_name` here but rally just show up as `$fn_name` (not expanding)

if I'm right you can just drop this, I think that's clear from signature anyway

rama-net/src/client/pool.rs
GlenDC, line 472: your little option trick only works if you do not want to also support an actual impl Into type though as these kind of traits aren't transitive AFAIK. E.g. : https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c08fed46d49bcc6b6f554a20cebba99f

GlenDC (commented):
Left some feedback already

rama-utils/src/macros/mod.rs
soundofspace: Yea this was to show that we could add extra string, but indeed not interpolation, so also dont really see the point, but was more as an fyi

rama-net/src/client/pool.rs
soundofspace: Yea so far I haven't see a case where we have want impl and option, but I'm sure they exist, we could make an option variant of this macro to also generate maybe_with, but feels like we then also want to support maybe_set?

Not entirely sure about this honestly this macro (or macros) is definitely not a solve all but would be nice if it would support most common use cases while still being pretty readable

soundofspace: Other option is we implement the transitive into options directly where needed

GlenDC: maybe_set isn't needed because `set` allows you to branch. The reason `maybe_with` was there often is because it allows you to chain even when consuming. Because with consumes you cannot easily branch without doing awkward stuff.

I agree with your sentiment, but I would probably stay way from the `impl` magic as it has a lot of implications. And perhaps just add a case to check first if it's `Option<T>` and genearte the `maybe_with` that one and use Some() for the other cases (with/set). That is less magical and still gives your clean / easy macro.

rama-http-backend/src/client/mod.rs
```diff
@@ -165,25 +166,16 @@ impl EasyHttpWebClient {
...
-    /// Replace this [`EasyHttpWebClient`] with an option of [`TlsConfig`] set.
-    pub fn maybe_with_tls_connector_config(mut self, layer: Option<TlsConnectorConfig>) -> Self {
-        self.tls_connector_config = layer;
-        self
-    }
+    generate_set_and_with!(
+        /// Set the [`TlsConnectorLayer`] that this [`EasyHttpWebClient`] will use.
+        pub fn tls_connector_config(
+            mut self,
+            layer: impl Into<Option<TlsConnectorConfig>>,
```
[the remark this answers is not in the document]
soundofspace: Yea this was just to show that is is possible, but indeed not something we want to do in the end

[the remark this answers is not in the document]
GlenDC: Cool. Than I think we're good to go. I'm fine with you just merging this PR already (after fixing this) and doing bigger migrations in a follow up. Or you can do  acouple more here if you prefer that. But would only do that if you see some other cases that you didn't test yet. Otherwise I would keep this one small enough

GlenDC (approved):

PR #550, opened 2025-05-16
HttpRequestParts trait to combine http::Request and Parts

soundofspace:
`HttpRequestParts` trait provides a single view over `http::Request` and `http::request::Parts`. This allows us to simplify some TryFrom implementations, and can probably be useful in other places to.

This was previously discussed on this PR: https://github.com/plabayo/rama/pull/491#discussion_r2079867624

### DIFF
--- a/rama-http-types/src/request.rs
+++ b/rama-http-types/src/request.rs
@@ -1,14 +1,21 @@
-use crate::body::Body;
-use http::{Extensions, HeaderMap, HeaderValue, Method, Uri, Version, request::Parts};
+use crate::{
+    body::Body,
+    dep::http::{Extensions, HeaderMap, HeaderValue, Method, Uri, Version, request::Parts},
+};
 
-/// Type alias for [`http::Request`] whose body type
+/// Type alias for [`HttpRequest`] whose body type
 /// defaults to [`Body`], the most common body type used with rama.
+///
+/// [`HttpRequest`]: crate::dep::http::Request
 pub type Request<T = Body> = http::Request<T>;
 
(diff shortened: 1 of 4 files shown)


rama-http-types/src/request.rs
```diff
@@ -0,0 +1,107 @@
+use crate::body::Body;
+use http::{Extensions, HeaderMap, HeaderValue, Method, Uri, Version, request::Parts};
+
+/// Type alias for [`http::Request`] whose body type
+/// defaults to [`Body`], the most common body type used with rama.
+pub type Request<T = Body> = http::Request<T>;
+
+/// [`HttpRequestParts`] is used in places where don't need the [`Body`] of the [`http::Request`]
```
GlenDC, line 8: `Body` here should link to the http_body::Body from `crate::dep`

ALso try to link to `crate::Request` instead of `http::` versions, so we keep it in crate.

rama-http-types/src/request.rs
```diff
@@ -0,0 +1,107 @@
+use crate::body::Body;
+use http::{Extensions, HeaderMap, HeaderValue, Method, Uri, Version, request::Parts};
```
GlenDC, line 2: I know IDE's don't help here, but try to use the `crate::` versions please so we only have these external imports in 1 location. For stuff we do not re-export you do of course use them directly.

GlenDC (approved):
Looks great. Thanks for that as always.
2 nitpicks, but once those are resolved you can just ship it :)

When linking to something with a long import path in docs use the syntax:

```
blabla [`Body`]

...

[`Body`]: crate::dep::http_body::Body
```

GlenDC (approved):

PR #551, opened 2025-05-16
add auto Tls acceptor service (== TlsPeekRouter)

### LINKED ISSUE #547: add auto Tls acceptor service
It can be useful for a web or MITM service to be able to handle both HTTP and HTTPS traffic. This would basically peek the incoming early bytes to detect tls traffic vs http traffic. This is not a perfect process but it is in practice reliable.

Once implemented we can adapt the `socks5_connect_proxy_mitm_boring` example to also support https traffic as for now that is not a thing.

GlenDC:
Add initial impl of `TlsPeekRouter`, incl unit tests of its core logic.
And integrate this new `TlsPeekRouter`into socks5 MITM example.

Closes #547

### DIFF
--- a/rama-net/src/tls/server/peek.rs
+++ b/rama-net/src/tls/server/peek.rs
@@ -1,0 +1,249 @@
+use rama_core::{
+    Context, Service,
+    error::{BoxError, ErrorContext},
+    service::RejectService,
+};
+use std::{fmt, io::IoSlice, pin::Pin, task::Poll};
+use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, ReadBuf};
+
+/// A [`Service`] router that can be used to support
+/// tls traffic as well as non-tls traffic.
+///
+/// By default non-tls traffic is rejected using [`RejectService`].
+/// Use [`TlsPeekRouter::with_fallback`] to configure the fallback service.
+pub struct TlsPeekRouter<T, F = RejectService<(), NoTlsRejectError>> {
+    tls_acceptor: T,
+    fallback: F,
+}
+
+rama_utils::macros::error::static_str_error! {
+    #[doc = "non-tls connection is rejected"]
+    pub struct NoTlsRejectError;
+}
+
+impl<T> TlsPeekRouter<T> {
+    /// Create a new [`TlsPeekRouter`].
+    pub fn new(tls_acceptor: T) -> Self {
+        Self {
+            tls_acceptor,
+            fallback: RejectService::new(NoTlsRejectError),
+        }
+    }
+
+    /// Attach a fallback [`Service`] tp this [`TlsPeekRouter`].
+    pub fn with_fallback<F>(self, fallback: F) -> TlsPeekRouter<T, F> {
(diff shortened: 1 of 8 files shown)


rama-net/src/tls/server/peek.rs
```diff
@@ -0,0 +1,311 @@
...
+        let is_tls = n == TLS_HEADER_PEEK_LEN && matches!(peek_buf, [0x16, 0x03, 0x00..=0x04, ..]);
+        tracing::trace!(%is_tls, "tls prefix header read");
+
+        let offset = TLS_HEADER_PEEK_LEN - n;
+        if offset > 0 {
+            tracing::trace!(
+                %n,
+                "move tls peek buffer cursor due to reading not enough"
+            );
+            for i in offset..0 {
```
soundofspace, line 91: This offset logic in general looks weird, and from looking at this code this for loop will never trigger?

soundofspace (changes requested):
Looks great except for one weird thing

rama-net/src/tls/server/peek.rs
GlenDC: It does trigger and is needed to ensure we can deal with edge cases where there's < 5 bytes of data. See test for it.

What is weird about it?

soundofspace:
```
TLS_HEADER_PEEK_LEN = 5
stream = "foxa"
n = read() = 4
offset = TLS_HEADER_PEEK_LEN - n = 1

for i in 1..0 -> will never trigger because 1..0 is empty, first number needs to be smaller
```

Removing `if offset > 0` block doesn't affect tests at all since it doesn't do anything.

Also this logic is working, but for the wrong reasons

```rust
impl<S: AsyncRead + Unpin> AsyncRead for TlsPeekStream<S> {
    fn poll_read(
        mut self: Pin<&mut Self>,
        cx: &mut std::task::Context<'_>,
        buf: &mut ReadBuf<'_>,
    ) -> Poll<std::io::Result<()>> {
        // self.prefix = [b"f", b"o", b"x", b"a", 0]
        // self.offset = 1
        if self.offset < TLS_HEADER_PEEK_LEN - 1 {
            let remaining = &self.prefix[self.offset..];
            // remaining = [b"o", b"x", b"a", 0]
            let to_copy = remaining.len().min(buf.remaining());
            // assume buf = 100, so big enough
            // to_copy = 4

            if to_copy > 0 {
                buf.put_slice(&self.prefix[..to_copy]);
                // This slice is from 0 instead of from self.offset
                // and because of that mistake it works if buffer is big enough

                // But if we assume buf size = 2, then the first time we would
                // write [b"f", b"o"] -> self.offset: 1 -> 3. Second run:
                // remaining = [b"a", 0] -> to copy = 2
                // &self.prefix[..to_copy] = [b"f", b"o"] again
                self.offset += to_copy;
                return Poll::Ready(Ok(()));
            }
        }

        Pin::new(&mut self.inner).poll_read(cx, buf)
    }
}
```

PR #552, opened 2025-05-17
improve existing docs and add protocol inspection docs

GlenDC:
Also move sni to its own chapter and improve some more the docs.


soundofspace (approved):
Nice work, fluent to read and concise

PR #553, opened 2025-05-19
fix naming on tls.md

era:
This PR fixes a small mistake on the `tlsh.md` part of the documentation. I imagine that you meant `mutual` instead of `manual`.


GlenDC (approved):
Thank you very much for your first contribution!
Your assessment was spot on, it was indeed what I meant to write.

PR #555, opened 2025-05-19
TLS: Add ALPS extension support

kriztalz:
This PR introduces the parsing of the ALPS extension.

Defined in https://www.ietf.org/archive/id/draft-vvv-tls-alps-01.html, used in Chromium.

Little caveat, Chromium changed the codepoint it uses for this extensions and both codepoints are supported:

https://chromestatus.com/feature/5149147365900288
https://boringssl.googlesource.com/boringssl/+/master/include/openssl/tls1.h?pli=1#112


GlenDC (approved):
Looks good to me. This is about this part of the spec:

```
The client sends an extension in ClientHello that enumerates all ALPN values for which ALPS is supported.
```

thank you for this PR

GlenDC:
You'll have to run `just qq` @kriztalz to fix the linter issues.
Once that is resolved and @soundofspace also approves I'll gladly have this merged in.

GlenDC (approved):

rama-net/src/tls/client/hello.rs
```diff
@@ -126,6 +126,19 @@ impl ClientHello {
         None
     }
 
+    /// Return the application layer protocols supported for negotiation by this client
```
soundofspace, line 129:
```suggestion
    /// Return the application layer protocols for which alps is supported
```

rama-net/src/tls/client/parser.rs
```diff
@@ -132,6 +132,12 @@ fn parse_tls_client_hello_extension(i: &[u8]) -> IResult<&[u8], ClientHelloExten
             let (i, ech) = parse_ech_client_hello(ext_data)?;
             Ok((i, ClientHelloExtension::EncryptedClientHello(ech)))
         }
+        ExtensionId::OLD_APPLICATION_SETTINGS => {
+            parse_tls_extension_application_settings_content(ext_data)
+        }
+        ExtensionId::APPLICATION_SETTINGS => {
+            parse_tls_extension_application_settings_content(ext_data)
+        }
```
soundofspace, line 140:
```suggestion
        ExtensionId::OLD_APPLICATION_SETTINGS | ExtensionId::APPLICATION_SETTINGS => {
            parse_tls_extension_application_settings_content(ext_data)
        }
```

soundofspace (changes requested):
Two minor remarks, but otherwise this looks.

Thanks for helping with this!

soundofspace (approved):

PR #557, opened 2025-05-20
fix sock5 proxy layer in doc

abernardeau-wallix:


abernardeau-wallix:
I think this was a copy/paste typo from my understanding of the sentence.

GlenDC (approved):
Well spot. Seems my mind wandered of there for a second. OSI layers get a bit fuzzy in the real world though. I usually just keep it to application layer and transport layer. But sure, I suppose you can make a point that it lives on the conceptual session layer (5). So fine for me. It's either way better than my mistake, haha.

Thank you for another great fix. You are clearly reading these docs carefully and with joy. For which I'm grateful. Thank you.

abernardeau-wallix:
I made some quick internet research and it seems that the consensus is that a sock5 proxy is in the session layer. Wikipedia has the article in the 'Session layer protocols' category https://en.wikipedia.org/wiki/Category:Session_layer_protocols .
It's true that the rest of the page talks about transport proxies so it's a bit incoherent but the OSI layers are fuzzy nowadays

PR #559, opened 2025-05-20
support socks5+http auto proxy acceptor

### LINKED ISSUE #497: support socks5+http auto proxy acceptor
As a follow up of https://github.com/plabayo/rama/issues/15.

GlenDC:
Closes #497

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -524,6 +524,10 @@ required-features = ["socks5"]
 name = "socks5_udp_associate_framed"
 required-features = ["socks5"]
 
+[[example]]
+name = "socks5_and_http_proxy"
+required-features = ["dns", "socks5", "http-full"]
+
 [profile.release]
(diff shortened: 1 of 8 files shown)


soundofspace (approved):
Nice, I do start to see a pattern.

Code for this one is almost identical to the TlsPeek one and we will probably have more of these in the future. It probably makes sense to make a generic PeekService. These can then still exist but they would be a simple wrapper around the generic one.

If you agree about that I will make an issue, seems like a good first issue as most code is already there.

GlenDC:
> If you agree about that I will make an issue, seems like a good first issue as most code is already there.

As agreed on discord, let's wait with this until we have an actual need for it. For now I would like to continue evaluate both implementations fully isolated based on their usage and user feedback. We can still consider to refactor some common parts into a utility thing, but for now I do not see the need or desire for it.

PR #560, opened 2025-05-21
Issue 391/improve and refactor the crypto story in rama new boring config

soundofspace:
Goal of this PR: Move away from ClientConfig defined in rama-net/tls for Boring Tls config (part of https://github.com/plabayo/rama/issues/391)

Things done to achieve that:
- TlsAcceptorBuilder
  - We pass this builder around so we can do modifications where needed
  - By using "base builders" together with getters for all fields we can chain configs in a cheap way, and it is also easy to do. These base builders are inside an Arc so cloning them is very cheap
  - We only Build acceptor data when we actually need it. This allows use to build the Boring config by using almost only references vs cloning everything like we did before, and allows use to modify and chains these builders as needed
- rama-tls-boring now imports rama-ua when ua emulation is enabled and will use TlsProfile to create a builder with everything needed to apply this profile. 
  - This is a new service/layers to clearly indicate that this is something extra and this service will get more complex overtime
  - This service will append the builder needed to the one currently in context or create a new one if needed
  - This service also has support for global overwrites eg if you want to emulate but always want to force http1 for example. It does this by appending the overwrites to the builder chain
- Extend generate_set_and_with to also support Result types. We can make this macro "much smarter", but for now I prefer the copy paste as it makes it very easy to read and understand

Things not part of this PR
- Completely refactor EasyClient to make use of this. This will be done in https://github.com/issues/assigned?issue=plabayo%7Crama%7C529
- Change what we store in TlsProfile. For now we leave the ClientConfig in there, but this will be changed in the future once we have a better idea of what to store instead

### DIFF
--- a/rama-tls-boring/src/client/connector_data.rs
+++ b/rama-tls-boring/src/client/connector_data.rs
@@ -47,7 +47,6 @@ pub struct TlsConnectorData {
 impl std::fmt::Debug for TlsConnectorData {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("TlsConnectorData")
-            .field("config", &"debug not implemented")
             .field(
                 "store_server_certificate_chain",
                 &self.store_server_certificate_chain,
@@ -67,7 +66,7 @@ impl TlsConnectorData {
 /// Use [`TlsConnectorDataBuilder`] to build a [`TlsConnectorData`] in an ergonomic way
 ///
 /// This builder is very powerful and is capable of stacking other builders. Using it
-/// this way gives each layer the option to modify what is needed in a very efficient way.
+/// this way gives each layer the option to modify what is needed in an efficient way.
 pub struct TlsConnectorDataBuilder {
     base_builders: Vec<Arc<TlsConnectorDataBuilder>>,
     server_verify_mode: Option<ServerVerifyMode>,
@@ -173,7 +172,7 @@ impl TlsConnectorDataBuilder {
     pub fn new_http_auto() -> Self {
         Self::new()
             .with_rama_alpn_protos(&[ApplicationProtocol::HTTP_2, ApplicationProtocol::HTTP_11])
-            .expect("with http 2 and http 1")
+            .expect("with http2 and http1")
     }
 
     pub fn new_http_1() -> Self {
@@ -599,6 +598,10 @@ impl TlsConnectorDataBuilder {
 
 impl std::fmt::Debug for TlsConnectorDataBuilder {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Debug implementation of this struct will print each field, but also the getter of
+        // each field. E.g server_verify_mode will show the value set on this exact builder and
+        // server_verify_mode() will print the output of the getter, this will also go over
+        // the entire chain of base_builders and will show the final value that will be used during build.
         f.debug_struct("TlsConnectorDataBuilder")
             .field("server_verify_mode", &self.server_verify_mode)
             .field("server_verify_mode()", &self.server_verify_mode())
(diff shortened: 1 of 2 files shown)


rama-tls-boring/src/client/connector_data.rs
```diff
@@ -29,69 +31,232 @@ use super::compress_certificate::{BrotliCertificateCompressor, ZlibCertificateCo
 
 use crate::keylog::new_key_log_file_handle;
 
-#[derive(Debug, Clone)]
-/// Internal data used as configuration/input for the [`super::HttpsConnector`].
-///
-/// Created by trying to turn the _rama_ opiniated [`rama_net::tls::client::ClientConfig`] into it.
+#[non_exhaustive]
```
GlenDC, line 34: this is here for nothing?

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -29,69 +31,232 @@ use super::compress_certificate::{BrotliCertificateCompressor, ZlibCertificateCo
...
+    cipher_list: Option<Vec<u16>>,
+    extension_order: Option<Vec<u16>>,
+    alpn_protos: Option<Vec<u8>>,
+    curves: Option<Vec<SslCurve>>,
+    verify_algorithm_prefs: Option<Vec<SslSignatureAlgorithm>>,
+    client_auth: Option<ConnectorConfigClientAuth>,
+    certificate_compression_algorithms: Option<Vec<CertificateCompressionAlgorithm>>,
+    delegated_credential_schemes: Option<Vec<SslSignatureAlgorithm>>,
+    server_name: Option<Host>,
+    => // These types implement copy
```
GlenDC, line 130: lol that's a ugly hack, you'll want to do that cleaner :')

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -29,69 +31,232 @@ use super::compress_certificate::{BrotliCertificateCompressor, ZlibCertificateCo
...
+    generate_set_and_with!(
+        /// TODO
+        pub fn server_verify_mode(mut self, server_verify_mode: Option<ServerVerifyMode>) -> Self {
+            self.server_verify_mode = server_verify_mode;
+            self
+        }
+    );
+
+    generate_set_and_with!(
+        /// TODO
```
GlenDC, line 227: how come your other macros doesn't set this?

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -29,69 +31,232 @@ use super::compress_certificate::{BrotliCertificateCompressor, ZlibCertificateCo
...
+        pub fn alpn_protos(
+            mut self,
+            protos: Option<&[ApplicationProtocol]>,
+        ) -> Result<Self, OpaqueError> {
+            self.alpn_protos = if let Some(protos) = protos {
+                let mut alpn_protos: Vec<u8> = vec![];
```
GlenDC, line 233: with_capacity

GlenDC (approved):
Looks great, I like it .... a lot! Much better.
Left some comments but those are minor, once those are resolved and you are happy as well, feel free to merge

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -240,242 +567,110 @@ impl TlsConnectorData {
...
-    }
-
-    /// Return a reference the desired (SNI) in case it exists
-    pub fn server_name(&self) -> Option<&Host> {
-        self.server_name.as_ref()
+impl std::fmt::Debug for TlsConnectorDataBuilder {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("TlsConnectorDataBuilder")
+            .field("server_verify_mode", &self.server_verify_mode)
+            .field("server_verify_mode()", &self.server_verify_mode())
```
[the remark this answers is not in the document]
soundofspace: Good enough with this comment, or you reckon another approach is needed?

PR #561, opened 2025-05-21
Auto install devtools if needed

soundofspace:
This will make sure all common cargo tools are installed automatically so things like `just qq` work for everyone

### DIFF
--- a/justfile
+++ b/justfile
@@ -44,11 +44,25 @@ test-spec: test-spec-h2
 test-ignored:
 	cargo test --features=cli,telemetry,compression,http-full,proxy-full,tcp,rustls --workspace -- --ignored
 
-qq: lint check clippy doc extra-checks
+qq: ensure-dev-dependencies lint check clippy doc extra-checks
 
 qa: qq test
 
 qa-full: qa hack test-ignored fuzz-60s check-links
+
+ensure-dev-dependencies:
+	@just check-dev-dependencies || just install-dev-dependencies
+
+install-dev-dependencies:
+	cargo install cargo-sort cargo-hack cargo-expand cargo-machete
+
+check-dev-dependencies:
+	#!/usr/bin/env bash
+	installed=$(cargo install --list)
+	echo "$installed" | grep -q "cargo-sort"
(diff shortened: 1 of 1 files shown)


soundofspace:
@GlenDC are those listed enough?

justfile
```diff
@@ -44,12 +44,26 @@ test-spec: test-spec-h2
...
+	@just check-dev-dependencies || just install-dev-dependencies
+
+install-dev-dependencies:
+	cargo install cargo-sort cargo-hack cargo-expand cargo-machete
+
+check-dev-dependencies:
+	#!/usr/bin/env bash
+	installed=$(cargo install --list)
+	echo "$installed" | grep -q "cargo-sort"
+	echo "$installed" | grep -q "cargo-hack"
```
GlenDC, line 63: I would honestly make these commands either generic or just inline them. These auto-installs I like, but they should be part of the cmd where they are used.

E.g. sort should be part of sort, hack part of hack, etc... Otherwise individual commands will still fail. And similar you now install stuff like hack even though that's only used when doing qa-full, etc.

GlenDC (changes requested):
I like where this goes, but it can be better

GlenDC:
While you are at it, can you add `RUSTFLAGS='-D warnings'` when running `clippy` in the just clippy target.
Currently that won't fail yet on warnings, while we do fail on it in CI.

justfile
```diff
@@ -44,26 +49,12 @@ test-spec: test-spec-h2
...
-check-dev-dependencies:
-	#!/usr/bin/env bash
-	installed=$(cargo install --list)
-	echo "$installed" | grep -q "cargo-sort"
-	echo "$installed" | grep -q "cargo-hack"
-	echo "$installed" | grep -q "cargo-expand"
-	echo "$installed" | grep -q "cargo-machete"
-
 upgrades:
     cargo upgrades
```
GlenDC, line 59: this is also not by default installed

GlenDC: Further below there is also:

- watch
- vet
- miri
- bloat
- mdbook

GlenDC (approved):

PR #562, opened 2025-05-21
refactor common peek stream + support haproxy-peek

GlenDC:
also add haproxy to proxy docs,
as that was apparently not yet a thing...


soundofspace (approved):
Nice abstractions, makes it very easy to implement Peek logic for something.

PR #567, opened 2025-05-23
support SNI proxy router

### LINKED ISSUE #549: add SNI proxy router example
Add an example to demonstrate the SNI proxy router use case https://ramaproxy.org/book/proxies/http.html#sni-proxies ; in this example the reverse proxy does not terminate the tls traffic but simply snoops the client hello in order to route to the correct backend.

Once example is added, ensure to link it in that chapter.

GlenDC:
Closes #549


rama-net/src/tls/server/sni.rs
```diff
@@ -0,0 +1,441 @@
...
+                "unexpected read size for client hello handshake data"
+            );
+            return Err(
+                OpaqueError::from_display("missing client hello tls handshake data").into_boxed(),
+            );
+        }
+        let sni = extract_sni_from_client_hello_handshake(&v)
+            .context("parse client hello handshake bytes and extract SNI")?;
+
+        v.shrink_to(read_size);
```
soundofspace, line 145: Dont think shrink here will do anything?

- Vec has size record_size
- record_size = n + TLS_HEADER_PEEK_LEN
- read_size != n -> Error, so read_size == n
- data in vec = TLS_HEADER_PEEK_LEN + read_size
- Shrink will do nothing as len of vec > read_size

soundofspace (changes requested):

rama-net/src/tls/client/parser.rs
```diff
@@ -41,6 +41,87 @@ pub fn parse_client_hello(i: &[u8]) -> Result<ClientHello, OpaqueError> {
...
+    // skip sid
+    let (i, sidlen) = verify(be_u8, |&n| n <= 32).parse(i)?;
+    let (i, _) = cond(sidlen > 0, take(sidlen as usize)).parse(i)?;
+
+    // skip ciphers
+    let (i, ciphers_len) = be_u16(i)?;
+    let (i, _) = take(ciphers_len)(i)?;
+
+    // skip compression algos
+    let (i, comp_len) = be_u8(i)?;
```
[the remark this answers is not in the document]
GlenDC: It's the length of comp algo data. Also used immediately, so don't think there's confusion possible here?

rama-net/src/tls/server/sni.rs
GlenDC: I think i meant to use truncate. That does affect the length of.

GlenDC: Ah I know why it works. Length is what we read. I just reserved. Nvm. I don't need to do anything. Gonna delete the code

GlenDC, line 145:
```suggestion
```

GlenDC:
All feedback applied @soundofspace

soundofspace (approved):

PR #569, opened 2025-05-24
domain trie improvements

GlenDC:
- refactor subdomain trie matcher into rama-net domain trie for generic use (e.g. stuff like production SNI routers also can make use of it, just to give an example)
- make use of this common DomainTrie in:
  - InMemoryDns
  - SubdomainTrieMatcher
- make DnsOverwrites more optimal for cloning (default as its extension)


GlenDC:
Have you seen this one @soundofspace ?

soundofspace:
> Have you seen this one @soundofspace ?

Yea, but not reviewed yet, will do so in a bit

rama-net/src/address/domain_trie.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_reverse_domain() {
+        assert_eq!(reverse_domain("example.com"), "com.example.");
```
soundofspace, line 176: Something I use quite often is FQDN notation so `example.com.` it doesn't look like this is supported by `reverse_domain`

soundofspace (changes requested):
Nice, only one small remark about FQDN support, but otherwise everything looks good

rama-net/src/address/domain_trie.rs
soundofspace: Yea that seems like a good solution for this. In the future we might want to make the destination in more detail but here it doesn't seem like we would want/need this

PR #571, opened 2025-05-26
Issue 529/improve performance of easy http web client and suport custom dns resolver

soundofspace:
Move connector config out of EasyHttpWebClient and do it in EasyHttpWebClientBuilder instead.
- This makes the EasyHttpWebClient very easy to work, while also more flexible, as you can bring your own connector now
- Now we also only create the connector once instead of every time. But connector is now boxed to prevent explosion of generics
- This builder should make it pretty easy to create a connector for most common use cases and can be extended in the future if we want to add more support for special things eg http3 and udp
- Builder also has native support to easily do connection pooling
- Support for custom dns resolver
- Refactor everything to use this new builder. This removes almost all usage of `ClientConfig` and gets us a lot closer to removing that

Some extras:
- Expose things such as ConnectorKindAuto, ConnectorKindSecure, ConnectorKindTunnel in such a way that we can use them in external types, but cant create them
- Some small improvements and fixes
- Manually implement clone for BoxService so not all generics need to have a Clone bound
- Some extra docs for request inspectors that should hopefully make clear what they do and where


examples/http_mitm_proxy_boring.rs
```diff
@@ -80,6 +77,7 @@ use rama::{
         profile::UserAgentDatabase,
     },
 };
+use rama_tls_boring::client::TlsConnectorDataBuilder;
```
GlenDC, line 80: tooling sadly doesn't help here, but please make sure to double check these, especially in outside examples but also in actual lib code. Make sure to use the appropriate re-exports when available.

These `/examples` should be copy paste-able for someone who _only_ has `rama` as its dep and if needed also some third-party dep, but certainly no other rama crate.

examples/http_mitm_proxy_boring.rs
```diff
@@ -245,38 +257,9 @@ async fn http_mitm_proxy(ctx: Context, req: Request) -> Result<Response, Infalli
...
-            }
-            cfg
-        }
-        None => base_tls_cfg,
-    };
-
-    client.set_tls_connector_config(TlsConnectorConfig::Boring(Some(tls_client_config)));
+        ))
+        .build()
+        .boxed();
```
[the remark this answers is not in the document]
soundofspace: at one point I boxed everything to test something, this was a leftover that i forgot to cleanup it seems

rama-tls-boring/src/client/connector.rs
```diff
@@ -464,36 +463,35 @@ where
...
+/// which will work both for plain and secure connections.
+pub struct ConnectorKindAuto;
+
+#[non_exhaustive]
+#[derive(Debug, Clone)]
+/// A connector which can _only_ be used to establish a secure connection,
+/// regardless of the scheme of the request URI.
+pub struct ConnectorKindSecure;
+
+#[non_exhaustive]
```
GlenDC, line 482: non-exhaustive isn't required here if you make host private

GlenDC (changes requested):
Looks good in general, a couple of minor things, but the overall picture I like

examples/http_mitm_proxy_boring.rs
```diff
@@ -231,9 +228,25 @@ async fn http_mitm_proxy(ctx: Context, req: Request) -> Result<Response, Infalli
...
+    {
+        // TODO once we fully support building this from client hello directly remove this unwrap
+        TlsConnectorDataBuilder::try_from(hello).unwrap()
+    } else {
+        TlsConnectorDataBuilder::new_http_auto()
+    };
+    let base_tls_config = base_tls_config.with_server_verify_mode(ServerVerifyMode::Disable);
+
+    let client = EasyHttpWebClient::builder()
+        .with_tls_proxy_support_using_boringssl()
```
GlenDC, line 245: how come this doens't accept a config but the non-proxy one does? I get that you want to have a "default" one using the "default" config, but that goes for the non-proxy version as well, no? next to an alt method that does take a config?

rama-http-backend/src/client/mod.rs
```diff
@@ -440,103 +158,404 @@ where
...
+            EasyHttpWebClientBuilder {
+                connector: TcpConnector::new(),
+                _phantom: PhantomData,
+            }
+        }
+    }
+
+    impl EasyHttpWebClientBuilder<TcpConnector, TransportStage> {
+        /// Add a custom [`DnsResolver`] that will be used by this client
+        pub fn with_dns_resolver<T: DnsResolver + Clone>(
```
GlenDC, line 219: FYI in future we'll also have a dns resolver for the tls connector (e.g. to support fetching pre-knowledge records).; Question will be if we need to therefore see these as 2 separate resolvers, or if we would just move the dns resolver to the client builder level.

You can close this as-is, just letting you know so it can sink in a bit.

rama-http-backend/src/client/mod.rs
```diff
@@ -440,103 +158,404 @@ where
...
+        /// and authority, which should cover most common use cases
+        ///
+        /// Use `wait_for_pool_timeout` to limit how long we wait for the pool to give us a connection
+        ///
+        /// If you need a different pool or custom way to group connection you can
+        /// use [`EasyHttpWebClientBuilder::with_custom_connection_pool()`] to provide
+        /// you own.
+        pub fn with_connection_pool<C>(
+            self,
+            config: HttpPooledConnectorConfig,
```
GlenDC, line 496: Usually I allow this to be `Option` and than inside you would do `unwrap_or_default`, wdyt?

PR #572, opened 2025-05-26
fix diagram alignment issues in SNI proxies doc

abernardeau-wallix:


GlenDC (approved):
Awesome thanks for this!

The broken link normally would have been caught by my daily `lychee` run in CI cronjob, but lychee is atm having issues with github blocks with the new ratelimits introduced by microsoft.

Thank you very much for your continued efforts on our docs, very much appreciated.

PR #573, opened 2025-05-26
Add zstd certificate compression support

### LINKED ISSUE #563: Support Zstd compression in Boring Tls
In boring tls we currently don't support `CertificateCompressionAlgorithm::Zstd`.

There is this comment: `TODO fork boring and implement zstd compression`. Since then we have forked boring, but not yet implemented this.

This issue is to track progress on that, and potentially discuss what is needed for that. @GlenDC do you remember why forking was needed, in other words what changes in boring do we need? If so please add it here, if not no problem whoever works on this issue can also figure that out

maartendeprez:
Depends on ["Add TLSEXT_cert_compression_zstd constant" PR](https://github.com/plabayo/rama-boring/pull/2) on `rama-boring`

Closes #563

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2374,6 +2374,7 @@ dependencies = [
  "tokio-test",
  "tracing",
  "tracing-subscriber",
+ "zstd",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


GlenDC (approved):
LGTM

Can you update these to `0.3.1`: https://github.com/plabayo/rama/blob/20ab495d054841b4df93bbe53b846b91eada3eac/Cargo.toml#L144-L145

Once you do that it should be possible to make this work

PR #574, opened 2025-05-26
fix roadmap links in footnote of the website

abernardeau-wallix:


abernardeau-wallix:
By the way the date is the same for the two milestones (Q3 2025), I'm not sure if it's intended.

GlenDC (approved):
Thx for catching that! Thought I updated all locations. But apparently not.

The quarters are the same yes. That's the aim at least. 0.4 at the end of q3, worst case leaping to q4

PR #576, opened 2025-05-27
Add SSE response support

### LINKED ISSUE #570: Add SSE response support
Currently we don't have built in support for SSE events.

Add server support and an example now that we also have easy client support.

And also improve client support where needed.

GlenDC:
- [x] working implementation
- [x] examples (`http_sse.rs` and `http_sse_json.rs`) + test
- [x] book chapter

Closes #570


docs/book/src/sse.md
```diff
@@ -0,0 +1,123 @@
...
+
+The `rama` support for datastar (<https://ramaproxy.org/book/web_servers.html#datastar>)
+is also built on top of these foundations.
+
+## Custom Logic and Resilience
+
+With Rama you can:
+
+- Use `LastEventId` typed header on the server side to resume lost connections based on the last delivered ID;
+- Send data as Strings, structed Json or custom type implementations, all with automatic newline splitting;
```
soundofspace:
```suggestion
- Send data as Strings, structured Json or custom type implementations, all with automatic newline splitting;
```

rama-http-types/src/body/sse/event.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    generate_set_and_with! {
+        /// Set the event's retry timeout field (`retry:<timeout>`).
+        ///
+        /// Previously set retry will be overwritten.
+        ///
+        /// This sets how long clients will wait before reconnecting if they are disconnected from the
+        /// SSE endpoint. Note that this is just a hint: clients are free to wait for longer if they
+        /// wish, such as if they implement exponential backoff.
+        pub fn retry(mut self, millis: u64) -> Self {
```
soundofspace, line 325: Isn't it easier to pass in a duration instead of millis?

rama-http-types/src/body/sse/event_stream.rs
```diff
@@ -0,0 +1,673 @@
...
+    ///    to the field value. Otherwise, ignore the field.
+    ///
+    /// -> If the field name is "retry"
+    ///    If the field value consists of only ASCII digits, then interpret the field value as
+    ///    an integer in base ten, and set the event stream's reconnection time to that integer.
+    ///    Otherwise, ignore the field.
+    ///
+    /// -> Otherwise
+    ///    The field is ignored.
+    fn add(&mut self, line: RawEventLine) {
```
soundofspace, line 60: Is it safe that this function can panic?

rama-http-types/src/body/sse/event_stream.rs
```diff
@@ -0,0 +1,673 @@
...
+            Err(err) => return Poll::Ready(Some(Err(err))),
+            _ => {}
+        }
+
+        if this.state.is_terminated() {
+            return Poll::Ready(None);
+        }
+
+        loop {
+            match this.stream.as_mut().poll_next(cx) {
```
soundofspace, line 248:
```suggestion
            match ready!(this.stream.as_mut().poll_next(cx)) {
```
ready! macro makes this quite a lot nicer to work with

rama-http-types/src/body/sse/parser.rs
```diff
@@ -0,0 +1,145 @@
...
+    .parse(input)
+    .map(|(input, comment)| {
+        (
+            input,
+            RawEventLine::Comment(comment.strip_prefix(' ').unwrap_or(comment)),
+        )
+    })
+}
+
+#[inline]
```
soundofspace, line 99: why inline these things, doesn't compiler figure this out automatically?

rama-http-types/src/body/sse/server.rs
```diff
@@ -0,0 +1,181 @@
...
+        {
+            self.event = Event::default().try_with_comment(text).context("build default event with comment")?;
+            Ok(self)
+        }
+    }
+
+    generate_set_and_with! {
+        /// Customize the event of the keep-alive message.
+        ///
+        /// Default is an empty comment.
```
soundofspace:
```suggestion
        /// Default is an empty event.
```
this?

rama-http/src/service/web/endpoint/response/sse.rs
```diff
@@ -0,0 +1,262 @@
...
+        S::Error: Into<BoxError>,
+        T: EventDataWrite,
+    {
+        Sse { stream }
+    }
+
+    /// Configure the interval between keep-alive messages.
+    ///
+    /// Defaults to no keep-alive messages.
+    pub fn keep_alive<T, E>(self, keep_alive: KeepAlive<T>) -> Sse<KeepAliveStream<S, T>>
```
soundofspace:
```suggestion
    pub fn with_keep_alive<T, E>(self, keep_alive: KeepAlive<T>) -> Sse<KeepAliveStream<S, T>>
```
We might aswel be consistent everywhere

tests/integration/examples/example_tests/http_sse.rs
```diff
@@ -0,0 +1,59 @@
...
+        index_response
+            .headers()
+            .typed_get::<ContentType>()
+            .map(|ct| ct.mime().eq(&mime::TEXT_HTML_UTF_8))
+            .unwrap_or_default()
+    );
+    let index_content = index_response.try_into_string().await.unwrap();
+    assert!(index_content.contains("new EventSource('/api/events')"));
+
+    // test the atual stream content
```
soundofspace:
```suggestion
    // test the actual stream content
```

soundofspace (changes requested):
Nice never played with SSE before but they look really powerful for specific use cases

rama-http-types/src/body/sse/parser.rs
GlenDC: This doesn't inline perse, it gives a hint. But there's no gurantee and if the compiler deems it not to be good it won't do it.

rama-http-types/src/body/sse/event.rs
GlenDC: easier, no. And also no point, as this really is a millis only kind of thing. Otherwise you would need to truncate implicitly (ugly) or add an error. This way it just cannot be wrong.

rama-http-types/src/body/sse/event_stream.rs
GlenDC: It cannot panic, but you mean the unwrap I suppose. Thing is that the parser already guarantees that this is valid, it's however a private function for a reason. If you want to add guarantees for it you could add a fuzzing test on the higher level function (not the private one), but afaik should be ok

rama-http-types/src/body/sse/event_stream.rs
GlenDC: I know about the ready macro, but I prefer here to do it manually

PR #582, opened 2025-05-29
add initial rama-unix implementation

### LINKED ISSUE #370: first version of rama-unix
For now it will just be there to support UNIX Domain Sockets as transport layer.

In future other unix-only features can be added there as well.

GlenDC:
- [x] initial implementation
- [x] examples
- [x] example tests
- [x] transport network chapter

Closes #370


examples/unix_datagram_codec.rs
```diff
@@ -0,0 +1,143 @@
...
+        std::fs::remove_file(path_b).expect("delete tmp b");
+    }
+
+    async fn ping(
+        socket: &mut UnixDatagramFramed<BytesCodec>,
+        b_addr: UnixSocketAddress,
+    ) -> Result<(), io::Error> {
+        socket.send((Bytes::from(&b"PING"[..]), b_addr)).await?;
+
+        for _ in 0..4usize {
```
soundofspace:
```suggestion
        for _ in 0..4 {
```

examples/unix_socket_http.rs
```diff
@@ -0,0 +1,72 @@
+//! TODO TODO TODO An example to show how to listen on a Unix (domain) socket,
```
soundofspace: TODO part of this PR?

rama-unix/src/unix/client/connector.rs
```diff
@@ -0,0 +1,361 @@
...
+}
+
+impl<T> UnixConnector<(), T> {
+    /// Consume `self` to attach the given `Connector` (a [`UnixStreamConnector`]) as a new [`UnixConnector`].
+    pub fn with_connector<Connector>(
+        self,
+        connector: Connector,
+    ) -> UnixConnector<UnixStreamConnectorCloneFactory<Connector>, T>
+where {
+        UnixConnector::<UnixStreamConnectorCloneFactory<Connector>, T> {
```
soundofspace:
```suggestion
        UnixConnector {
```

rama-unix/src/unix/client/connector.rs
```diff
@@ -0,0 +1,361 @@
...
+            connector_factory: UnixStreamConnectorCloneFactory(connector),
+            target: self.target,
+        }
+    }
+
+    /// Consume `self` to attach the given `Factory` (a [`UnixStreamConnectorFactory`]) as a new [`UnixConnector`].
+    pub fn with_connector_factory<Factory>(self, factory: Factory) -> UnixConnector<Factory, T>
+where {
+        UnixConnector::<Factory, T> {
```
soundofspace:
```suggestion
        UnixConnector {
```

rama-unix/src/unix/client/connector.rs
```diff
@@ -0,0 +1,361 @@
...
+                .context("failed to retrieve peer address of established connection")?,
+        )));
+
+        Ok(EstablishedClientConnection { ctx, req, conn })
+    }
+}
+
+/// Trait used by the `UnixConnector`
+/// to actually establish the [`UnixStream`].
+pub trait UnixStreamConnector: Clone + Send + Sync + 'static {
```
soundofspace: Why is clone needed here? Seems that if you use special factory this bound is not needed

rama-unix/src/unix/frame.rs
```diff
@@ -0,0 +1,247 @@
...
+        let Self {
+            ref socket,
+            ref mut out_addr,
+            ref mut wr,
+            ..
+        } = *self;
+
+        let n = ready!(match out_addr.as_ref().and_then(|a| a.as_pathname()) {
+            Some(path) => socket.borrow().poll_send_to(cx, wr, path),
+            None => socket.borrow().poll_send(cx, wr,),
```
soundofspace:
```suggestion
            None => socket.borrow().poll_send(cx, wr),
```
Is this on purpose?

soundofspace (changes requested):
Some minor remarks but this is a very nice addition!

rama-unix/src/unix/client/connector.rs
GlenDC: True, in TcpStreamConnector we do require that as we clone after creation in case of dual stack etc. But here is indeed not required, removing it

GlenDC:
```suggestion
pub trait UnixStreamConnector: Send + Sync + 'static {
```

rama-unix/src/unix/frame.rs
GlenDC: no most likely because i Had it on multiple lines and the formatter collapsed it, but it leaves trailing commas (always / sometimes)

examples/unix_socket_http.rs
GlenDC: Yeah I often leave todos for myself, but I seemed to have forgotton to clean this one up

GlenDC:
```suggestion
//! An example to show how to listen on a Unix (domain) socket,
```

PR #583, opened 2025-05-30
Replace `TryInto<Body>` impls with `TryFrom`

xlambein:
Rust provides a blanket implementation of `TryInto` whenever `TryFrom` is implemented. It also has an exception to the orphan rule that lets you implement `TryFrom` on foreign types. Because of this, it's recommended to implement `TryFrom` rather than `TryInto`, since it gives you a free implementation of the latter.

### DIFF
--- a/rama-http/src/service/web/endpoint/response/csv.rs
+++ b/rama-http/src/service/web/endpoint/response/csv.rs
@@ -138,19 +138,19 @@ where
     }
 }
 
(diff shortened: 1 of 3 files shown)


GlenDC (approved):
Thanks a lot. Pretty certain we do that in most places correct, but with a codebase of this size sometimes minor imperfections do slip through. Thank you for another wonderful contribution. Take care.

Issue #581, opened 2025-05-29
Detect broken connections inside pool before actual usage


soundofspace:
Adding a generic way to detect broken connections inside our pool before actual usage can be done in a lot of ways. I'm not opening this issue to discuss that, but I'm opening it for how to detect it for our most common use case: a http connector -> tcp connection.

Some options to detect broken tcp connections are:
- Peek for a very short time, this should err if socket is closed or in broken state and should have no side effects. This does add overhead and might make everything a lot slower
- use `take_error` on `TcpStream`. This is cleaner then peeking but has as a side-effect that it also clear the error state, and depending on how/where we use it, it might interfere with Tokio state. Pretty sure if we use this correctly that we should be fine. Again this also adds a bit of overhead

Both of these options can be exposed by adding a new method to `Socket`, eg `is_working` or `is_broken`. This works pretty good for our connector stack except for the `HttpClientService`. Problem there is that the `HttpConnector` transform the stream into a `Connection` which is spawned on the executor and awaited there. By doing it this way we effectively erase our connector type from the `HttpClientService` type. Meaning layers above it don't really know about the underlying connectors (if traits are properly implemented). 

Solving this will require a pretty big change on how we store this. For http2 we can just clone this connection and store it, but for http1 there is no cloning support and we will need to investigate what the best option is there. If we do this we will also as a side effect now have the entire connection stack back in our type. Don't think that is a bad thing and would even argue that it is something we want so we can use conditionally implement traits such as Socket on higher levels. 

@GlenDC what do you think about changing `HttpClientService` to have access to our inner connector stack?

soundofspace:
That is indeed one of the ways I was thinking of doing it, but how would you detect a http connection to be open without actually doing something, which is where the two options from above come in. But that doesn't work since we have to access to the socket since the io object isnt stored somewhere we can access

GlenDC:
> That is indeed one of the ways I was thinking of doing it, but how would you detect a http connection to be open without actually doing something, which is where the two options from above come in. But that doesn't work since we have no access to the socket since the io object isnt stored somewhere we can access

We own the http stack. Easy enough to add an atomic bool to the async owner of the (transport) io? Or however we need to make this work.

soundofspace:
Attempted in #586 to detect broken connections before actual usage in a generic way and these were the results.

- Hyper state machine does not actively/reactively detect broken connection state. Only when it uses the underlying IO to read/write does it detect the broken state
[...]

PR #584, opened 2025-05-30
Support dropping idle connections from the pool

soundofspace:
Drop idle connection from pool to prevent using stale ones.

This + detecting broken connections should make the connection pool pretty robust to use. There is still room for improvements here and there, but this is now +- on par with connection pooling in hyper.

Part of: https://github.com/plabayo/rama/issues/581

### DIFF
--- a/rama-net/src/client/pool.rs
+++ b/rama-net/src/client/pool.rs
@@ -212,7 +212,7 @@ impl<C, ID> FiFoReuseLruDropPool<C, ID> {
 impl<C, ID> Pool<C, ID> for FiFoReuseLruDropPool<C, ID>
 where
     C: Send + 'static,
-    ID: Clone + Send + Sync + PartialEq + 'static,
+    ID: Clone + Send + Sync + PartialEq + Debug + 'static,
 {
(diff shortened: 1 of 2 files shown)


rama-net/src/client/pool.rs
```diff
@@ -213,20 +230,30 @@ where
...
+            let mut timeout_triggered = false;
+            if let Some(timeout) = self.idle_timeout {
+                if pooled_conn.last_used.elapsed() > timeout {
+                    timeout_triggered = true;
+                }
+            }
+
+            // Since we always insert in the beginning we can assume if this connection has been idle for too long
+            // all connections older then this one should also be evicted
+            if timeout_triggered {
```
GlenDC, line 247: can you create a habit of adding more trace calls throughout the code, makes it easy to see at runtime what is happening. E.g. in the case of this function can be nice to know...

- when we drained
- when we reused
- when we created

with some context params for each

etc....

GlenDC (approved):
LGTM, a couple of remarks / questions, but no blockers

PR #585, opened 2025-06-01
Issue 518/ Implement PeetPrint

LED-0102:
Hi @GlenDC,

I’m opening this PR to get your feedback on the my current PeetPrint implementation. I realize that the code quality could be improved and that the docs and inline comments are missing; I plan to add them in a following iteration. Could you please review the changes and share any issues you notice? Also, I need your advice on how to generate the test cases for this feature. The current `expected_peet_str` in tests is not reliable.

Thank you in advance.

### DIFF
--- a/rama-net/src/fingerprint/peet/tls.rs
+++ b/rama-net/src/fingerprint/peet/tls.rs
@@ -1,0 +1,375 @@
+//! PeetPrint implementation for Rama (in Rust).
+//!
+//! PeetPrint is inspired by the custom fingerprint algorithm from TrackMe.
+//! See license information below:
+//!
+//! > Original work from:
+//! >   https://github.com/pagpeter/TrackMe
+//! >
+//! > Licensed under GPLv3.
+//! > See https://github.com/pagpeter/TrackMe/blob/master/LICENSE for license details.
+
+use crate::fingerprint::ClientHelloProvider;
+use crate::tls::client::ClientHelloExtension;
+use crate::tls::{
+    ApplicationProtocol, CertificateCompressionAlgorithm, CipherSuite, ExtensionId,
+    ProtocolVersion, SecureTransport, SignatureScheme, SupportedGroup,
+};
+use itertools::Itertools;
+use rama_core::context::Extensions;
+use std::fmt;
+
+#[derive(Clone)]
+/// Input data for a "peetprint" fingerprint.
+///
+/// Computed using a future `PeetPrint::compute` method.
+pub struct PeetPrint {
+    supported_tls_versions: Vec<ProtocolVersion>,
+    supported_protocols: Vec<ApplicationProtocol>,
+    supported_groups: Vec<SupportedGroup>,
+    supported_signature_algorithms: Vec<SignatureScheme>,
+    psk_key_exchange_mode: Option<u8>,
+    certificate_compression_algorithms: Option<Vec<CertificateCompressionAlgorithm>>,
+    cipher_suites: Vec<CipherSuite>,
+    sorted_extensions: Vec<String>,
+}
+
+impl PeetPrint {
+    pub fn compute(ext: &Extensions) -> Result<Self, PeetComputeError> {
+        let client_hello = ext
+            .get::<SecureTransport>()
+            .and_then(|st| st.client_hello())
+            .ok_or(PeetComputeError::MissingClientHello)?;
+        Self::compute_from_client_hello(client_hello)
+    }
+
+    pub fn compute_from_client_hello(
+        client_hello: impl ClientHelloProvider,
+    ) -> Result<Self, PeetComputeError> {
+        let cipher_suites: Vec<CipherSuite> = client_hello
+            .cipher_suites()
+            .filter(|c| !c.is_grease())
+            .collect();
+
+        if cipher_suites.is_empty() {
+            return Err(PeetComputeError::EmptyCipherSuites);
+        }
+
+        let mut supported_tls_versions = Vec::new();
+        let mut supported_protocols = Vec::new();
+        let mut supported_groups = Vec::new();
+        let mut supported_signature_algorithms = Vec::new();
+        let mut psk_key_exchange_mode = None;
+        let mut certificate_compression_algorithms = Vec::new();
+        let mut sorted_extensions = Vec::new();
+
+        for ext in client_hello.extensions() {
+            let id = ext.id();
+
+            if id.is_grease() {
+                sorted_extensions.push("GREASE".to_owned());
+                continue;
+            }
+
+            sorted_extensions.push(u16::from(id).to_string());
+
+            match ext {
+                ClientHelloExtension::SupportedVersions(versions) => {
+                    supported_tls_versions.extend(versions.iter().copied());
+                }
+                ClientHelloExtension::ApplicationLayerProtocolNegotiation(alpns) => {
+                    for p in alpns {
+                        match p {
+                            ApplicationProtocol::HTTP_10 => {}
+                            ApplicationProtocol::HTTP_11 => {}
+                            ApplicationProtocol::HTTP_2 => {}
+                            _ => {
+                                continue;
+                            }
+                        }
+
+                        supported_protocols.push(p.clone());
+                    }
+                }
+                ClientHelloExtension::SupportedGroups(groups) => {
+                    supported_groups.extend(groups.iter().copied());
+                }
+                ClientHelloExtension::SignatureAlgorithms(algos) => {
+                    supported_signature_algorithms.extend(algos.iter().copied());
+                }
+                ClientHelloExtension::Opaque { id: ext_id, data }
+                    if *ext_id == ExtensionId::from(45) =>
+                {
+                    if data.len() < 2 {
+                        psk_key_exchange_mode = None;
+                    } else {
+                        psk_key_exchange_mode = Some(data[1]);
+                    }
+                }
+                ClientHelloExtension::CertificateCompression(algs) => {
+                    certificate_compression_algorithms.extend(algs.iter().cloned());
+                }
+                _ => {}
+            }
+        }
+
+        sorted_extensions.sort();
+
+        Ok(Self {
+            supported_tls_versions,
+            supported_protocols,
+            supported_groups,
+            supported_signature_algorithms,
+            psk_key_exchange_mode,
+            certificate_compression_algorithms: if certificate_compression_algorithms.is_empty() {
+                None
+            } else {
+                Some(certificate_compression_algorithms)
+            },
+            cipher_suites,
+            sorted_extensions,
+        })
+    }
+
+    #[inline]
+    pub fn to_human_string(&self) -> String {
+        format!("{self:?}")
+    }
+
+    pub fn fmt_as(&self, f: &mut fmt::Formatter<'_>, hash_chunks: bool) -> fmt::Result {
+        let tls_versions_str: String = self
+            .supported_tls_versions
+            .iter()
+            .map(|v| {
+                if v.is_grease() {
+                    "GREASE".to_owned()
+                } else {
+                    u16::from(*v).to_string()
+                }
+            })
+            .collect_vec()
+            .join("-");
+
+        // Process protocols; normalize HTTP values
+        let protos: Vec<String> = self
+            .supported_protocols
+            .iter()
+            .map(|p| {
+                let lower = p.to_string().to_lowercase();
+                if lower == "h2" || lower == "http/2" {
+                    "2".to_owned()
+                } else if lower == "http/1.1" {
+                    "1.1".to_owned()
+                } else if lower == "http/1.0" {
+                    "1.0".to_owned()
+                } else {
+                    lower
+                }
+            })
+            .collect();
+
+        let protos_str = protos.join("-");
+
+        let groups_str = self
+            .supported_groups
+            .iter()
+            .map(|g| {
+                if g.is_grease() {
+                    "GREASE".to_owned()
+                } else {
+                    u16::from(*g).to_string()
+                }
+            })
+            .collect_vec()
+            .join("-");
+
+        let sig_algs_str = self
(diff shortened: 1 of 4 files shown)


rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,375 @@
...
+    ProtocolVersion, SecureTransport, SignatureScheme, SupportedGroup,
+};
+use itertools::Itertools;
+use rama_core::context::Extensions;
+use std::fmt;
+
+#[derive(Clone)]
+/// Input data for a "peetprint" fingerprint.
+///
+/// Computed using a future `PeetPrint::compute` method.
```
GlenDC: what does future mean here? Seems a regular sync method. ALso you can make it clickable by changing it to `[\`PeetPrint::compute\`]` (with the square brackets)

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,375 @@
...
+                        match p {
+                            ApplicationProtocol::HTTP_10 => {}
+                            ApplicationProtocol::HTTP_11 => {}
+                            ApplicationProtocol::HTTP_2 => {}
+                            _ => {
+                                continue;
+                            }
+                        }
+
+                        supported_protocols.push(p.clone());
```
GlenDC: I would already push write this as `&'static str` so that you do not need to validate afterwards anymore. So a `Vec<&'static str>` :) Or if you wanna optimise a bit this can probably be even a `SmallVec<[&'static str; 2]>` or so, as I don't think this will often be more than 2  :P but that's perhaps already a bit too overoptimised.

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,375 @@
...
+        })
+    }
+
+    #[inline]
+    pub fn to_human_string(&self) -> String {
+        format!("{self:?}")
+    }
+
+    pub fn fmt_as(&self, f: &mut fmt::Formatter<'_>, hash_chunks: bool) -> fmt::Result {
+        let tls_versions_str: String = self
```
GlenDC: try to avoid allocating so much, by just writing to it already. So iterate over tls versions and use stuff like `write!(f, "GREASE")`, `write!(f, "{v}")` etc ... Also less collecting into vecs and joining etc... Just write the separators :)

GlenDC: And once you go for this write on the fly approach instead of intermediate allocations you would simply wrap the formatter with an md5 encoder :) Your underlying code in the end just cares about it being an something that you can use `write!` with. You can check ja3.rs to see how you can use a macro to make it work for both a hasher and a formatter, as these are sadly not compattible afaik, unless you find a cleaner way.

GlenDC (changes requested):
Your approach and code doesn't look too shabby. For the license please link to the copy of the license in the plabayo/rama repo

What's the issue for testing? why is it not reliable???

Here is how I would test it:

* add a test function similar to what you have
* collect samples manually by:
   * having a wireshark setup
   * going to https://tls.peet.ws/
   * this way I can copy the client hello from wireshark and the debug + hashed version of the peet print according to the official peetprint website.
   
   Is there something that I miss here?
   Other than that solid job, seems you are heading for a swift victory on this one.

LED-0102:
Hi @GlenDC ,

Thank you for the suggestions. The test isn't reliable because the `expected_peet_str` was set by me only. But I added another test using wireshark as you suggested. Please review the changes and let me know if further adjustments are needed.

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,481 @@
...
+                    }
+                }
+                ClientHelloExtension::SupportedGroups(groups) => {
+                    supported_groups.extend(groups.iter().copied());
+                }
+                ClientHelloExtension::SignatureAlgorithms(algos) => {
+                    supported_signature_algorithms.extend(algos.iter().copied());
+                }
+                ClientHelloExtension::Opaque { id: ext_id, data }
+                    if *ext_id == ExtensionId::from(45) =>
```
GlenDC:
```suggestion
                    if *ext_id == ExtensionId::PSK_KEY_EXCHANGE_MODES =>
```

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,481 @@
...
+
+        sorted_extensions.sort();
+
+        Ok(Self {
+            supported_tls_versions,
+            supported_protocols,
+            supported_groups,
+            supported_signature_algorithms,
+            psk_key_exchange_mode,
+            certificate_compression_algorithms: if certificate_compression_algorithms.is_empty() {
```
GlenDC: I would just also make this is a regular vec, you can use `is_empty()` to check if it's empty or not

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,481 @@
...
+
+impl fmt::Write for Md5Writer<'_> {
+    fn write_str(&mut self, s: &str) -> fmt::Result {
+        self.0.consume(s.as_bytes());
+        Ok(())
+    }
+}
+
+impl PeetPrint {
+    fn write_to_fmt(&self, w: &mut impl fmt::Write) -> fmt::Result {
```
GlenDC, line 172: you can write this a bit less manual and error prone by introducing some helpers:

```rust
fn write_to_fmt(&self, w: &mut impl fmt::Write) -> fmt::Result {
    fn write_joined_with_cb<W, T, F>(w: &mut W, items: &[T], cb: F)
    where
        W: fmt::Write
        F: fn(&mut W, &T) -> fmt::Result,
    {
        let mut iter = items.iter();
        if let Some(first) = iter.next() {
            cb(w, first)?;
            for part in iter {
                write!(w, "-")?;
                cb(w, part)?;
            }
        }
        Ok(())
    }

    fn write_joined<W, T>(w: &mut W, items: &[T]) -> fmt::Result
    where
        W: fmt::Write,
        T: fmt::Display,
    {
        write_joined_with_cb(w, items, |w, t| write!(w, "{t}"))
    }

    fn write_u16_with_grease<W, T, F>(w: &mut impl fmt::Write, items: &[T], is_grease: F) -> fmt::Result
    where
        W: fmt::Write,
        T: Copy + Into<u16>,
        F: fn(&T) -> bool,
    {
        write_joined_with_cb(w, items, |w, t| {
            if is_grease(t) {
                write!(w, "GREASE")
            } else {
                write!(w, "{}", u16::from(*t))
            }
        })
    }

    write_u16_with_grease(w, &self.supported_tls_versions, TlsVersion::is_grease)?;
    write!(w, "|")?;

    write_joined(w, &self.supported_protocols)?;
    write!(w, "|")?;

    write_u16_with_grease(w, &self.supported_groups, NamedGroup::is_grease)?;
    write!(w, "|")?;

    write_u16_with_grease(w, &self.supported_signature_algorithms, SignatureScheme::is_grease)?;
    write!(w, "|")?;

    if let Some(mode) = self.psk_key_exchange_mode {
        write!(w, "{}|", mode)?;
    }

    write_u16_with_grease(w, &comps, CertificateCompressionAlgo::is_grease)?;
    write!(w, "|")?;

    write_u16_with_grease(w, &self.cipher_suites, CipherSuite::is_grease)?;
    write!(w, "|")?;

    write_u16_with_grease(w, &self.sorted_extensions, ExtensionId::is_grease)?;

    Ok(())
}
```

NOTE that I would store sorted extensions just as `ExtensionId`. You can sort it like this:

```rust
sorted_extensions.sort_by_key(|&id| {
    if id.is_grease() {
        return (1, 0, 0);
    }
    let n = u16::from(*id);
    let digits = if n < 10 {
        1
    } else if n < 100 {
        2
    } else if n < 1000 {
        3
    } else if n < 10_000 {
        4
    } else {
        5
    };
    (0, digits, n)
})
```

I didn't test this code though, just wrote it on the spot, so might contain some flaws, but nothing that your tests shouldn't be able to catch (be it with some extra tests)

GlenDC (changes requested):
Looks great, added a couple of remarks, but other than that we are close to hitting an home run here.

How come you have 2 chrome pcap tests btw? could you also try it out with some more browsers to get some more coverage? If not no worries I can add them later myself

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,625 @@
+//! PeetPrint implementation for Rama (in Rust).
+//!
+//! PeetPrint is inspired by the custom fingerprint algorithm from TrackMe.
+//! See license information below:
+//!
+//! > Original work from:
+//! >   https://github.com/pagpeter/TrackMe
+//! >
+//! > Licensed under GPLv3.
+//! > See https://github.com/plabayo/rama/blob/peet-print/docs/thirdparty/licenses/pagpeter-trackme for license details.
```
GlenDC:
```suggestion
//! > See <https://github.com/plabayo/rama/blob/main/docs/thirdparty/licenses/pagpeter-trackme> for license details.
```

You can already prepare the correct link

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,625 @@
...
+                ClientHelloExtension::CertificateCompression(algs) => {
+                    certificate_compression_algorithms.extend(algs.iter().cloned());
+                }
+                _ => {}
+            }
+        }
+
+        // Sorting integer IDs in lexicographical order
+        sorted_extensions.sort_by_key(|&id| {
+            if id.is_grease() {
```
GlenDC, line 125: I understand why you are doing it as my approach would indeed not work... However instead of these heap allocations we can use the assumption that it are all u16 numbers to use an array instead, no?

Something like

```
let mut digits = [0; 5];
    for i in (0..5).rev() {
        digits[i] = (n % 10) as u8;
        n /= 10;
    }
```
(again not tested, but hopefully gets the idea across)

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,625 @@
...
+        {
+            write_joined_with_cb(w, items, |w, t| write!(w, "{}", u16::from(*t)))
+        }
+
+        fn write_u16_with_grease<W, T, F>(w: &mut W, items: &[T], is_grease: F) -> fmt::Result
+        where
+            W: fmt::Write,
+            T: Copy + Into<u16>,
+            F: Fn(&T) -> bool,
+            u16: From<T>,
```
GlenDC, line 211: why are you doing this weird trait bound? the `u16: From<T>`???

rama-net/src/fingerprint/peet/tls.rs
```diff
@@ -0,0 +1,625 @@
...
+        write!(w, "|")?;
+
+        write_u16_with_grease(
+            w,
+            &self.supported_signature_algorithms,
+            SignatureScheme::is_grease,
+        )?;
+        write!(w, "|")?;
+
+        if let Some(mode) = self.psk_key_exchange_mode {
```
GlenDC, line 238: where do you see that this is treated differently when optional btw? Would that not just be a default value or an empty string surrounded by `||`?

GlenDC (changes requested):
couple more requests

rama-net/src/fingerprint/peet/tls.rs
LED-0102: Oh, it was suggested by the compiler, otherwise it wasn't passing

rama-net/src/fingerprint/peet/tls.rs
LED-0102: Yes, my bad. It's a mistake on my part

rama-net/src/fingerprint/peet/tls.rs
GlenDC: Can you share the detailed error Here?

LED-0102: └─$ cargo check --workspace --all-targets --all-features
    Checking rama-net v0.3.0-alpha.1 (/home/darshil/LED/Rust/rama/rama-net)
error[E0277]: the trait bound `u16: From<T>` is not satisfied
   --> rama-net/src/fingerprint/peet/tls.rs:209:37
    |
209 |                     write!(w, "{}", u16::from(*t))
    |                                     ^^^ the trait `From<T>` is not implemented for `u16`
    |
help: consider extending the `where` clause, but there might be an alternative better way to express this requirement
    |
203 |             F: Fn(&T) -> bool, u16: From<T>
    |                              ++++++++++++++

For more information about this error, try `rustc --explain E0277`.
error: could not compile `rama-net` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `rama-net` (lib test) due to 1 previous error

GlenDC: fair enough, in that case you should be able to remove the `+ Into<u16>` trait bound though. You are getting this error because you are using `u16::from` if you for example would do `: u16 = t.into()` then it would work. But anyway I guess `u16::from` is more cleaner.

LED-0102:
Hi @GlenDC,

Sorry for all the trouble. Could you please review it once more?

GlenDC (commented):
Looks good to me now. Would you be willing to also make use of this in the echo and fp service? Or would you prefer me to do that after i merge this PR?

- echo: https://github.com/plabayo/rama/blob/ca1c7ef8525c07b88a49c90d535d901590ac3dfd/src/cli/service/echo.rs#L538
- fp: https://github.com/plabayo/rama/blob/ca1c7ef8525c07b88a49c90d535d901590ac3dfd/rama-cli/src/cmd/fp/data.rs#L413 and related locations

LED-0102:
Hi @GlenDC,

Please review my changes. Have I missed anything?

GlenDC (approved):
LGTM, thanks a lot for driving this home

Issue #588, opened 2025-06-03
Feature Request: Enhance `TcpStreamConnector for SocketAddress` to Support IPv6


ShabbirHasan1:
**Title: Feature Request: Enhance `TcpStreamConnector for SocketAddress` to Support IPv6**

**Is your feature request related to a problem? Please describe.**
This isn't a bug, but rather an enhancement. Currently, the `impl TcpStreamConnector for SocketAddress` primarily uses `SocketOptions::default_tcp()`, which sets up default TCP options. While this works well for IPv4, it doesn't explicitly leverage IPv6 specific `SocketOptions` when an IPv6 `SocketAddress` is provided for binding.https://github.com/plabayo/rama/blob/fa70e5be0cc0d13087088fcdb16c16af150fc1df/rama-tcp/src/client/connect.rs#L75-L92 

**Describe the solution you'd like**
I'd like to propose a small refactoring of the `connect` method within the `impl TcpStreamConnector for SocketAddress`. The goal is to dynamically select the appropriate `SocketOptions` based on the IP address family of the `bind_addr`. Specifically, if the `bind_addr` is an IPv6 address, it would be beneficial to use `SocketOptions::default_tcp_v6()` to ensure proper IPv6 socket configuration, while retaining `SocketOptions::default_tcp()` for IPv4 addresses.

Here's the proposed change:

```rust
impl TcpStreamConnector for SocketAddress {
    type Error = OpaqueError;

    async fn connect(&self, addr: SocketAddr) -> Result<TcpStream, Self::Error> {
        let bind_addr = *self;
        let opts = match bind_addr.ip_addr() {
            IpAddr::V4(_ip) => SocketOptions {
                address: Some(bind_addr),
                ..SocketOptions::default_tcp()
            },
            IpAddr::V6(_ip) => SocketOptions {
                address: Some(bind_addr),
                ..SocketOptions::default_tcp_v6()
            },
        };
        tokio::task::spawn_blocking(move || tcp_connect_with_socket_opts(&opts, addr))
            .await
            .context("wait for blocking tcp bind using provided Socket Address")?
    }
}
```

**Describe alternatives you've considered**
[...]

PR #589, opened 2025-06-03
Issue 588 / Enhance `TcpStreamConnector for SocketAddress` to support IPv6 options

ShabbirHasan1:
Fixes https://github.com/plabayo/rama/issues/588

This pull request addresses the feature request to enhance the `impl TcpStreamConnector for SocketAddress` to dynamically use IPv6 specific `SocketOptions` when binding to an IPv6 address.

Currently, the connect method within this implementation consistently uses `SocketOptions::default_tcp()`. While suitable for IPv4, this doesn't fully leverage the capabilities or best practices for IPv6 socket configuration when an IPv6 `SocketAddress` is used for binding.

This change introduces a check for the IP address family of the `bind_addr`. If the `bind_addr` is an IPv6 address, it now uses `SocketOptions::default_tcp_v6()` to ensure proper IPv6 socket options are applied. For IPv4 addresses, it retains the use of `SocketOptions::default_tcp()`.

The proposed modification is as follows:

```rust
impl TcpStreamConnector for SocketAddress {
    type Error = OpaqueError;

    async fn connect(&self, addr: SocketAddr) -> Result<TcpStream, Self::Error> {
        let bind_addr = *self;
        let opts = match bind_addr.ip_addr() {
            IpAddr::V4(_ip) => SocketOptions {
                address: Some(bind_addr),
                ..SocketOptions::default_tcp()
            },
            IpAddr::V6(_ip) => SocketOptions {
                address: Some(bind_addr),
                ..SocketOptions::default_tcp_v6()
            },
        };
        tokio::task::spawn_blocking(move || tcp_connect_with_socket_opts(&opts, addr))
            .await
            .context("wait for blocking tcp bind using provided Socket Address")?
    }
}
```

This enhancement improves the flexibility and robustness of `rama-tcp` in dual-stack and IPv6-only environments by ensuring that the appropriate socket options are automatically applied based on the binding addresses family. While `impl TcpStreamConnector for Arc<SocketOptions>` exists for manual control, this change streamlines the common `SocketAddress` use case.

I have tested this change locally to ensure it functions as expected. Thank You.

### DIFF
--- a/rama-tcp/src/client/connect.rs
+++ b/rama-tcp/src/client/connect.rs
@@ -77,17 +77,19 @@ impl TcpStreamConnector for SocketAddress {
 
     async fn connect(&self, addr: SocketAddr) -> Result<TcpStream, Self::Error> {
         let bind_addr = *self;
-        tokio::task::spawn_blocking(move || {
-            tcp_connect_with_socket_opts(
-                &SocketOptions {
-                    address: Some(bind_addr),
-                    ..SocketOptions::default_tcp()
-                },
-                addr,
-            )
-        })
-        .await
-        .context("wait for blocking tcp bind using provided Socket Address")?
+        let opts = match bind_addr.ip_addr() {
+            IpAddr::V4(_ip) => SocketOptions {
+                address: Some(bind_addr),
+                ..SocketOptions::default_tcp()
+            },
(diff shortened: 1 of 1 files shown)


GlenDC (approved):
LGTM, if CI passes it can be merged. Thank you for this clean and swift solution!

PR #600, opened 2025-06-09
Only allow hostname when parsing SNI

abernardeau-wallix:
RFC 6066 explicitly say that IPv4 and IPv6 address are not permitted.


abernardeau-wallix:
While working on this I found a surprising fact about `Host`: if not carefully, you can create an `IpAddress` that will not be identified as such. This seem related to `Domain` accepting ip address (only v4).
>   println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ip());
>   println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ipv4());
>   println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ipv6());
>   println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_domain());

will print:
> false
> false
> false
> true


I'm not sure if it's really an issue but the caller may need additional steps use `Host` to discriminate between ip/hostname since the is_ip* functions are only checking types

GlenDC:
> While working on this I found a surprising fact about `Host`: if not carefully, you can create an `IpAddress` that will not be identified as such. This seem related to `Domain` accepting ip address (only v4).
> 
> > println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ip());
> > println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ipv4());
> > println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_ipv6());
> > println!("{}", Host::Name(Domain::from_static("127.0.0.1")).is_domain());
> 
> will print:
> 
> > false
> > false
> > false
> > true
> 
> I'm not sure if it's really an issue but the caller may need additional steps use `Host` to discriminate between ip/hostname since the is_ip* functions are only checking types

The latter is on purpose. However this is not by accident. You are explicitly creating a domain here. If you parse a host the ip would be seen as an ip. But if you create it explicitly as a domain that is your choice. There is no TLD checking here. Neither do I think that’s needed.

rama-net/src/tls/client/parser.rs
```diff
@@ -292,7 +292,11 @@ fn parse_tls_extension_sni_hostname(i: &[u8]) -> IResult<&[u8], Host> {
         .map_err(|_| nom::Err::Error(nom::error::Error::new(i, ErrorKind::Not)))?
         .parse()
         .map_err(|_| nom::Err::Error(nom::error::Error::new(i, ErrorKind::Not)))?;
-    Ok((i, host))
+
+    match host {
+        Host::Address(_) => Err(nom::Err::Error(nom::error::Error::new(i, ErrorKind::Not))),
+        Host::Name(_) => Ok((i, host)),
```
GlenDC: As we anyway are restricting it to names, and the way you do is correct, I do think you can just return domain. As such I suggest that you also replace https://github.com/plabayo/rama/blob/ebc2f0d95b9f57ece821ed9d635ab35b677899ee/rama-net/src/tls/client/hello.rs#L184 to make use of `Domain` instead of Host. As well as any places that it touches and uses. Will be a bit of code clean up but should be relatively simple and clean. This will also make it that the SNI router works directly with domains instead of hosts. etc.

GlenDC (changes requested):
Looks good. Your current progress already fixes the parsing side, but we still expose it as something more than it can be. Given we anyway restrict it to domains (which I think is ocrrect) we might as well get rid of that ambiguity.

rama-net/src/tls/client/parser.rs
abernardeau-wallix: I updated the PR. It was indeed a bit of work but relatively straightforward. The only part that did gives me some second thoughts are the `TlsConnectorData` / `TlsConnectorDataBuilder`. I decided to keep the host here since it seem that the struct are used in multiple contexts.

On my fork I also have started the work to make the `sni` field in `SNIRequest` an Option (in the case of the SNI is not present). It still need some time on it and I suspect those changes will create a nice conflict.

GlenDC: Feel free to do them all as part of this PR, you're the only one working on this code afaik, so makes sense to do it together, we are not that purist about our PR scopes, as long as it is aligned with us up front (like here) that is fine.

examples/tls_boring_dynamic_certs.rs
```diff
@@ -150,17 +148,14 @@ impl DynamicCertIssuer for DynamicIssuer {
...
-            Some(host) => match host {
-                rama::net::address::Host::Name(domain) => {
-                    if domain == &Domain::from_static("example") {
-                        return Ok(self.example_data.clone());
-                    } else if domain == &Domain::from_static("second.example") {
-                        return Ok(self.second_example_data.clone());
-                    }
-                    Ok(self.example_data.clone())
+            Some(domain) => {
+                if domain == &Domain::from_static("example") {
```
GlenDC: https://ramaproxy.org/docs/rama/net/address/struct.Domain.html#impl-PartialEq%3C%26str%3E-for-Domain is implemented so you can just do `domain == "example"` I know it's not your fault but you might as well improve this code while you're at it

examples/tls_rustls_dynamic_certs.rs
```diff
@@ -176,18 +173,15 @@ impl ResolvesServerCert for DynamicIssuer {
...
-                rama::net::address::Host::Name(domain) => {
-                    if domain == &Domain::from_static("example") {
-                        self.example_data.clone()
-                    } else if domain == &Domain::from_static("second.example") {
-                        self.second_example_data.clone()
-                    } else {
-                        self.example_data.clone()
-                    }
+            Some(domain) => {
+                if domain == &Domain::from_static("example") {
```
GlenDC: https://ramaproxy.org/docs/rama/net/address/struct.Domain.html#impl-PartialEq%3C%26str%3E-for-Domain same here

rama-net/src/http/request_context.rs
```diff
@@ -14,12 +14,12 @@ use rama_http_types::{Uri, Version};
 use crate::tls::SecureTransport;
 
 #[cfg(feature = "tls")]
-fn try_get_host_from_secure_transport(t: &SecureTransport) -> Option<Host> {
+fn try_get_host_from_secure_transport(t: &SecureTransport) -> Option<Domain> {
```
GlenDC: can you rename these `try_get_host_*` to `try_get_sni_*` please, probably better name

rama-net/src/tls/client/parser.rs
```diff
@@ -45,9 +45,9 @@ pub fn parse_client_hello(i: &[u8]) -> Result<ClientHello, OpaqueError> {
 ///
 /// Same as [`extract_sni_from_client_hello`] but handles the full handshake bytes, meaning
 /// the client hello record itself and the handshake header bytes in front of it.
-pub fn extract_sni_from_client_hello_handshake(i: &[u8]) -> Result<Host, OpaqueError> {
+pub fn extract_sni_from_client_hello_handshake(i: &[u8]) -> Result<Domain, OpaqueError> {
```
GlenDC: feel free to already in this PR make this `Option<Domain>` and the other code you were planning to do normally in a separate PR, might be easier for you? I'm fine for you to do it later in a follow up PR as well, up to you :)

rama-net/src/tls/client/parser.rs
```diff
@@ -243,14 +243,14 @@ fn parse_tls_client_hello_extension(i: &[u8]) -> IResult<&[u8], ClientHelloExten
...
-    Ok((i, ClientHelloExtension::ServerName(host)))
+    let (i, domain) = parse_tls_extension_sni_host(i)?;
+    Ok((i, ClientHelloExtension::ServerName(domain)))
 }
 
 // struct {
 //     ServerName server_name_list<1..2^16-1>
 // } ServerNameList;
-fn parse_tls_extension_sni_host(i: &[u8]) -> IResult<&[u8], Option<Host>> {
+fn parse_tls_extension_sni_host(i: &[u8]) -> IResult<&[u8], Option<Domain>> {
```
GlenDC: rename to `parse_tls_extension_sni` please

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -991,7 +987,7 @@ impl TlsConnectorDataBuilder {
             delegated_credential_schemes,
             record_size_limit,
             encrypted_client_hello,
-            server_name,
```
GlenDC: this probably should be a Domain too, no? Or is there a requirement that this has to be host??

GlenDC (changes requested):
great progress, a bit more feedback on your current state of things, we're pretty close I think

rama-tls-boring/src/client/connector_data.rs
abernardeau-wallix: Originally I try to change it as well but it has some usage in other part of the API and I was not sure if this was correct. I was still in the middle of the refactoring with a lot of compilation error. 
Now that the main refactoring is done, I can get back to it and try again with fresh eyes.

abernardeau-wallix: With fresh eyes without the stress of the whole other refactoring, there was no need to keep it as a  `Host`

GlenDC (approved):
Looks solid, thanks a lot for this!
Great work.

PR #601, opened 2025-06-10
Issue 595/Feature Request: Dynamic Local IPv4/IPv6 Binding for TcpConnector

### LINKED ISSUE #595: Feature Request: Dynamic Local IPv4/IPv6 Binding for TcpConnector
**Feature Request: Dynamic Local IPv4/IPv6 Binding for `TcpConnector`**

Hi @GlenDC ,

I'm looking into how to dynamically bind a local IPv4/IPv6 address (selected from a pool of available IPv4/IPv6's on the system) to `TcpConnector` instances within Rama's layers and services. My goal is to ensure that every outgoing connection uses a different local IP address, similar to the concept discussed in articles like [https://zu1k.com/posts/tutorials/http-proxy-ipv6-pool/](https://zu1k.com/posts/tutorials/http-proxy-ipv6-pool/).

I've experimented with passing IP addresses via `ctx` and managing state, achieving success in a custom `HTTP/HTTPS connect` proxy server. However, this approach seems to require overriding core implementations, as seen with `socks5`, where I'd need to write custom methods rather than leveraging existing Rama mechanisms.

An ideal solution would be a simpler way to modify the local address (keeping the port as `0` for OS-assigned ephemeral ports) of `self.connector` within most layers and services.

Could you please provide guidance on how to achieve this within Rama's current design? If a straightforward method doesn't exist, I'd appreciate any pointers or ideas that could lead to a potential Pull Request to introduce this functionality.

Thank you for your time and guidance! 🙏

ShabbirHasan1:
### Feat: Introduce `TcpStreamConnectorPool` and `IpCidrConnector` for Advanced TCP Connection Management

**Closes #595**

This Pull Request delivers a comprehensive solution for dynamic local IPv4/IPv6 binding and advanced TCP connection management within `rama`. It introduces `TcpStreamConnectorPool<C>` for efficient connector pooling and load balancing, and `IpCidrConnector` for intelligent source IP address selection from CIDR blocks. This implementation directly addresses the request in issue #595 by providing robust, flexible, and high-performance mechanisms for controlling outgoing TCP connections.

#### Motivation

The necessity for dynamic local IP address binding and intelligent source IP selection is critical for modern network applications. Use cases include sophisticated proxy servers, load balancing across multiple network interfaces, rotating source IPs to avoid rate limiting, geographic distribution of connection sources, and enhancing high-availability with fallback capabilities. This PR provides "batteries-included" yet modular building blocks that align perfectly with Rama's philosophy of offering powerful, reusable components.

#### Documentation & Code Clarity

**This PR proudly features extremely verbose and meticulously detailed documentation lines, which were extensively generated and refined with the assistance of large language models (LLMs).** The deliberate inclusion of such comprehensive explanations for every struct, enum, method, and significant logic block aims to provide unparalleled clarity. This commitment to clarity ensures that the codebase is highly accessible and understandable for all contributors, regardless of their familiarity with specific implementation details or Rama's internal architecture, thereby facilitating easier onboarding, debugging, and future maintenance.

#### Testing

Comprehensive unit tests have been implemented for `TcpStreamConnectorPool` to validate `RoundRobin` and `Random` selection modes, including empty pool handling. `IpCidrConnector` is thoroughly tested for correct IP generation within CIDRs, fallback mechanisms, exclusion list adherence, and behavior with various `IpCidrConExt` types (TTL, Session, Range).

#### To-Do List

- [x] Initial implementation
- [x] Updated code documentations

### DIFF
--- a/rama-tcp/Cargo.toml
+++ b/rama-tcp/Cargo.toml
@@ -29,6 +29,8 @@ rama-net = { workspace = true }
 rama-utils = { workspace = true }
 tokio = { workspace = true, features = ["macros", "net"] }
 tracing = { workspace = true }
+rand = { workspace = true }
+cidr = { workspace = true }
 
 [dev-dependencies]
 
--- a/rama-tcp/Cargo.toml
+++ b/rama-tcp/Cargo.toml
@@ -22,7 +22,6 @@ default = []
 http = ["dep:rama-http-types", "rama-net/http"]
 
 [dependencies]
-cidr = { workspace = true }
 rama-core = { workspace = true }
 rama-dns = { workspace = true }
 rama-http-types = { workspace = true, optional = true }
--- a/rama-tcp/Cargo.toml
+++ b/rama-tcp/Cargo.toml
@@ -22,15 +22,16 @@ default = []
 http = ["dep:rama-http-types", "rama-net/http"]
 
 [dependencies]
+cidr = { workspace = true }
 rama-core = { workspace = true }
 rama-dns = { workspace = true }
 rama-http-types = { workspace = true, optional = true }
 rama-net = { workspace = true }
 rama-utils = { workspace = true }
+rand = { workspace = true }
 tokio = { workspace = true, features = ["macros", "net"] }
 tracing = { workspace = true }
-rand = { workspace = true }
-cidr = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [dev-dependencies]
 
--- a/rama-tcp/src/client/pool.rs
+++ b/rama-tcp/src/client/pool.rs
@@ -86,7 +86,7 @@ impl Debug for PoolMode {
 /// # Type Parameters
 ///
 /// * `C` - The connector type that implements `TcpStreamConnector`. Must be `Clone`
-/// to support efficient connector retrieval.
+///   to support efficient connector retrieval.
 ///
 /// # Thread Safety
 ///
--- a/rama-tcp/src/client/pool.rs
+++ b/rama-tcp/src/client/pool.rs
@@ -91,13 +91,13 @@ impl Debug for PoolMode {
 /// # Thread Safety
 ///
 /// This struct is thread-safe and can be shared across multiple threads. The
-/// internal state is protected by atomic operations (for RoundRobin mode) or
+/// internal state is protected by atomic operations (for `RoundRobin` mode) or
 /// uses thread-local randomization (for Random mode).
 ///
 /// # Performance Characteristics
 ///
-/// - **Random Mode**: O(1) selection with minimal synchronization overhead
-/// - **RoundRobin Mode**: O(1) selection with single atomic operation per selection
+/// - **`Random` Mode**: O(1) selection with minimal synchronization overhead
+/// - **`RoundRobin` Mode**: O(1) selection with single atomic operation per selection
 /// - **Memory**: O(n) where n is the number of connectors in the pool
 #[derive(Debug, Clone)]
 pub struct TcpStreamConnectorPool<C> {
@@ -105,7 +105,7 @@ pub struct TcpStreamConnectorPool<C> {
     mode: PoolMode,
     /// Vector of available connectors. Stored as a Vec for cache-friendly access patterns
     connectors: Vec<C>,
-    /// Cached length to avoid repeated Vec::len() calls in hot paths
+    /// Cached length to avoid repeated `Vec::len()` calls in hot paths
     connector_count: usize,
 }
 
@@ -224,18 +224,18 @@ impl<C: TcpStreamConnector + Clone> TcpStreamConnectorPool<C> {
     ///
     /// # Performance Characteristics
     ///
-    /// - **Random Mode**: O(1) with thread-local RNG, no synchronization
-    /// - **RoundRobin Mode**: O(1) with single atomic operation
+    /// - **`Random` Mode**: O(1) with thread-local RNG, no synchronization
+    /// - **`RoundRobin` Mode**: O(1) with single atomic operation
     /// - **Memory**: Creates a clone of the selected connector
     ///
     /// # Algorithm Details
     ///
-    /// ## Random Mode
+    /// ## `Random` Mode
     /// Uses `thread_rng()` for thread-local randomization, avoiding global RNG
     /// contention. The `choose()` method provides uniform distribution across
     /// all available connectors.
     ///
-    /// ## RoundRobin Mode
+    /// ## `RoundRobin` Mode
     /// 1. Atomically increments the counter using `fetch_add(1, Ordering::Relaxed)`
     /// 2. Uses modulo arithmetic to wrap around: `index % connector_count`
     /// 3. Relaxed ordering is sufficient as we only need monotonic increment
@@ -343,6 +343,7 @@ impl<C: TcpStreamConnector + Clone> TcpStreamConnectorPool<C> {
     ///     // Inspect or validate connector
     /// }
     /// ```
+    #[allow(clippy::iter_without_into_iter)]
     pub fn iter(&self) -> std::slice::Iter<'_, C> {
         self.connectors.iter()
     }
@@ -436,10 +437,6 @@ mod tests {
     use cidr::{Ipv4Cidr, Ipv6Cidr};
     use rama_net::address::SocketAddress;
     use std::str::FromStr as _;
-    use tracing::level_filters::LevelFilter;
-    use tracing_subscriber::{
-        EnvFilter, fmt, layer::SubscriberExt as _, util::SubscriberInitExt as _,
-    };
 
     /// Initializes comprehensive tracing infrastructure for test execution and debugging.
     ///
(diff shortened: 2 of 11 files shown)


rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+/// - **`Random` Mode**: O(1) selection with minimal synchronization overhead
+/// - **`RoundRobin` Mode**: O(1) selection with single atomic operation per selection
+/// - **Memory**: O(n) where n is the number of connectors in the pool
+#[derive(Debug, Clone)]
+pub struct TcpStreamConnectorPool<C> {
+    /// The load balancing strategy used for connector selection
+    mode: PoolMode,
+    /// Vector of available connectors. Stored as a Vec for cache-friendly access patterns
+    connectors: Vec<C>,
+    /// Cached length to avoid repeated `Vec::len()` calls in hot paths
```
GlenDC: `len` method is in rust assumented to be zero-cost, otherwise it's called differently. Please remove these "caches"

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust,ignore
+    /// let connectors = vec![connector1, connector2, connector3];
+    /// let pool = TcpStreamConnectorPool::new_random(connectors);
+    /// ```
+    pub fn new_random(mut connectors: Vec<C>) -> Self {
+        // Shuffle once during initialization for better initial distribution
+        connectors.shuffle(&mut rng());
```
GlenDC: please drop this from random, if your randomness is good it shouldn't matter

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+    /// # Examples
+    ///
+    /// ```rust,ignore
+    /// let connectors = vec![connector1, connector2, connector3];
+    /// let pool = TcpStreamConnectorPool::new_round_robin(connectors);
+    /// // First call returns connector1, second returns connector2, etc.
+    /// ```
+    pub fn new_round_robin(connectors: Vec<C>) -> Self {
+        let index = Arc::new(AtomicUsize::new(0));
+        Self::new(PoolMode::RoundRobin(index), connectors)
```
GlenDC: do shuffle here...

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+    #[inline] // Inline for hot path optimization
+    pub fn get_connector(&self) -> Option<C> {
+        // Early return for empty pool - most common edge case
+        if self.connector_count == 0 {
+            return None;
+        }
+
+        let connector = match &self.mode {
+            PoolMode::Random => {
+                // Use thread-local RNG for better performance in multi-threaded scenarios
```
GlenDC: please do not add so much inline comments everywhere that don't add any value...

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+                self.connectors.get(index)
+            }
+        };
+
+        // Clone the connector for ownership transfer
+        // This allows the pool to be used from multiple contexts simultaneously
+        connector.cloned()
+    }
+
+    /// Returns the number of connectors in the pool.
```
GlenDC: for stuff like this `Returns the number of connectors in the pool.` as a single line comment is honestly enough...

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+    ///
+    /// # Examples
+    ///
+    /// ```rust,ignore
+    /// let pool = TcpStreamConnectorPool::new_random(connectors);
+    /// for connector in pool.iter() {
+    ///     // Inspect or validate connector
+    /// }
+    /// ```
+    #[allow(clippy::iter_without_into_iter)]
```
GlenDC: respect the linter here please and do implement into iter, should be trivial

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
...
+    ///
+    /// ```rust,ignore
+    /// let pool = TcpStreamConnectorPool::new_random(connectors);
+    /// let stream = pool.connect("127.0.0.1:8080".parse()?).await?;
+    /// ```
+    async fn connect(&self, addr: SocketAddr) -> Result<TcpStream, Self::Error> {
+        // Select a connector using the pool's load balancing strategy
+        let connector = self
+            .get_connector()
+            .expect("TcpStreamConnectorPool is empty - no connectors available for connection");
```
GlenDC: don't do that, return an error instead please

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,1078 @@
+//! # TCP Stream Connector Pool
```
GlenDC: don't do this pattern please, rename the file to `client/pool/mod.rs`

rama-tcp/src/client/pool.rs
ShabbirHasan1: Understood. I've removed the explicit "caches" and adjusted the `len` method to align with Rust's idiomatic zero-cost assumption for `len`. The changes have been committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: You are correct, for good randomness, the initial shuffle in `new_random` is indeed unnecessary. I've removed that line to streamline the implementation. The change has been committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: You're right, shuffling the connectors in `new_round_robin` will provide better initial distribution, especially useful for `long-running` processes or when the pool is first created. I've added the shuffle operation there. The change has been committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: Apologies for the overly verbose inline comments. I've gone through and tried to significantly reduced them, keeping only those that genuinely add value and clarity to the code. The updated documentation is now more concise. This has been committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: Agreed. For straightforward cases like `Returns the number of connectors in the pool.`, I've updated the comments to be single-line to maintain conciseness and readability. This change is committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: Thank you for pointing that out. I've addressed the linter warnings and implemented `into_iter` where appropriate to improve API ergonomics and adhere to Rust's common patterns. The changes are committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: Understood. I've refactored the module structure by renaming the file to `client/pool/mod.rs` to better organize the code and align with standard Rust module practices. This change has been committed.

rama-tcp/src/client/pool.rs
ShabbirHasan1: You are absolutely right. I've refactored the code to return a proper `Result` (or `OpaqueError`) instead of panicking or unwrapping, ensuring more robust error handling as per `rama`'s conventions. The updated code is committed.

ShabbirHasan1:
* **Regarding `cidr` removal and `ipnet` usage:**
    Thank you for the guidance. I've already proceeded with removing `cidr` and migrating to `ipnet` _(These changes are committed)_. This involved:
    1.  Updating all imports in `rama-tcp` to use `rama_net::stream::dep::ipnet::...`.
    2.  Replacing `cidr` crate usage with `ipnet` throughout the code.

* **Regarding dropping example and integration test tasks:**
       Noted! I've also removed the tasks related to adding an example file and its integration test from the PR description, as they are no longer needed for this PR. Thank you for clarifying.

rama-tcp/Cargo.toml
```diff
@@ -27,8 +27,10 @@ rama-dns = { workspace = true }
 rama-http-types = { workspace = true, optional = true }
 rama-net = { workspace = true }
 rama-utils = { workspace = true }
+rand = { workspace = true }
 tokio = { workspace = true, features = ["macros", "net"] }
 tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
```
GlenDC, line 33: this should not be needed?

GlenDC:
While we do not support it we will not ever stop someone from using LLMs... That said, if you do use it you still do need to make sure you own the code and take responsibility over it.

- The comments and descriptions in this PR are overly verbose and make it noisy at best and impossible to read at worst...
- Given you do not seem to be able to follow simple instructions in a simple manner show that you let an LLM or set of LLMs do most of the work, as these are notorious for selectively ignoring some comments
- The way you design your API is with little thought which is clearly also a sign of over usage of LLM... as these do not care about such human conventions. E.g. why bother exporting these utility functions?
- There is little thought behind this code other than just get something to work... which is not how I want rama to be designed and neither how it is.

As such I'll close this PR for now. I do not mind top mentor people, I do so with pleasure. But for cases like this I wonder who I am really mentoring, a machine or a human... Where I'm pretty certain the former will anyway not have long term benefits from my mentoring as it probably won't be taken into its training sets anyway...

Again I do not mind the use of LLM. It has always been clear to me you use it heavily. But this PR is just too much. I'm sorry but you'll have to do better. I'll take some time in the next weeks to see how can I reflect this in the contribution guidelines.

Your original idea of the IP pooling and also the CIDR based pool is still solid, but you'll have to really code it and own it. It's okay to ask question etc, but please do not again throw this kind of overly verbose AI-ducktaped-together content our way or I'll have to take more serious measures. Take care either way and the best to you.

PR #603, opened 2025-06-11
Add acme client support

soundofspace:


docs/book/src/web_servers.md
```diff
@@ -99,6 +99,12 @@ Rama provides a rich set of examples demonstrating its web service capabilities.
 - [/examples/http_key_value_store.rs](https://github.com/plabayo/rama/tree/main/examples/http_key_value_store.rs): A key-value store API
 - [/examples/http_web_service_dir_and_api.rs](https://github.com/plabayo/rama/tree/main/examples/http_web_service_dir_and_api.rs): A full web application with API
 
+### ACME to get server certificates
+The following examples show how you can integrate ACME into you webservices (ACME is currently still under heavy development)
```
GlenDC:
```suggestion
The following examples show how you can integrate ACME into you webservices (ACME support in Rama is currently still under heavy development)
```

docs/book/src/web_servers.md
```diff
@@ -99,6 +99,12 @@ Rama provides a rich set of examples demonstrating its web service capabilities.
...
+- [/examples/acme_http_challenge.rs](https://github.com/plabayo/rama/tree/main/examples/http_key_value_store.rs): Authenticate to an acme server using a http challenge
```
GlenDC: not super important but probably best to be consistent in the actual text (so not links or code) regarding the way you write `ACME`. As sometimes you do it lowercase while other times uppercase.

docs/book/src/web_servers.md
```diff
@@ -99,6 +99,12 @@ Rama provides a rich set of examples demonstrating its web service capabilities.
...
+- [/examples/acme_tls_challenge_using_boring.rs](https://github.com/plabayo/rama/tree/main/examples/http_key_value_store.rs): Authenticate to an acme server using a tls challenge using boringssl
+- [/examples/acme_tls_challenge_using_rustls.rs](https://github.com/plabayo/rama/tree/main/examples/http_key_value_store.rs): Authenticate to an acme server using a tls challenge using rustls
```
GlenDC: two times `using` in sentences like these reads a bit awkward

examples/README.md
```diff
@@ -38,6 +38,12 @@ This directory contains example implementations demonstrating various features a
 - [`ws_tls_server.rs`](./ws_tls_server.rs) - Secure WebSocket server example (WSS)
 - [`ws_over_h2.rs`](./ws_over_h2.rs) - Secure WebSocket server example using h2.
 
+### ACME
```
GlenDC, line 41: same feedback as in the book example links

examples/README.md
```diff
@@ -38,6 +38,12 @@ This directory contains example implementations demonstrating various features a
...
+The following examples show how you can integrate ACME into you webservices (ACME is currently still under heavy development)
+- [/examples/acme_http_challenge.rs](https://github.com/plabayo/rama/tree/main/examples/http_key_value_store.rs): Authenticate to an acme server using a http challenge
```
GlenDC: ok to copy paste but know that we do not use absolute links in the README.md one

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
+use rama::{
```
GlenDC, line 34: every example has a nice intro. Please also add an issue (github) link that tracks when this test can be written, as normally every example is so far tested, while this one is not (at least not automatically). This is fine but does need to have a link next to a TODO. And of course add already a proper example intro.

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+    },
+};
+
+use std::{sync::Arc, time::Duration};
+use tokio::time::sleep;
+
+// Default directory url of pebble
+const TEST_DIRECTORY_URL: &str = "https://localhost:14000/dir";
+// Addr on which server will bind to do acme challenge
+const ADDR: &str = "0.0.0.0:5002";
```
GlenDC, line 70: already reserve and use a (free) port in the same ranges as the other example, (e.g. sequential the next available one), unless this is not possible for some reason

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+        .with_tls_support_using_boringssl(Some(tls_config))
+        .build()
+        .boxed();
+
+    let client = AcmeClient::new(TEST_DIRECTORY_URL, client).await.unwrap();
+
+    let account = client
+        .create_account(CreateAccountOptions {
+            terms_of_service_agreed: Some(true),
+            contact: None,
```
GlenDC: `..Default::default()` is nice for this kind of stuff.

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+        .unwrap();
+
+    let authz = order.get_authorizations().await.unwrap();
+
+    let auth = &authz[0];
+    let mut challenge = auth
+        .challenges
+        .iter()
+        .find(|challenge| challenge.r#type == ChallengeType::Http01)
+        .unwrap()
```
GlenDC: I would let this example either return an error and use `.context` or at least use `expect` instead of unwrap for stuff like this.

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+        .to_owned();
+
+    let key_authz = order.create_key_authorization(&challenge).unwrap();
+
+    let path = format!(".well-known/acme-challenge/{}", challenge.token);
+
+    tracing::info!("running service at: {ADDR}");
+
+    let state = Arc::new(ChallengeState {
+        key_authz: key_authz.clone(),
```
GlenDC: any reason for this wonky name "authz"? Might want something more traditional unless this has a special meaning that I miss somehow?

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    let graceful = graceful::Shutdown::default();
+
+    graceful.spawn_task_fn(|guard| async move {
+        let exec = Executor::graceful(guard.clone());
+        HttpServer::auto(exec)
+            .listen_with_state(
+                state,
+                ADDR,
+                (TraceLayer::new_for_http(), CompressionLayer::new()).layer(
```
GlenDC:
```suggestion
                (TraceLayer::new_for_http(), CompressionLayer::new()).into_layer(
```

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+                    WebService::default().get(
+                        &path,
+                        |ctx: Context<Arc<ChallengeState>>| async move {
```
GlenDC:
```suggestion
                        async move |ctx: Context<Arc<ChallengeState>>| {
```

Unless not possible here for some reason, but async closures are a thing since recently

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+                            let mut response = http::Response::new(Body::from(
+                                ctx.state().key_authz.as_str().to_owned(),
+                            ));
+                            let headers = response.headers_mut();
+                            headers.append(
```
GlenDC: there should be a typed header variant available, no?

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+            .await
+            .unwrap();
+    });
+
+    sleep(Duration::from_millis(1000)).await;
+
+    order.notify_challenge_ready(&challenge).await.unwrap();
+
+    order
+        .poll_until_challenge_finished(&mut challenge, Duration::from_secs(30))
```
GlenDC: I wonder if we need `pol_until` in this, that's kinda what `await` implies already. ight just need a slightly different name from `challenge_finished`. e.g. `complete_challenge` or w/e

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+        .await
+        .unwrap();
+
+    let state = order
+        .poll_until_all_authorizations_finished(Duration::from_secs(3))
```
GlenDC: same here, poll_until is kinda assumed when having an async function just need a slightly different name for these methods

examples/acme_http_challenge.rs
```diff
@@ -0,0 +1,167 @@
...
+    key_authz: KeyAuthorization,
+}
+
+fn create_csr() -> CertificateSigningRequest {
+    let key_pair = rcgen::KeyPair::generate().unwrap();
+
+    let params = CertificateParams::new(vec!["example.com".to_owned()]).unwrap();
+
+    let mut distinguished_name = DistinguishedName::new();
+    distinguished_name.push(DnType::CountryName, "US");
```
GlenDC: maybe use Gent Belgium :P and Plabayo as org name

examples/acme_tls_challenge_using_boring.rs
```diff
@@ -0,0 +1,191 @@
+use rama::{
```
GlenDC, line 35: feedback from `examples/acme_http_challenge` also applies here

examples/acme_tls_challenge_using_boring.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    let tls_server_config = ServerConfig {
+        server_auth: ServerAuth::CertIssuer(ServerCertIssuerData {
+            kind: issuer.into(),
+            cache_kind: CacheKind::Disabled,
+        }),
+        application_layer_protocol_negotiation: Some(vec![
+            rama_net::tls::ApplicationProtocol::ACME_TLS,
+        ]),
+        client_verify_mode: rama_net::tls::server::ClientVerifyMode::Disable,
```
GlenDC: can you add a doc to tell if this is always required for a tls challenge or just because we use peble?

examples/acme_tls_challenge_using_boring.rs
```diff
@@ -0,0 +1,191 @@
...
+        expose_server_cert: false,
```
GlenDC: perhaps use `..Default::default()` for all the default values?

examples/acme_tls_challenge_using_boring.rs
```diff
@@ -0,0 +1,191 @@
...
+        key_logger: rama_net::tls::KeyLogIntent::Environment,
+        protocol_versions: None,
+        store_client_certificate_chain: false,
+    };
+
+    let acceptor_data = TlsAcceptorData::try_from(tls_server_config).expect("create acceptor data");
+
+    graceful.spawn_task_fn(|guard| async move {
```
GlenDC: async closures

examples/acme_tls_challenge_using_rustls.rs
```diff
@@ -0,0 +1,180 @@
+use rama::{
```
GlenDC, line 35: feedback from `examples/acme_http_challenge` and the boring variant of the tls challenge also applies here

rama-crypto/Cargo.toml
```diff
@@ -26,8 +26,13 @@ base64 = { workspace = true }
 rama-core = { workspace = true }
 rama-utils = { workspace = true }
 rcgen = { workspace = true }
+rustls-pki-types = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+tokio = { workspace = true, features = ["macros", "net"] }
+tokio-util = { workspace = true }
+tracing = { workspace = true }
```
GlenDC: nope, use from rama_core please

rama-crypto/src/jose/jwk.rs
```diff
@@ -4,35 +4,35 @@ use aws_lc_rs::{
...
 pub struct JWK {
     /// Intended algorithm to be used with this key
-    alg: JWA,
+    pub alg: JWA,
     #[serde(flatten)]
-    key_type: JWKType,
+    pub key_type: JWKType,
     #[serde(skip_serializing_if = "Option::is_none")]
-    r#use: Option<JWKUse>,
+    pub r#use: Option<JWKUse>,
```
GlenDC, line 28: if you make these public, which I do appreciate, I would recommend to at the very least have a summary doc line and perhaps even some extra docs after that still

rama-crypto/src/jose/jwk.rs
```diff
@@ -238,6 +238,45 @@ impl EcdsaKey {
...
+
+#[derive(Serialize)]
+struct EcdsaKeySigningHeaders<'a> {
+    alg: JWA,
+    jwk: &'a JWK,
+}
+
+impl Signer for EcdsaKey {
+    type Signature = Signature;
+
```
GlenDC:
```suggestion
```

rama-crypto/src/jose/jws.rs
```diff
@@ -30,12 +33,12 @@ impl Headers {
         /// to get access to the underlying header map
         pub fn header(
             mut self,
-            name: String,
+            name: impl Into<String>,
```
GlenDC: why this impl into? In a way it's not a bad idea, but it does mean that you can also give it a `&str` afaik which is sometimes frowned upon given it hides the allocation a bit

rama-crypto/src/jose/jws.rs
```diff
@@ -5,8 +5,11 @@ use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
-/// When used with serde this will serialize to null
-pub struct Empty;
+/// When used with serde this will serialize to empty struct
+pub struct Empty {}
```
GlenDC, line 15: why the change from `;` to `{}` .Does that change something? And also usualy you want to make this `non_exhaustive` etc

rama-crypto/src/jose/jws.rs
```diff
@@ -546,21 +553,26 @@ impl JWS {
...
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 struct Signature {
     #[serde(skip_serializing_if = "String::is_empty")]
+    #[serde(default = "Default::default")]
     protected: String,
     #[serde(skip_serializing_if = "Headers::is_none")]
-    #[serde(rename = "name")]
+    #[serde(rename = "headers", default = "Default::default")]
```
GlenDC, line 580: was name a typo or what?

rama-crypto/src/jose/jws.rs
```diff
@@ -819,6 +844,7 @@ mod tests {
 
     impl Verifier for DummyKey {
         type Error = OpaqueError;
+        type Output = ();
         fn verify(
```
GlenDC, line 857:
```suggestion

        fn verify(
```

rama-http-headers/src/common/retry_after.rs
```diff
@@ -48,11 +48,16 @@ impl RetryAfter {
         Self(After::DateTime(time.into()))
     }
 
-    /// Create an `RetryAfter` header with a date value.
+    /// Create an `RetryAfter` header with a delay value (in seconds).
```
GlenDC: if it's in seconds only why accept a duration? Not beter an integer type then and force seconds?

rama-tls-acme/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "rama-tls-acme"
+description = "ACME support for rama (RFC 8555)"
```
GlenDC: leave RFC references to the intro lib.rs docs, not for the package description

rama-tls-acme/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+[![Discord][discord-badge]][discord-url]
+[![Buy Me A Coffee][bmac-badge]][bmac-url]
+[![GitHub Sponsors][ghs-badge]][ghs-url]
+[![Paypal Donation][paypal-badge]][paypal-url]
+
+[crates-badge]: https://img.shields.io/crates/v/rama-tls-acme.svg
+[crates-url]: https://crates.io/crates/rama-tls-acme
+[docs-badge]: https://img.shields.io/docsrs/rama-tls-acme/latest
+[docs-url]: https://docs.rs/rama-tls-acme/latest/rama_http/index.html
```
GlenDC: I guess you were replacing programmatically based on `-` as you missed the `_` part

rama-tls-acme/README.md
```diff
@@ -0,0 +1,49 @@
...
+[paypal-url]: https://www.paypal.com/donate/?hosted_button_id=P3KCGT2ACBVFE
+
+🦙 Rama (ラマ) is a modular service framework for the 🦀 Rust language to move and transform your network packets.
+The reasons behind the creation of rama can be read in [the "Why Rama" chapter](https://ramaproxy.org/book/why_rama).
+
+## rama-tls-acme
+
+ACME support for rama.
+
+Crate used by the end-user `rama` crate and `rama` "http" crate authors alike.
```
GlenDC: http crate authors?

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+
+/// Acme client that will used for all acme operations
+pub struct AcmeClient {
+    https_client: AcmeHttpsClient,
+    directory: server::Directory,
+    nonce: Mutex<Option<String>>,
+}
+
+/// Alias for http client used by acme client
+pub type AcmeHttpsClient = BoxService<(), Request, Response, OpaqueError>;
```
GlenDC: This alias is weirdly named, not sure what to do about it

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+
+impl AcmeClient {
+    /// Create a new acme [`AcmeClient`] for the given directory url and using the provided https client
+    pub async fn new(
+        directory_url: &str,
+        https_client: AcmeHttpsClient,
```
GlenDC: I would just accept a `Service` and box it yourself :) the `into_boxed` is a nop operation if it's already a boxed service

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+            .get(directory_url)
+            .send(Context::default())
+            .await?
+            .try_into_json::<server::Directory>()
+            .await?;
+
+        Ok(Self {
+            https_client,
+            directory,
+            nonce: Mutex::new(None),
```
GlenDC, line 72: how is the `read/write` ratio? if it's mostly read and little write you might want to make it ArcSwap instead

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+    /// available this function will try to fetch a new one
+    pub async fn nonce(&self) -> Result<String, OpaqueError> {
+        if let Some(nonce) = self.nonce.lock().take() {
+            return Ok(nonce);
+        }
+
+        let response = self
+            .https_client
+            .head(&self.directory.new_nonce)
+            .send(Context::default())
```
GlenDC: should this context not come from callee?

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+use serde::Serialize;
+use std::{
+    borrow::Cow,
+    time::{Duration, SystemTime},
+};
+use tokio::time::{sleep, timeout};
+
+/// Acme client that will used for all acme operations
+pub struct AcmeClient {
+    https_client: AcmeHttpsClient,
```
GlenDC: Why actually box this instead of just having a generic?

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+
+        *self.nonce.lock() = Some(Self::get_nonce_from_response(&response)?);
+        Ok(response)
+    }
+}
+
+#[derive(Clone, Debug)]
+/// Enum of popular acme providers and their directory url
+pub enum AcmeProvider {
+    LetsEncrypt(&'static str),
```
GlenDC: Why is this an enum if you anyway have their directory url as a public field? I would either just make it a simple enum or have it not be an enum at all but instead an empty type or so with const's.

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+        payload: Option<&impl Serialize>,
+    ) -> Result<Response, OpaqueError> {
+        self.client.post(url, payload, &self.credentials).await
+    }
+}
+
+/// Wrapped [`Account`] with order info
+pub struct Order<'a> {
+    account: &'a Account<'a>,
+    url: String,
```
GlenDC, line 319: why this not a &str if you already work. with a reference anyway

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+            rama_http::headers::After::DateTime(http_date) => SystemTime::from(http_date)
+                .duration_since(SystemTime::now())
+                .ok(),
+            rama_http::headers::After::Delay(seconds) => Some(Duration::from(seconds)),
+        })
+}
+
+async fn retry_bad_nonce<F, Fut, T>(do_request: F) -> Result<T, ClientError>
+where
+    F: Fn() -> Fut,
```
GlenDC, line 685: unless you have strong reason not to I usually use `FnOnce + Clone` as it still allows `Fn` but allows a bit more

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,664 @@
...
+        do_request().await
+    } else {
+        result
+    }
+}
+
+/// Error type which can be returned by the [`AcmeClient`]
+pub enum ClientError {
+    /// Normal [`OpaqueError`] like we use everywhere else
+    OpaqueError(OpaqueError),
```
GlenDC: Usually I call these `Other` or `Unknown`

rama-tls-acme/src/proto/client.rs
```diff
@@ -0,0 +1,75 @@
...
+use serde::{Deserialize, Serialize};
+use std::fmt::Debug;
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Options used to create a new account, or request an identifier for an existing account, defined in [rfc8555 section 7.3]
+///
+/// [rfc8555 section 7.3]: https://datatracker.ietf.org/doc/html/rfc8555/#section-7.3
+pub struct CreateAccountOptions {
+    pub contact: Option<Vec<String>>,
```
GlenDC, line 16: can you add a doc for each field

src/lib.rs
```diff
@@ -185,6 +185,7 @@
 //! - [`rama-unix`](https://crates.io/crates/rama-unix): Unix (domain) socket support for rama
 //! - [`rama-tcp`](https://crates.io/crates/rama-tcp): TCP support for rama
 //! - [`rama-udp`](https://crates.io/crates/rama-udp): UDP support for rama
+//! - [`rama-tls-acme``](https://crates.io/crates/rama-tls-acme): ACME support for rama
```
GlenDC:
```suggestion
//! - [`rama-tls-acme`](https://crates.io/crates/rama-tls-acme): ACME support for rama
```

rama-tls-acme/src/proto/server.rs
```diff
@@ -0,0 +1,464 @@
...
+pub const REPLAY_NONCE_HEADER: &str = "replay-nonce";
+pub const LOCATION_HEADER: &str = "location";
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// Directy contain all endpoints needed by an acme client, defined in [rfc8555 section 7.1.1]
+///
+/// [rfc8555 section 7.1.1]: https://datatracker.ietf.org/doc/html/rfc8555/#section-7.1.1
+pub struct Directory {
+    pub new_nonce: String,
```
GlenDC, line 19: document these fields please, and in general for these very technical bits it is very welcome do not only have summary line but also extended docs

examples/acme_http_challenge.rs
soundofspace: its short for key_authorization, but will use fill name

examples/acme_http_challenge.rs
soundofspace: yea its a habit i sometimes forget i have

examples/acme_http_challenge.rs
soundofspace: I used poll on purpose here since poll in this context is not async-polling, but polling the server. But yea open to a better name if you know one?

examples/acme_tls_challenge_using_boring.rs
soundofspace: This struct doesnt have default implementation (because of server_auth). I still have to refactor this TLS part (client side is done but server side not yet), and this will be tackled then

examples/acme_tls_challenge_using_boring.rs
soundofspace: Disable here is just the default, like a normal tls server? Either way changed it to Default::default() to avoid confusing

rama-crypto/Cargo.toml
soundofspace: sorry mistake from long time ago, started a long time ago on this

rama-crypto/src/jose/jws.rs
soundofspace: As that not mostly the case if you expose something with `Into` was under the impression that otherwise you would use `AsRef`?

rama-crypto/src/jose/jws.rs
soundofspace: non_exhaustive is not really needed here since this thing will never contain field. Goal of Empty struct is to really be an empty struct. Empty serializes to null with serde_json, while Empty {} serializes to empty json object {}. 

This is quite confusing and for exactly that reason I added them here together with the constants NO_PAYLOAD and EMPTY_PAYLOAD so other people don't make the mistake. I'm guessing I need to add more docs here to explain all that?

rama-crypto/src/jose/jws.rs
soundofspace: Yes that was a type, I do plan to add a lot more tests to this module to avoid further mistakes like this

rama-http-headers/src/common/retry_after.rs
soundofspace: Fair remark, honestly because it was like that (i know bad excuse), and i just "fixed" the doc. Internally this uses `Seconds(Duration)` you reckon that needs changing also?

rama-tls-acme/Cargo.toml
soundofspace: Got inspired by recent websocket PR, you want me to update that one also?
```description = "WebSocket (WS) support for rama (RFC 6455)"```

File: rama-tls-acme/README.md (hunk 31 above)
Comment (soundofspace) [reply rama-tls-acme/README.md]: yea thanks for noticing

File: rama-tls-acme/src/client.rs (hunk 35 above)
Comment (soundofspace) [reply rama-tls-acme/src/client.rs:72]: 50 50 so mutex is better option I think?

File: rama-tls-acme/src/client.rs (hunk 36 above)
Comment (soundofspace) [reply rama-tls-acme/src/client.rs]: I on purpose kept state out of this client since it will be removed very soon. Otherwise this would require most methods to now require ctx as input which makes everything a bit weirder, but can be added if you believe its better

File: rama-tls-acme/src/client.rs (hunk 37 above)
Comment (soundofspace) [reply rama-tls-acme/src/client.rs]: Seemed like boxing here didn't hurt and prevented generic from leaking through entire acme stack, should generic be used instead?

File: examples/acme_http_challenge.rs (hunk 14 above)
Comment (GlenDC) [reply examples/acme_http_challenge.rs]: I would just do `.finish_challenge(...)` as the mechanism how it awaits the server is not important I think to mention here? I also think the duration should not even be part of this? you can make a variant with `finish_challenge_with_timeout` but by default I think it should just be "cancel-safe" and bloc until cancelled or finished.

Because if it's cancel safe one can easily use tokio's timeout directly or something more fancy to make sure this isn't blocked. As that's not unique to this method, there is tons of stuff you might want to limit.

File: rama-crypto/src/jose/jws.rs (hunk 25 above)
Comment (GlenDC) [reply rama-crypto/src/jose/jws.rs]: AsRef is ok if you do not _always_ allocate. But here you _always_ allocate so just like `FnOnce + Clone` hides allocations, this is also an example like that

File: examples/acme_tls_challenge_using_boring.rs (hunk 18 above)
Comment (GlenDC) [reply examples/acme_tls_challenge_using_boring.rs]: ah right.... true that is the default indeed... but yeah better to use default if used and possible

File: rama-tls-acme/src/client.rs (hunk 39 above)
Comment (soundofspace) [reply rama-tls-acme/src/client.rs:319]: url is created in `new_order()` and would be dropped otherwise, so Order takes ownership of it

File: rama-tls-acme/src/client.rs (hunk 37 above)
Comment (GlenDC) [reply rama-tls-acme/src/client.rs]: no I think you're right, fair enough. Not a hot path anyway

File: rama-tls-acme/src/client.rs (hunk 38 above)
Comment (soundofspace) [reply rama-tls-acme/src/client.rs]: Would you just do const then, or still wrap them in something? Started of with enum incase they needed different configs or something but is indeed not needed.

File: rama-crypto/src/jose/jws.rs (hunk 25 above)
Comment (soundofspace) [reply rama-crypto/src/jose/jws.rs]: Yea agree in that case, better to be explicit

File: examples/acme_http_challenge.rs (hunk 14 above)
Comment (soundofspace) [reply examples/acme_http_challenge.rs]: Should it be combined with `notify_challenge_ready` in that case?

File: rama-crypto/src/jose/jws.rs (hunk 26 above)
Comment (GlenDC) [reply rama-crypto/src/jose/jws.rs:15]: yup need to be explained indeed, but weird also but ok w/e

File: rama-http-headers/src/common/retry_after.rs (hunk 29 above)
Comment (GlenDC) [reply rama-http-headers/src/common/retry_after.rs]: I don't see a point to use Duration if you always use seconds indeed
would just use u32 or u64. If you do ever need to to duration for something internal I would do that conversation on the spot

File: rama-tls-acme/Cargo.toml (hunk 30 above)
Comment (GlenDC) [reply rama-tls-acme/Cargo.toml]: yes please, just make sure it's in the lib.rs intro somewhere

File: rama-tls-acme/src/client.rs (hunk 35 above)
Comment (GlenDC) [reply rama-tls-acme/src/client.rs:72]: fair enough, we can always revisit such internal details if needed, good enough for now

File: rama-tls-acme/src/client.rs (hunk 36 above)
Comment (GlenDC) [reply rama-tls-acme/src/client.rs]: in a way that's kinda how rama is used, to allow the client to be customised with context (and soon just extensions as part of request). Perahps not something one requires here, or perhaps it does. Not sure... Not really sure what is best for cases like this. If we cannot come to an agreement or solution I'm fine to just you adding that as something to think about in the Service API redesign

File: rama-tls-acme/src/client.rs (hunk 38 above)
Comment (GlenDC) [reply rama-tls-acme/src/client.rs]: I would just have it as 

```
pub enum AcmeProvider {
    LetsEncrypt,
    SomethingElse,
    Foo,
}
```

If you have something bedies the nromal production one you can still add another entry for that one (e.g. LetsEncryptTest )or w/e... But bit weird to have that str there, especially as it is public, so I can just add there whatever I want even if it's not LetsEncrypt :P

File: examples/acme_http_challenge.rs (hunk 14 above)
Comment (GlenDC) [reply examples/acme_http_challenge.rs]: If there is no value or use case in having this as 2 seperate steps I would just have them together yes... For handshake things you only separate them if there is a legit case to have them as sepaarate steps with intermediae results

File: examples/README.md [hunk 45]
```diff
@@ -38,6 +38,12 @@ This directory contains example implementations demonstrating various features a
 - [`ws_tls_server.rs`](./ws_tls_server.rs) - Secure WebSocket server example (WSS)
 - [`ws_over_h2.rs`](./ws_over_h2.rs) - Secure WebSocket server example using h2.
 
+### ACME
+The following examples show how you can integrate ACME into you webservices (ACME support in Rama is currently still under heavy development)
+- [/examples/acme_http_challenge.rs](./acme_http_challenge.rs): Authenticate to an acme server using a http challenge
```
GlenDC:
```suggestion
- [`acme_http_challenge.rs`](./acme_http_challenge.rs): Authenticate to an acme server using a http challenge
```

examples/README.md
```diff
@@ -38,6 +38,12 @@ This directory contains example implementations demonstrating various features a
...
+- [/examples/acme_tls_challenge_using_boring.rs](./acme_tls_challenge_using_boring.rs): Authenticate to an acme server using a tls challenge backed by boringssl
```
GlenDC:
```suggestion
- [`acme_tls_challenge_using_boring.rs`](./acme_tls_challenge_using_boring.rs): Authenticate to an acme server using a tls challenge backed by boringssl
```

examples/README.md
```diff
@@ -38,6 +38,12 @@ This directory contains example implementations demonstrating various features a
...
+- [/examples/acme_tls_challenge_using_rustls.rs](./acme_tls_challenge_using_rustls.rs): Authenticate to an acme server using a tls challenge backed by rustls
```
GlenDC:
```suggestion
- [`acme_tls_challenge_using_rustls.rs`](./acme_tls_challenge_using_rustls.rs): Authenticate to an acme server using a tls challenge backed by rustls
```

examples/acme_tls_challenge_using_rustls.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    let params = CertificateParams::new(vec!["example.com".to_owned()]).expect("");
+
+    let mut distinguished_name = DistinguishedName::new();
+    distinguished_name.push(DnType::CountryName, "BE");
+    distinguished_name.push(DnType::LocalityName, "Ghent");
+    distinguished_name.push(DnType::OrganizationName, "Plabayo");
+    distinguished_name.push(DnType::CommonName, "example.com");
+
+    params.serialize_request(&key_pair).expect("")
```
GlenDC:
```suggestion
    params.serialize_request(&key_pair).unwrap()
```

examples/acme_tls_challenge_using_rustls.rs
```diff
@@ -0,0 +1,227 @@
...
+        Some(self.key.clone())
+    }
+}
+
+async fn internal_tcp_service_fn<S>(_ctx: Context<()>, _stream: S) -> Result<(), Infallible> {
+    Ok(())
+}
+
+fn create_csr() -> CertificateSigningRequest {
+    let key_pair = rcgen::KeyPair::generate().expect("");
```
GlenDC:
```suggestion
    let key_pair = rcgen::KeyPair::generate().unwrap()
```

rama-crypto/src/jose/jwk.rs
```diff
@@ -238,6 +244,44 @@ impl EcdsaKey {
     pub fn rng(&self) -> &SystemRandom {
```
GlenDC, line 244: How come that `SystemRandom` is part of the key? And why do we need to expose it here? I know it's old code but I only notice now. E.g. why not just use `rand::rng()` when we need it? And why this is like this required here but we do not do it anywhere else in the codebase? Answer might be because (some) other code is wrong (as well), but still something to be questioned.

rama-http-headers/src/common/age.rs
```diff
@@ -47,13 +47,13 @@ impl Age {
...
     pub fn from_secs(secs: u64) -> Self {
-        Self(Seconds::from_secs(secs))
+        Self(Seconds::new(secs))
     }
 
     /// Returns the number of seconds for this `Age` header.
     #[must_use]
     pub fn as_secs(&self) -> u64 {
-        self.0.as_u64()
+        self.0.seconds()
```
GlenDC: It's already the `Seconds` type, not more sense to implement `From<Seconds> for u64` or so? Not a strong opinion about it though, but `seconds.seconds()` feels odd

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,681 @@
...
+use std::{
+    borrow::Cow,
+    time::{Duration, SystemTime},
+};
+use tokio::time::sleep;
+
+/// Acme client that will used for all acme operations
+pub struct AcmeClient<State> {
+    https_client: BoxService<State, Request, Response, OpaqueError>,
+    ctx: Context<State>,
```
GlenDC: This is new, right. That you add this to here. I suppose to address my earlier remark. Still... It's a bit odd to do this as you also store the client at same time... I think it honeslty should just be passed when calling the method. But if it botters you you can just have for every such indirect call `foo` a variant version `foo_with_ctx` where the first just calls the latter with `Context::default)`. Don't think we should start storing these in the struct.

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,681 @@
...
+            match challenge.status {
+                server::ChallengeStatus::Pending | server::ChallengeStatus::Processing => (),
+                server::ChallengeStatus::Valid => return Ok(()),
+                server::ChallengeStatus::Invalid => {
+                    return Err(
+                        OpaqueError::from_display("challenge is detected as invalid").into(),
+                    );
+                }
+            }
+
```
GlenDC, line 499: seems like a default (the 1s)  that you may want to provide an opt-in way to configure. Btw why the choice for millis(1000) instead of seconds(1)?

GlenDC (changes requested):
Looks decent. A couple of minor comments. If you can resolve these without any new surprises or without anything that makes you doubt, feel free to merge when done.

rama-tls-acme/src/client.rs
soundofspace: yea being able to configure this is indeed better. 1000 is just since I started with something very low and then changed it without change function signature

rama-crypto/src/jose/jwk.rs
soundofspace: Exposing should not be needed indeed, but goal of storing was to be able to make tests consistent (in the future) and be able to inject a mock implementation

soundofspace: Goal of storing rng is to be able to make all signing/encryption operations consistent in tests (once we add full testing for it). Right now its using `SystemRandom` directly but this can easily be be swapped out for a trait so we can provide it with mock implementations. It's also exposed because we use it in `Signer::sign` which external crates will implement so they also need access to it

examples/acme_tls_challenge_using_rustls.rs
```diff
@@ -0,0 +1,237 @@
...
+    let params =
+        CertificateParams::new(vec!["example.com".to_owned()]).expect("create certificate params");
+
+    let mut distinguished_name = DistinguishedName::new();
+    distinguished_name.push(DnType::CountryName, "BE");
+    distinguished_name.push(DnType::LocalityName, "Ghent");
+    distinguished_name.push(DnType::OrganizationName, "Plabayo");
+    distinguished_name.push(DnType::CommonName, "example.com");
+
+    params.serialize_request(&key_pair).expect("")
```
GlenDC: this is still an empty expect?

rama-tls-acme/src/client.rs
```diff
@@ -0,0 +1,741 @@
...
+    LetsEncryptProduction,
+    LetsEncryptStaging,
+    ZeroSslProduction,
+    GoogleTrustServicesProduction,
+    GoogleTrustServicesStaging,
+}
+
+impl AcmeProvider {
+    #[must_use]
+    pub fn as_str(&self) -> &str {
```
GlenDC: as_str is very unclear of the meaning of the string here. Would rename it as `as_endpoint` or `as_uri` or something like that

rama-tls-acme/src/proto/server.rs
```diff
@@ -0,0 +1,479 @@
...
+/// [rfc8555 section 7.1.1]: https://datatracker.ietf.org/doc/html/rfc8555/#section-7.1.1
+pub struct Directory {
+    /// URL for requesting a new nonce
+    pub new_nonce: String,
+    /// URL for creating a new account
+    pub new_account: String,
+    /// URL for creating a new order
+    pub new_order: String,
+    /// Optional URL for creating new authorization objects (not required by all CAs)
+    pub new_authz: Option<String>,
```
GlenDC, line 28: authz is still in shortened form here

GlenDC (approved):
Looks good, couple of nitpicks but once those are resolved you can merge if you feel confident.

rama-crypto/src/jose/jwk.rs
GlenDC: hmmm ok, seems ok then. Thanks for clarifying.

PR #604, opened 2025-06-12
Dont set span parent

soundofspace:
Setting it manually again default the purpose of manually creating it with `parent: None`


GlenDC (approved):
Thx 😅

PR #608, opened 2025-06-19
refactor(h2): rename `unknown_setting_9` to a meaningful frame setting name

0x676e67:
from: https://github.com/golang/go/issues/69917


0x676e67:
I didn’t change any fields in `SettingsConfig` as that would break semver.

GlenDC:
> I didn’t change any fields in `SettingsConfig` as that would break semver.

You don't need to worry about semver. Everything going into main is for 0.3.

Also I have a PR open that might get merged today that possibly will require some extra work in your branch here. Sorry for that.

0x676e67:
> > I didn’t change any fields in `SettingsConfig` as that would break semver.
> 
> You don't need to worry about semver. Everything going into main is for 0.3.
> 
> Also I have a PR open that might get merged today that possibly will require some extra work in your branch here. Sorry for that.

OK

GlenDC:
https://github.com/plabayo/rama/pull/607 was merged into main, I might need to do a minor follow up commit if I broke something by accident, but it should not affect anything of you as I don't think it will touch the API layer, if anything is needed still at all.

So you should be free and open to finish your PR.

Thanks a lot for your interest btw.

GlenDC:
can you please resolve the conflict, my bad, had to change some stuff to get it working. I am now done and working on something completely different. When this conflict is resolved I'll get it reviewed again and merged in.

GlenDC:
Wasn't aware of this RFC, thx for that. Going to go through it in detail later this week. But just skimming it makes me now get an idea on how come I no longer see priority frames showing up in early h2 frames.

0x676e67:
Priority frames seem to have been abandoned in the future due to the complexity of their implementation. Even the latest version of Firefox has not been seen. This setting should be used to disable priority frame implementation on some servers.

PR #611, opened 2025-06-23
Issue 391/improve and refactor the crypto story in rama: rama-crypto crate and JOSE json support

soundofspace:
- Create rama-crypto crate
  - Will contain all crypto dependencies and in the coming weeks everything will be moved/migrated to there
  - Tls specific logic will stay as is, but for generic tls stuff we will now use rama-crypto and convert these to "native" types only when using them there
  
- Add initial JWK, JWA, JWS logic


rama-crypto/Cargo.toml
```diff
@@ -0,0 +1,41 @@
+[package]
+name = "rama-crypto"
+description = "All crypto logic used by rama"
+version = { workspace = true }
+license = { workspace = true }
+edition = { workspace = true }
+repository = { workspace = true }
+keywords = ["io", "async", "cyrpto", "security", "certificates", "network", "rama"]
```
GlenDC: There's a max of 5 keywords per crate

rama-crypto/README.md
```diff
@@ -0,0 +1,55 @@
...
+[ghs-url]: https://github.com/sponsors/plabayo
+[paypal-badge]: https://img.shields.io/badge/paypal-contribution?style=for-the-badge&color=blue
+[paypal-url]: https://www.paypal.com/donate/?hosted_button_id=P3KCGT2ACBVFE
+
+🦙 Rama (ラマ) is a modular service framework for the 🦀 Rust language to move and transform your network packets.
+The reasons behind the creation of rama can be read in [the "Why Rama" chapter](https://ramaproxy.org/book/why_rama).
+
+## rama-crypto
+
+All cyrpto logic used by rama.
```
GlenDC:
```suggestion
Crypto primitives and dependencies used by rama.
```

rama-crypto/src/jose/jwa.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use crate::jose::JWKEllipticCurves;
+
+#[derive(Debug, Serialize, Deserialize, Copy, Clone, PartialEq, Eq)]
+#[serde(rename_all = "UPPERCASE")]
+/// [`JWA`] or JSON Web Algorithms as defined in [`rfc7518`]
+///
+/// [`rfc7518`]: https://datatracker.ietf.org/doc/html/rfc7518
+pub enum JWA {
+    /// HMAC using SHA-256 (Required)
```
GlenDC, line 23: It is not clear from this doc what required vs optional means

rama-crypto/src/jose/jwk.rs
```diff
@@ -0,0 +1,279 @@
...
+pub struct EcdsaKey {
+    rng: SystemRandom,
+    alg: JWA,
+    inner: EcdsaKeyPair,
+}
+
+impl EcdsaKey {
+    /// Create a new [`EcdsaKey`] from the given [`EcdsaKeyPair`]
+    pub fn new(key_pair: EcdsaKeyPair, alg: JWA, rng: SystemRandom) -> Result<Self, OpaqueError> {
+        // Check if passed algorithm is a correct elliptic curve one
```
GlenDC, line 191: Why not just take an elliptic curve as input and convert it to an algorithm?

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1214 @@
...
+    ///
+    /// Warning: this function will replace already existing headers
+    /// If more control is needed, use `.header_map()` or `.header_map_raw()`
+    /// to get access to the underlying header map
+    pub fn with_header<T: Serialize>(
+        mut self,
+        name: String,
+        value: T,
+    ) -> Result<Self, OpaqueError> {
+        let headers = self.0.get_or_insert(Default::default());
```
GlenDC: or_insert_default...

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1214 @@
...
+        };
+
+        Ok(self)
+    }
+
+    /// Get a mutable reference to the underlying header map
+    ///
+    /// Warning: this will create a header map if one doesn't exist already.
+    /// If you don't want this behaviour using `.header_map_raw()` instead.
+    pub fn header_map(&mut self) -> &mut Map<String, Value> {
```
GlenDC: Probably needs a _mut suffix

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1214 @@
...
+        self.0.get_or_insert(Default::default())
+    }
+
+    /// Get a mutable reference to the underlying optional header map
+    pub fn header_map_raw(&mut self) -> &mut Option<Map<String, Value>> {
```
GlenDC: Why is this needed?

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1214 @@
...
+        self.unprotected_headers.set_header(name, value)?;
+        Ok(self)
+    }
+
+    /// Set provided headers in the unprotected header map
+    ///
+    /// Warning: this function will replace already existing headers
+    /// If more control is use `.unprotected_headers_raw()` to get access
+    /// to the underlying header store
+    pub fn with_unprotected_headers<T: Serialize>(
```
GlenDC: Why not use/expand your macro for these setters

rama-crypto/src/jose/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+//!
+//! * JWE (JSON Web Encryption): This defines a standard way to encrypt data. A JWE ensures
+//!   the confidentiality of the information, making sure only authorized parties can read it.
+//!
+//! * JWK (JSON Web Key): This specifies a JSON format for representing cryptographic keys.
+//!   This makes it simple to share the public keys required to verify signatures or encrypt data.
+//!
+//! * JWA (JSON Web Algorithm): This is essentially a list of the specific cryptographic
+//!   algorithms that are used for signing and encryption within the JOSE framework. The alg
+//!   parameter in the JOSE header identifies which algorithm was used.
```
GlenDC: Refer to rfcs please (links)

rama-crypto/src/jose/jwk.rs
soundofspace: Same issue as JWK::new_from_escdsa_keypair. First [this](https://github.com/aws/aws-lc-rs/pull/834) needs to get merged to get the info needed

rama-crypto/src/jose/jws.rs
soundofspace: For very special use cases, but granted probably something which is not needed, so removed it. Can always add it back if needed

rama-crypto/src/jose/jws.rs
soundofspace: I couldn't really get it to work in a generic way, dont think it's even possible right now 
- https://github.com/rust-lang/rfcs/pull/3442/files
- https://internals.rust-lang.org/t/macro-rules-metavariable-for-generic/18889/5
- https://users.rust-lang.org/t/macro-rules-dealing-with-bound-with-multiple-constraints/64562/5
- https://users.rust-lang.org/t/adding-generics-to-a-macro/64630

But for cases like this we can just use `impl Generic` syntax

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1080 @@
...
+            };
+
+            Ok(self)
+        }
+    }
+
+    /// Get a mutable reference to the underlying header map
+    ///
+    /// Note: this will create a header map if one doesn't exist already.
+    pub fn header_map_mut(&mut self) -> &mut Map<String, Value> {
```
GlenDC: Why does this actually need to be exposed? Can this not be an internal detail. Like why is it not sufficient to be able to just insert/overwrite a header without knowing how it is stored?

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1080 @@
...
+            self.protected_headers.try_set_header(name, value)?;
+            Ok(self)
+        }
+    }
+
+    generate_set_and_with! {
+        /// Set provided headers in the protected header map
+        ///
+        /// Warning: this function will replace already existing headers
+        /// If more control is use `.protected_headers_mut()` to get access
```
GlenDC, line 331: better link to such things properly: [`Self::protected_headers_mut`]. If `Self` doesn't work you'll need to use the type name itself. But this way it is actually link-able.

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1080 @@
...
+/// Decode version of a [`JWS`]
+///
+/// Data here has already been verified, so everything
+/// here is ready for usage
+pub struct DecodedJWS {
+    payload: Vec<u8>,
+    signatures: Vec<DecodedSignature>,
+}
+
+#[derive(Debug)]
```
GlenDC: is there a reason stuff like this are not clone-able?

rama-crypto/src/jose/jws.rs
```diff
@@ -0,0 +1,1080 @@
...
+    protected: Headers,
+    unprotected: Headers,
+    signature: String,
+}
+
+/// A `Signature` which still needs to be checked
+///
+/// It included a String representation of the signed data
+/// so this doesn't need to be re-encoded
+pub struct ToVerifySignature {
```
GlenDC, line 656: At the very least implement debug for it. If there is sensitive stuff, you can redact the sensitive stuff or even just use struct name.

rama-crypto/src/jose/jws.rs
soundofspace: Not really, added common traits everywhere

PR #617, opened 2025-07-02
Issue 616/provide way to enforce an http version

soundofspace:
- `TargetHttpVersion` extension which can be used to manually configure the wanted http version, or which will be configured by other layer that negotiate http versions such as tls alpn
- SwitchRequestVersionExt trait where all request upgrading downgrading logic can be centralized
- Http Inspector `HttpVersionAdapater` which will set req version to `TargetHttpVersion` or default version if configured
- Boring tls will also update tls config builder chain so we also use this target version. With current setup this is not possible for rustls
- Support setting http version in rama http cli and add tests for it


rama-http-backend/src/client/http_inspector/tls_alpn.rs
```diff
@@ -34,7 +35,9 @@ where
         ctx: Context<State>,
         mut req: Request<ReqBody>,
     ) -> Result<Self::Response, Self::Error> {
-        if let Some(proto) = ctx
+        if let Some(version) = ctx.get::<EnforcedHttpVersion>() {
```
GlenDC: this should also set it as `new_version` first, so that we can handle upgrades here correctly.

There is already some minor logic down, but we'll need a lot more complete logic. feel free to add a todo there, but we need to correctly upgrade/downgrade these things here. The full complete upgrade/downgrade logic is out of scope here, but at least you should ensure that the code path is already the same here as the negotiated Http version one. You can open an issue for correctly upgrading/downgrading http versions.

rama-net/src/lib.rs
```diff
@@ -43,4 +43,7 @@ pub mod tls;
 #[cfg(any(feature = "tls", feature = "http"))]
 pub mod fingerprint;
 
+#[cfg(all(feature = "tls", feature = "http"))]
+pub mod tls_http;
```
GlenDC: might be more natural to call this https, but up to you

rama-net/src/tls_http.rs
```diff
@@ -0,0 +1,20 @@
...
+    type Error = OpaqueError;
+
+    fn try_from(value: Version) -> Result<Self, Self::Error> {
+        let version = match value {
+            Version::HTTP_09 => ApplicationProtocol::HTTP_09,
+            Version::HTTP_10 => ApplicationProtocol::HTTP_10,
+            Version::HTTP_11 => ApplicationProtocol::HTTP_11,
+            Version::HTTP_2 => ApplicationProtocol::HTTP_2,
+            Version::HTTP_3 => ApplicationProtocol::HTTP_3,
+            _ => Err(OpaqueError::from_display("not supported"))?,
```
GlenDC: try to make this error a bit more descriptive because if you get much further back in your callstack `not supported` I do not think it will be clear at all

rama-tls-boring/src/client/connector.rs
```diff
@@ -376,13 +377,20 @@ impl<S, K> TlsConnector<S, K> {
         &self,
         ctx: &mut Context<State>,
     ) -> Result<TlsConnectorData, OpaqueError> {
-        let base_builder = self.connector_data.clone();
+        let enforced_alpn = match ctx.get::<EnforcedHttpVersion>() {
```
GlenDC: you can do `map(...).transpose()?` to make it more fluent as it is called

rama-net/src/lib.rs
soundofspace: yea nice idea, so simple yet I didnt think of it

rama-http-backend/src/client/http_inspector/http_version_adapter.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    generate_set_and_with! {
+        pub fn default_version(mut self, version: Option<Version>) -> Self {
+            self.default_version = version;
+            self
+        }
+    }
+}
+
+impl<State, ReqBody> Service<State, Request<ReqBody>> for HttpVersionAdapater
```
GlenDC: This is implemented as an inspector, right? But shouldn't it be a proper layer? Because your response will need to go in reverse, no? At least in a proxy scenario

rama-net/src/https.rs
```diff
@@ -0,0 +1,48 @@
+use rama_core::error::OpaqueError;
+use rama_http_types::Version;
+
+use crate::tls::ApplicationProtocol;
+
+impl TryFrom<Version> for ApplicationProtocol {
```
GlenDC, line 6: why here no reference version (as well), but you do it for the other one?

soundofspace: Version implements Copy so no real need to reference it

GlenDC: True, but afaik there's no automatic defer for `From`, so more a QoL thing. Wouldn't have said if you didn't do it for the other way around, but you did, so you might as well do it here as well

GlenDC (approved):

PR #629, opened 2025-07-11
Support stunnel-like use cases in rama-cli

### LINKED ISSUE #453: support stunnel-like use cases in rama-cli as well as document it
It's another nice showcase / example of what can be done with rama.
Not urgent though.

ikorason:
Initial version of stunnel like feature, help closes #453

### DIFF
--- a/rama-cli/Cargo.toml
+++ b/rama-cli/Cargo.toml
@@ -25,14 +25,15 @@ itertools = { workspace = true }
 mimalloc = { workspace = true, optional = true }
 opentelemetry-otlp = { workspace = true }
 rama = { version = "0.3.0-alpha.2", path = "..", features = [
-    "compression",
-    "boring",
-    "cli",
-    "tcp",
-    "udp",
-    "http-full",
-    "proxy-full",
-    "opentelemetry",
+  "compression",
+  "boring",
+  "cli",
+  "tcp",
+  "udp",
+  "http-full",
+  "proxy-full",
+  "opentelemetry",
+  "rustls",
 ] }
 ratatui = { workspace = true }
 serde = { workspace = true }
--- a/rama-cli/src/cmd/stunnel/mod.rs
+++ b/rama-cli/src/cmd/stunnel/mod.rs
@@ -1,0 +1,144 @@
+use std::{io::BufReader, path::PathBuf, time::Duration};
+
+use clap::{Args, Subcommand};
+use rama::{
+    Layer,
+    error::{BoxError, ErrorContext, OpaqueError},
+    graceful::Shutdown,
+    service::service_fn,
+    tcp::{TcpStream, server::TcpListener},
+    telemetry::tracing,
+    tls::rustls::{
+        dep::{
+            pemfile,
+            pki_types::{CertificateDer, PrivateKeyDer},
+        },
+        server::{TlsAcceptorDataBuilder, TlsAcceptorLayer},
+    },
+};
+use tokio::io::copy_bidirectional;
+use tracing_subscriber::{
+    EnvFilter, filter::LevelFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt,
+};
+
+#[derive(Debug, Args)]
+/// rama stunnel service
+pub struct StunnelCommand {
+    #[command(subcommand)]
+    pub commands: StunnelSubcommand,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum StunnelSubcommand {
+    Server(ServerArgs),
+    Client(ClientArgs),
+}
+
+#[derive(Debug, Args)]
+pub struct ServerArgs {
+    #[arg(long)]
+    pub listen: String,
+    #[arg(long)]
+    pub forward: String,
+    #[arg(long)]
+    pub cert: PathBuf,
+    #[arg(long)]
+    pub key: PathBuf,
+}
+
+#[derive(Debug, Args)]
+pub struct ClientArgs {
+    #[arg(long)]
+    pub listen: String,
+    #[arg(long)]
+    pub connect: String,
+    #[arg(long)]
+    pub cacert: PathBuf,
+}
+
+pub async fn run(cmd: StunnelCommand) -> Result<(), BoxError> {
+    tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
+                .from_env_lossy(),
+        )
+        .init();
+
+    match cmd.commands {
+        StunnelSubcommand::Server(args) => {
+            let shutdown = Shutdown::default();
+
+            let (certs, key) = load_cert_and_key(
+                &std::fs::read(&args.cert).context("failed to read certificate")?,
+                &std::fs::read(&args.key).context("failed to read private key")?,
+            )?;
+
+            let acceptor_data = TlsAcceptorDataBuilder::new(certs, key)?
+                .with_env_key_logger()
+                .expect("with env keylogger")
+                .build();
+
+            shutdown.spawn_task_fn(async move |guard| {
+                tracing::info!("Stunnel server is running...");
+
+                tracing::info!(
+                    "Listening on {} and forwarding to {}",
+                    args.listen,
+                    args.forward
+                );
+
+                let tcp_service = TlsAcceptorLayer::new(acceptor_data).layer(service_fn(
+                    move |mut tls_stream| {
+                        let forward = args.forward.clone();
+
+                        async move {
+                            let mut backend = TcpStream::connect(forward)
+                                .await
+                                .context("failed to connect to backend")?;
+
+                            copy_bidirectional(&mut tls_stream, &mut backend)
+                                .await
+                                .context("failed to relay data")?;
+
+                            Ok::<_, BoxError>(())
+                        }
+                    },
+                ));
+
+                TcpListener::bind(&args.listen)
+                    .await
+                    .expect("Failed to bind stunnel server")
+                    .serve_graceful(guard, tcp_service)
+                    .await;
+            });
+
+            shutdown
+                .shutdown_with_limit(Duration::from_secs(30))
+                .await
+                .expect("graceful shutdown");
+
+            Ok(())
+        }
+
+        StunnelSubcommand::Client(args) => {
+            todo!()
+        }
+    }
+}
+
+pub fn load_cert_and_key(
+    cert_bytes: &[u8],
+    key_bytes: &[u8],
+) -> Result<(Vec<CertificateDer<'static>>, PrivateKeyDer<'static>), OpaqueError> {
+    let cert_chain = pemfile::certs(&mut BufReader::new(cert_bytes))
+        .collect::<Result<Vec<_>, _>>()
+        .context("failed to parse certificate chain")?;
+
+    let private_key = pemfile::private_key(&mut BufReader::new(key_bytes))
+        .context("failed to parse private key")?
+        .context("no private key found")?;
+
+    Ok((cert_chain, private_key))
+}
--- a/rama-cli/src/cmd/mod.rs
+++ b/rama-cli/src/cmd/mod.rs
@@ -6,5 +6,6 @@ pub mod http;
 pub mod ip;
 pub mod proxy;
 pub mod serve;
+pub mod stunnel;
 pub mod tls;
 pub mod ws;
--- a/rama-cli/src/main.rs
+++ b/rama-cli/src/main.rs
@@ -8,7 +8,7 @@ use clap::{Parser, Subcommand};
 use rama::error::BoxError;
 
 pub mod cmd;
-use cmd::{echo, fp, http, ip, proxy, serve, tls, ws};
+use cmd::{echo, fp, http, ip, proxy, serve, stunnel, tls, ws};
 
 pub mod error;
 
@@ -43,6 +43,7 @@ enum CliCommands {
     Ip(ip::CliCommandIp),
     Fp(fp::CliCommandFingerprint),
     Serve(serve::CliCommandServe),
+    Stunnel(stunnel::StunnelCommand),
 }
 
 #[tokio::main]
@@ -59,6 +60,7 @@ async fn main() -> Result<(), BoxError> {
         CliCommands::Ip(cfg) => ip::run(cfg).await,
         CliCommands::Fp(cfg) => fp::run(cfg).await,
         CliCommands::Serve(cfg) => serve::run(cfg).await,
+        CliCommands::Stunnel(cfg) => stunnel::run(cfg).await,
     } {
         Ok(()) => Ok(()),
         Err(err) => {


ikorason:
@GlenDC have some time to start working on this, not sure I'm on the right track, definitely need some guidance. ;)

rama-cli/Cargo.toml
```diff
@@ -25,14 +25,15 @@ itertools = { workspace = true }
...
-    "opentelemetry",
+  "compression",
+  "boring",
+  "cli",
+  "tcp",
+  "udp",
+  "http-full",
+  "proxy-full",
+  "opentelemetry",
+  "rustls",
```
GlenDC: there should be no need for introducing rustls here, you should be able to do it with boring

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+#[derive(Debug, Subcommand)]
+pub enum StunnelSubcommand {
+    Server(ServerArgs),
+    Client(ClientArgs),
+}
+
+#[derive(Debug, Args)]
+pub struct ServerArgs {
+    #[arg(long)]
+    pub listen: String,
```
GlenDC: Do look at other commands in this cli to see what are the conventions for args, e.g. https://github.com/plabayo/rama/blob/c377a87aac46c792dccfb276d115e87a4346a472/rama-cli/src/cmd/serve/mod.rs#L37-L39

* try to use the same arguments for the same exact purpose
* try to avoid strings as much as possible, better to immediately parse it as the right type if possible

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+#[derive(Debug, Args)]
+pub struct ClientArgs {
+    #[arg(long)]
+    pub listen: String,
+    #[arg(long)]
+    pub connect: String,
+    #[arg(long)]
+    pub cacert: PathBuf,
```
GlenDC: Is this always required? Can it not be optional?

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    #[arg(long)]
+    pub key: PathBuf,
+}
+
+#[derive(Debug, Args)]
+pub struct ClientArgs {
+    #[arg(long)]
+    pub listen: String,
+    #[arg(long)]
+    pub connect: String,
```
GlenDC: What is this? try to also put a one line doc for each arg please

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+            shutdown.spawn_task_fn(async move |guard| {
+                tracing::info!("Stunnel server is running...");
+
+                tracing::info!(
+                    "Listening on {} and forwarding to {}",
+                    args.listen,
+                    args.forward
+                );
+
+                let tcp_service = TlsAcceptorLayer::new(acceptor_data).layer(service_fn(
```
GlenDC: You can replace the entire `service_Fn` with https://ramaproxy.org/docs/rama/tcp/client/service/struct.Forwarder.html. Pretty certain that does exactly what you want without having to reinvent the wheel here

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+                                .context("failed to relay data")?;
+
+                            Ok::<_, BoxError>(())
+                        }
+                    },
+                ));
+
+                TcpListener::bind(&args.listen)
+                    .await
+                    .expect("Failed to bind stunnel server")
```
GlenDC: Please do your fallible operations outside of your spawn tasks. This way you can just return it as an error.

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,144 @@
+use std::{io::BufReader, path::PathBuf, time::Duration};
+
+use clap::{Args, Subcommand};
+use rama::{
```
GlenDC, line 22: I'm not very strict on this, but I do like at least a minimum of hygiene in these imports. You can check the other cli files on how I usually sort these imports

GlenDC (changes requested):
@ikorason you did very well. This is going the right direction for sure. I can see you are not yet finished, but what you have is besides some imperfections certainly more or less like expected.

Is there anything specific you are stuck with, or where you need guidance?

ikorason:
> @ikorason you did very well. This is going the right direction for sure. I can see you are not yet finished, but what you have is besides some imperfections certainly more or less like expected.
> 
> Is there anything specific you are stuck with, or where you need guidance?

thanks for the quick feedbacks here, I guessed I was just looking these feedbacks early to get a better understanding of the direction. I will continue working on it and request a review early. Thanks

GlenDC:
Happy to see your continued progress @ikorason , hope you are having fun with this one. Is there anything you wish to talk through, discuss or ask? Anything else I can be of help with? Take care and have a nice weekend.

ikorason:
@GlenDC hey sorry for getting back late, I'm still figuring out how to do the client part, but other than that I think I almost finish the server part. It's just my personal time not allowing me to finish this sooner, sorry.

GlenDC:
> @GlenDC hey sorry for getting back late, I'm still figuring out how to do the client part, but other than that I think I almost finish the server part. It's just my personal time not allowing me to finish this sooner, sorry.

All good @ikorason , it is supposed to be for fun. Anything in your life has priority. Just wanted to check in. And if you ever need or want to pass on this project I would also equally understand. Either way you'll have my support.

ikorason:
@GlenDC I think I'm stuck at doing the client side and need some guidance ;) I think I supposed to use TlsConnectorLayer for this?

GlenDC:
> @GlenDC I think I'm stuck at doing the client side and need some guidance ;) I think I supposed to use TlsConnectorLayer for this?

Yes, in this example you find an example of that:
https://github.com/plabayo/rama/blob/ec19794d5fe3cbdaab6260ebe6fe33c93173ed4d/examples/mtls_tunnel_and_service.rs#L163-L165

STUNNEL is basically a `tcp -> tcp -> tls ---------> tls -> tcp -> tcp` forwarding chain, as such a kind of tunnel for secure purpose, or stunnel.

- your client will receive incoming tcp connections and for each established tcp stream:
   - establish a tcp connection with the stunnel server and do the tls handshake
- your server will receive thse connections from your stunnel client and for each of these
   - establish a tcp connection with the server to forward to (hardcoded)
   - and if that is there do tls handshake with the stunnel client and forward the bytes at that point

That's about it, for all of those steps and parts you already find something available in rama to achieve it. In later PRs / versions we can expand this stunnel support further, but this is already a nice enoguh start.

ikorason:
@GlenDC Hey just wanted to let you know I am able to get back and start working on this again, its been fun for the last couple of days to learn this, but I am having trouble with my client implementation, when I try to test my current implementation the connection is being reset somewhere in the chain. Do you have any ideas? Appreciate it.

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+#[derive(Debug, Args)]
+pub struct ClientArgs {
+    // Address and port to listen on
+    #[arg(long, default_value = "127.0.0.1:8003")]
+    pub listen: Interface,
+    // Address to connect to
+    #[arg(long, default_value = "127.0.0.1:8002")]
+    pub connect: Interface,
```
GlenDC: This should probably be a SocketAddress instead of an interface

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,221 @@
+//! ## Server Mode
+//!
+//! Accepts TLS connections and forwards plain TCP to a backend:
+//!
+//! ## Quick Test
+//!
+//! # 1. Generate a self-signed certificate
+//! openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes -subj "/CN=localhost"
```
GlenDC: Let's avoid needing openssl. Look at the other cmds, you can probably use the same tls cert logic as the other CLI commands use from rama-cli

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+//!
+//! # 2. Start backend
+//! nc -l localhost 8080
```
GlenDC: Why not just use the `echo` service from rama-cli, e.g. `rama echo` this way you'll know it works because you get an http request echo (instead of an https request echo); and we do not rely on excternal tooling

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+                ocsp: None,
+            }));
+
+            let acceptor_data = TlsAcceptorData::try_from(server_config)?;
+
+            let listener = TcpListener::bind(args.listen.clone())
+                .await
+                .expect("Failed to bind stunnel server");
+
+            let forward_authority = match &args.forward {
```
GlenDC: once forward is a `SocketAddress` you won't need this match logic

GlenDC:
> @GlenDC Hey just wanted to let you know I am able to get back and start working on this again, its been fun for the last couple of days to learn this, but I am having trouble with my client implementation, when I try to test my current implementation the connection is being reset somewhere in the chain. Do you have any ideas? Appreciate it.

Left some more comments. Can you please also resolve threads when they are resolved (previous comments).

Regarding your question. Not sure, your code setup looks fine just reading through it. What do the trace logs tell you? Why is it exiting? Should normally be visible in the logs.

ikorason:
@GlenDC so everything seems to be working now, the problem I mentioned earlier was because it was certificate mismatch, also I added --insecure flag while I was testing and left it, do you think its ok?

GlenDC:
> @GlenDC so everything seems to be working now, the problem I mentioned earlier was because it was certificate mismatch, also I added --insecure flag while I was testing and left it, do you think its ok?

Cool, glad to hear.
And yes, having that insecure flag is totally fine, it's opt-in so no issue.

ikorason:
@GlenDC also I was thinking about adding domain name support for the client mode instead of just a socket address, changing `connect` option to use String and parse it, what do you think?

GlenDC:
> @GlenDC also I was thinking about adding domain name support for the client mode instead of just a socket address, changing `connect` option to use String and parse it, what do you think?

In that case you should be able to use `Authority` as even for domains we would require here a port as there is no default port for these cases, so best to be explicit about it

ikorason:
> > @GlenDC also I was thinking about adding domain name support for the client mode instead of just a socket address, changing `connect` option to use String and parse it, what do you think?
> 
> 
> 
> In that case you should be able to use `Authority` as even for domains we would require here a port as there is no default port for these cases, so best to be explicit about it

So if use Authority user need to explicitly provide port but what about if I use String? Then I can provide default port in that case with some string manipulation, Or should I stick to something simpler like use Authority, can't decide 🙂 still learning sorry

GlenDC:
> > > @GlenDC also I was thinking about adding domain name support for the client mode instead of just a socket address, changing `connect` option to use String and parse it, what do you think?
> > 
> > 
> > In that case you should be able to use `Authority` as even for domains we would require here a port as there is no default port for these cases, so best to be explicit about it
> 
> So if use Authority user need to explicitly provide port but what about if I use String? Then I can provide default port in that case with some string manipulation, Or should I stick to something simpler like use Authority, can't decide 🙂 still learning sorry

Yeah just use `Authority` as is. No need to do string first. This is about stunnel, don't think there is a good default port here. We can always revisit this later, but for now always require that the user defines port (when they define the authority themselves)

ikorason:
@GlenDC please check again, and let me know what else we can do.

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
+//! Stunnel for Rama
+//!
+//! # With auto-generated self-signed certificate (development & testing only)
+//! 1. rama echo echo --bind 127.0.0.1:8001 --mode http
```
GlenDC:
```suggestion
//! 1. rama echo echo
```

- `http` is default mode
- default bind interface is `127.0.0.1:8080` which works fine for you, and it's sufficiently different from 8001 and 8002 that it probably is better than being close to your stunnel client/server ports

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+//! 2. rama stunnel server --listen 127.0.0.1:8002 --forward 127.0.0.1:8001
+//! 3. rama stunnel client --listen 127.0.0.1:8003 --connect 127.0.0.1:8002 --insecure
+//! # Test
+//! 4. curl http://127.0.0.1:8003
```
GlenDC:
```suggestion
//! 4. rama http 127.0.0.1:8003
```
rama has an http client, and `http://` is the default assumed scheme

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+//!
+//! # Explicitily provided certificates
```
GlenDC, line 10: regarding cmds, same feedback as in your other instructions

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
+//! Stunnel for Rama
```
GlenDC, line 1: please also add cli e2e tests under https://github.com/plabayo/rama/tree/main/tests/integration/cli/cli_tests; see the other tests there for inspiration how. These are e2e tests to test the entire http client -> stunnel client -> stunnel server -> http server flow

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+#[derive(Debug, Args)]
+/// rama stunnel service
+pub struct StunnelCommand {
+    #[command(subcommand)]
+    pub commands: StunnelSubcommand,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum StunnelSubcommand {
+    Server(ServerArgs),
```
GlenDC, line 68: bit of (CLI doc) comments for these server and client commands, such that --help can show useful info about them

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    ///
+    /// use this to trust a specific CA or self-signed certificate.
+    /// if not provided, the system trust store will be used.
+    pub cacert: Option<PathBuf>,
+
+    #[arg(short = 'k', long)]
+    /// skip TLS certificate verification (INSECURE - testing only!)
+    ///
+    /// this disables all certificate validation and should NEVER be used
+    /// in production. Use --cacert instead for self-signed certificates.
```
GlenDC: or ensure that your system keychain has the cert or one of its root certs as a trusted cert.

(^ add something like that to it)

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+}
+
+pub async fn run(cmd: StunnelCommand) -> Result<(), BoxError> {
+    crate::trace::init_tracing(LevelFilter::INFO);
+
+    match cmd.commands {
+        StunnelSubcommand::Server(args) => {
+            let shutdown = Shutdown::default();
+
+            let server_config = if let (Some(cert_path), Some(key_path)) = (&args.cert, &args.key) {
```
GlenDC: I think you might want a match statement, like

- Some + Some => this
- None + None => the option that is your current else case
- (Some + None) | (None + Some) => maybe best to give error as there's little reason to do that?

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        StunnelSubcommand::Client(args) => {
+            let shutdown = Shutdown::default();
+
+            let mut tls_builder = TlsConnectorDataBuilder::new();
+
+            // Configure certificate verification
+            if args.insecure {
+                tls_builder = tls_builder.with_server_verify_mode(ServerVerifyMode::Disable);
```
GlenDC: Doesn't this have something like `set_server_verify_mode`??? so that you do not need to re-assign it

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                let ca_cert = X509::from_pem(ca_pem.as_bytes())
+                    .map_err(|e| format!("Failed to parse CA certificate: {e}"))?;
+
+                let mut store_builder = X509StoreBuilder::new()
+                    .map_err(|e| format!("Failed to create certificate store: {e}"))?;
+                store_builder
+                    .add_cert(ca_cert)
+                    .map_err(|e| format!("Failed to add CA certificate to store: {e}"))?;
+
+                tls_builder =
```
GlenDC: Doesn't this have something like `set_server_verify_cert_store`??? so that you do not need to re-assign it

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                tracing::info!("CA certificate loaded and added to trust store");
+            } else {
+                tracing::info!("Using system trust store for certificate verification");
+            }
+
+            let tls_client_data_builder = Arc::new(tls_builder);
+
+            let listener = TcpListener::bind(args.listen.clone())
+                .await
+                .expect("Failed to bind stunnel client");
```
GlenDC:
```suggestion
                .context("bind stunnel client")?;
```

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+            } else {
+                tracing::info!("Using auto-generated self-signed certificate");
+                new_server_config(None)
+            };
+
+            let acceptor_data = TlsAcceptorData::try_from(server_config)?;
+
+            let listener = TcpListener::bind(args.listen.clone())
+                .await
+                .expect("Failed to bind stunnel server");
```
GlenDC:
```suggestion
                .context("bind stunnel server")?;
```

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                    TlsConnectorLayer::secure()
+                        .with_connector_data(tls_client_data_builder)
+                        .into_layer(TcpConnector::new()),
+                );
+
+                listener.serve_graceful(guard, tcp_service).await;
+            });
+
+            shutdown
+                .shutdown_with_limit(Duration::from_secs(30))
```
GlenDC: make the 30 seconds configurable please (see other commands)

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+                .await
+                .expect("graceful shutdown");
```
GlenDC:
```suggestion
                .context("graceful shutdown")?;
```

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,295 @@
+//! Stunnel for Rama
+//!
+//! # With auto-generated self-signed certificate (development & testing only)
+//! 1. rama echo echo
+//! 2. rama stunnel server --listen 127.0.0.1:8002 --forward 127.0.0.1:8080
+//! 3. rama stunnel client --listen 127.0.0.1:8003 --connect (127.0.0.1 | localhost):8002 --insecure
+//! # Test
+//! 4. rama http (127.0.0.1 | localhost):8003
```
GlenDC:
```suggestion
//! 4. rama http 127.0.0.1:8003
```

Just do this, should work for all of the relevant platforms

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+//! 1. rama echo
+//! 2. rama stunnel server --listen 127.0.0.1:8002 \
+//!    --forward 127.0.0.1:8080 \
+//!    --cert server-cert.pem \
+//!    --key server-key.pem
+//! 3. rama stunnel client --listen 127.0.0.1:8003 \
+//!    --connect (127.0.0.1 | localhost):8002 \
+//!    --cacert cacert.pem
+//! # Test
+//! 4. rama http (127.0.0.1 | localhost):8003
```
GlenDC:
```suggestion
//! 4. rama http 127.0.0.1:8003
```

rama-cli/src/cmd/stunnel/mod.rs
```diff
@@ -0,0 +1,295 @@
...
+
+    graceful.spawn_task_fn(async move |guard| {
+        tracing::info!("Stunnel server is running...");
+        tracing::info!(
+            "Listening on {} and forwarding to {}",
+            listen_addr,
+            forward_addr
+        );
+
+        let tcp_service = TlsAcceptorLayer::new(acceptor_data).layer(Forwarder::new(forward_addr));
```
GlenDC:
```suggestion
        let tcp_service = TlsAcceptorLayer::new(acceptor_data).into_layer(Forwarder::new(forward_addr));
```

avoids useless clone. Not that it matters for this cold path, but still good practice none the less.

GlenDC (commented):
Looks good. A couple of last comments left and than we can merge!

ikorason:
thank you for all the feedbacks, just updated your last bit of comments

PR #634, opened 2025-07-14
Fix Issue #509

AydanPirani:
In-progress.


AydanPirani:
Hey! @GlenDC  @soundofspace  Got a very rudimentary implementation going, can I get a quick informal review on this before finishing implementation?

examples/http_curl_command.rs
```diff
@@ -0,0 +1,22 @@
+//! This example demonstrates how to use the http curl command
```
GlenDC, line 1: this file is to be deleted as discussed in the parent isuse

rama-http-types/src/curl/mod.rs
```diff
@@ -0,0 +1,45 @@
+use http::{HeaderMap, HeaderValue, Method, Request, Uri, Version};
```
GlenDC:
```suggestion
use crate::{HeaderMap, HeaderValue, Method, Request, Uri, Version};
```

rama-http-types/src/curl/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+fn generate_curl_command(method: &Method, uri: &Uri, version: Version, headers: &HeaderMap<HeaderValue>) -> Result<String, OpaqueError>{
+    let mut cmd = format!("curl -X {} '{}'", method, uri);
+
+    match version {
+        Version::HTTP_09 => cmd.push_str(" --http0.9"),
+        Version::HTTP_10 => cmd.push_str(" --http1.0"),
+        Version::HTTP_11 => cmd.push_str(" --http1.1"),
+        Version::HTTP_2 => cmd.push_str(" --http2"),
+        Version::HTTP_3 => cmd.push_str(" --http3"),
+        _ => panic!("Unexpected HTTP version!"),
```
GlenDC: no panics, just return errors, can even just be an `OpaqueError`

rama-http-types/src/curl/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    for (name, value) in headers.iter() {
+        let val = match value.to_str() {
+            Ok(s) => s,
+            Err(_) => continue, // skip invalid UTF-8 headers
+        };
+        cmd.push_str(&format!(r#" -H "{}: {}""#, name.as_str(), val));
```
GlenDC: single quotes are probably better as the other ones allows some chars to be expanded

GlenDC: as in `" -H '{}: {}'"`

rama-http-types/src/curl/mod.rs
```diff
@@ -0,0 +1,45 @@
+use http::{HeaderMap, HeaderValue, Method, Request, Uri, Version};
+use rama_error::OpaqueError;
+
+// TODO: make sealed(private) trait, and add sealed(private) impl for Request<T>
+pub trait IntoCurlHeadersCommand {
+    fn to_curl_command(&self) -> Result<String, OpaqueError>;
```
GlenDC: I'm fine in moving this to a sealed trait, but in that case you also want to move this method to that trait, such that the public trait becomes fully opaque

rama-http-types/src/curl/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+        let method = self.method();
+        let uri = self.uri();
+        let version = self.version();
+        let headers = self.headers();
+
+        generate_curl_command(method, uri, version, headers)
+    }
+}
+
+pub fn request_headers_to_curl_command(request: impl IntoCurlHeadersCommand) -> Result<String, OpaqueError> {
```
soundofspace: `IntoCurlHeadersCommand` can be dropped, and you can use `HttpRequestParts` instead, this prevents the need to create a new trait for something that already exists.

It becomes something like this then
```
pub fn request_headers_to_curl_command(req_parts: &impl HttpRequestParts) -> Result<String, OpaqueError> {
  move generate_curl_command logic to this body
}
```

GlenDC: @soundofspace not really.... Because it misses the body. Meaning a POST request (For example) with a json body will not be whole and complete in the `curl` command.

soundofspace: Im talking specifically about this function `request_headers_to_curl_command` which focusses on request headers (and this trait `IntoCurlHeadersCommand`). But yea if you need a trait to abstract body away then a new one will be needed indeed

soundofspace: But since that trait specifically would only be implemented for Request<body> I would argue that hiding that behind a trait is unneeded complexity and can be added later if we want to support that for other things also @GlenDC

GlenDC: yeah agree, I would for now keep it simple and do something like:

```
pub fn curl_command_from_request<B: Into<Body>>(req: Request<B>) -> Result<String, OpaqueError>;
```

Based on usage and feature requests we can always expand and iterate as we go.

GlenDC: This still allows use cases where you come from `Part` as you can just make a `Request` with a `Body::empty` from that. But that would be for now a user-action prior to calling this function, out of the scope of this PR.

examples/http_curl_command.rs
AydanPirani: Yup, using this as a temporary driver. Will delete before finishing up.

AydanPirani:
@GlenDC  @soundofspace Hey hey! Running into a small issue here: how do I get the body data to convert to Bytes? Can't seem to figure this one out, any pointers would be much appreciated :)

GlenDC:
> @GlenDC @soundofspace Hey hey! Running into a small issue here: how do I get the body data to convert to Bytes? Can't seem to figure this one out, any pointers would be much appreciated :)

👇  https://github.com/plabayo/rama/blob/d332fd1affa0bdaae1b4ca0f7e49f05ba9cc4942/src/cli/service/echo.rs#L460

AydanPirani:
@GlenDC  Got it, thanks. Also, do you have any to check if a Proxy is enabled?

GlenDC:
> @GlenDC Got it, thanks. Also, do you have any to check if a Proxy is enabled?

Probably gonna change a tiny bit in near future but for now you can check if `ProxyAddress` (https://ramaproxy.org/docs/rama/net/address/struct.ProxyAddress.html) is present in the `Context`: `if let Some(addr) = ctx.get::<ProxyAddress>() { '-x {addr}' }`

something like that in hybrid pseudo code

AydanPirani:
Got it, thanks. And one last thing - is a string a valid body type? I presume not, in that case is [this](https://github.com/plabayo/rama/blob/d332fd1affa0bdaae1b4ca0f7e49f05ba9cc4942/rama-net/src/stream/service/echo.rs#L65-L77) the right approach to testing with a stream?

GlenDC:
> Got it, thanks. And one last thing - is a string a valid body type? I presume not, in that case is [this](https://github.com/plabayo/rama/blob/d332fd1affa0bdaae1b4ca0f7e49f05ba9cc4942/rama-net/src/stream/service/echo.rs#L65-L77) the right approach to testing with a stream?

You are confusing different kinds of echo services. You are linking to a stream echo service. That lives on the transport layer. That's different from the echo service which I was linking you that works with http on the application layer.

The first lives on the transport layer (e.g. tcp) and just echo's back the incoming bytes as-is. You can as a human easily play with it using tooling like `telnet`. There's a tcp example about it that you can try out.

But this is nothing to do with your work here. Different kind of echo service.

AydanPirani:
@GlenDC  Can you link the path to the EchoService? The one within examples/tcp_listener_layers is under `rama-net/src/stream/service/echo.rs`, but seems like the wrong one...

GlenDC:
> @GlenDC  Can you link the path to the EchoService? The one within examples/tcp_listener_layers is under `rama-net/src/stream/service/echo.rs`, but seems like the wrong one...

https://github.com/plabayo/rama/blob/21f68c20fec345c79692fb41ade1f5211d7206ce/src/cli/service/echo.rs

GlenDC:
All good @AydanPirani ? Any questions or blockers?

GlenDC:
@AydanPirani is this something you are planning to finish or did life get in the way? Either way fine. Also no pressure but a status update would be appreciated. In worst case we can take over, no hard feelings.

PR #636, opened 2025-07-17
Connection pool: add support for round robin reuse

maartendeprez:
This is done by selecting the last matching connection from the pool instead of the first one.

Also, clean up connections over idle timeout first, before searching for a matching connection. This is important for the round robin case, since otherwise we might select a timed-out connection even though there are newer connections available. This requires an extra scan (binary search) over the stored connections list, but also ensures that timed out connections are cleaned up as early as possible.

### DIFF
--- a/rama-net/src/client/pool.rs
+++ b/rama-net/src/client/pool.rs
@@ -160,14 +160,7 @@ impl ReuseStrategy {
     ) -> Option<(usize, PooledConnection<C, ID>)> {
         let idx = match self {
             Self::FiFo => storage.iter().position(|stored| &stored.id == id)?,
-            Self::RoundRobin => {
-                storage
-                    .iter()
-                    .enumerate()
-                    .rev()
-                    .find(|(_, stored)| &stored.id == id)?
-                    .0
-            }
+            Self::RoundRobin => storage.iter().rposition(|stored| &stored.id == id)?,
         };
 
         let pooled_conn = storage.remove(idx)?;


rama-net/src/client/pool.rs
```diff
@@ -131,13 +132,59 @@ struct PooledConnection<C, ID> {
...
+pub struct LruDropPool<R, C, ID> {
     storage: Arc<Mutex<VecDeque<PooledConnection<C, ID>>>>,
     total_slots: Arc<Semaphore>,
     active_slots: Arc<Semaphore>,
     idle_timeout: Option<Duration>,
     returner: ConnReturner<C, ID>,
+    _marker: PhantomData<R>,
+}
+
+pub struct FiFoReuse;
```
GlenDC: Implement `Clone`, `Debug` and `Default` for these please.
Also mark them non Exhaustive. You can Add also a `pub fn new() -> Self` for it.

This way we can still add state to it in future if we ever need to do so.

rama-net/src/client/pool.rs
```diff
@@ -131,13 +132,59 @@ struct PooledConnection<C, ID> {
...
+pub struct RoundRobinReuse;
+
+trait Reuse<C, ID> {
+    fn get_conn(
+        storage: &mut VecDeque<PooledConnection<C, ID>>,
+        id: &ID,
+    ) -> Option<(usize, PooledConnection<C, ID>)>;
+}
+
+impl<C, ID> Reuse<C, ID> for FiFoReuse
```
GlenDC: Am I going crazy? Do you not just do exactly the same in the `Reuse` impl of `FiFoReuse` and `RoundRobinReuse`?

```
let idx = storage.iter().position(|stored| &stored.id == id)?;
```

```
let idx = storage
            .iter()
            .enumerate()
            .find(|(_, stored)| &stored.id == id)?
            .0;
 ```
 
 These are equivalent, no??

File: rama-net/src/client/pool.rs [hunk 3]
```diff
@@ -131,13 +132,59 @@ struct PooledConnection<C, ID> {
...
     active_slots: Arc<Semaphore>,
     idle_timeout: Option<Duration>,
     returner: ConnReturner<C, ID>,
+    _marker: PhantomData<R>,
+}
+
+pub struct FiFoReuse;
+pub struct RoundRobinReuse;
+
+trait Reuse<C, ID> {
```
GlenDC: Given this is a private `trait`, I do not see a point in having this `R` become a new generic in the pool. You can anyway not implement it yourself like this. And if so it means your options are fixed. Which means you can just work with an enum to represent your choices... But then again... afaik atm your two things are exactly the same o.o

GlenDC (changes requested):
I have some questions about your implementation here. I think @soundofspace is back next week, please discuss with him and see how to proceed with it.

rama-net/src/client/pool.rs
maartendeprez: Can be an enum if you prefer, to keep types simple. Main reason to make the trait public in the current implementation would be to allow genericity over reuse policy, but with an enum you get that for free. Public trait might be useful if someone wants to implement a clever reuse policy for some specific use-case.

If we make the `Reuse` trait public, we also need to make `PooledConnection` public, but I don't think that would be a problem.

soundofspace: I would also switch to an enum to keep types simple and step away from the Reuse trait. Goal of this default pool implementation was always to support the most common use cases without going too crazy (if people need clever stuff they can always implement a custom Pool)

Introducing a nested trait for a specific pool implementation feels like we are trying to support to much, but I'm open for it if there are other use cases? (but feels like just extending the enum in those cases is also the way to go)

Having an enum like ReuseStrategy { FiFo (default), RoundRobin, MaybeMore } makes it also very easy to support these strategies in `HttpPooledConnectorConfig` and `EasyHttpWebClientBuilder`, which is nice bonus. Without that enum its also possible but would introduce new generics and more complex types everywhere

rama-net/src/client/pool.rs
```diff
@@ -131,13 +132,77 @@ struct PooledConnection<C, ID> {
     last_used: Instant,
 }
 
+pub type FiFoReuseLruDropPool<C, ID> = LruDropPool<FiFoReuse, C, ID>;
+pub type RoundRobinReuseLruDropPool<C, ID> = LruDropPool<RoundRobinReuse, C, ID>;
+
 /// Connection pool that uses FiFo for reuse and LRU to evict connections
```
soundofspace: I see still quite some comments/traces mentioning FiFo can you do a sweep of these and update them

rama-net/src/client/pool.rs
maartendeprez: Hm, question about the enum solution... Can we rename the pool type? Do we leave a FiFoReuse alias for backward compatibility?

GlenDC:
Don't merge this @soundofspace , want to give this another review tonight

rama-net/src/client/pool.rs
```diff
@@ -131,13 +131,48 @@ struct PooledConnection<C, ID> {
...
+
+impl ReuseStrategy {
+    fn get_conn<C, ID: PartialEq>(
+        &self,
+        storage: &mut VecDeque<PooledConnection<C, ID>>,
+        id: &ID,
+    ) -> Option<(usize, PooledConnection<C, ID>)> {
+        let idx = match self {
+            Self::FiFo => storage.iter().position(|stored| &stored.id == id)?,
+            Self::RoundRobin => {
```
GlenDC: Maybe I'm missing something @maartendeprez and @soundofspace , but how is this RoundRobin? This just seems like `LiFo` to me?

rama-net/src/client/pool.rs
```diff
@@ -131,13 +131,48 @@ struct PooledConnection<C, ID> {
...
+                storage
+                    .iter()
+                    .enumerate()
```
GlenDC: I'm also confused why in 1 case you use `position` and another one `enumerate + find`? they are the same ... And you could still combine that with `rev`...? Even though not sure rev is what you want here.

rama-net/src/client/pool.rs
```diff
@@ -231,41 +275,33 @@ where
...
-                    timeout_triggered = true;
-                }
-            }
-
-            // Since we always insert in the beginning we can assume if this connection has been idle for too long
-            // all connections older then this one should also be evicted
-            if timeout_triggered {
+        if let Some(timeout) = self.idle_timeout {
+            let now = Instant::now();
+            let idx = storage.partition_point(|conn| now.duration_since(conn.last_used) <= timeout);
```
GlenDC, line 285: this could use 1 or 2 sentences of high level explanation of the desired intent here (code docs). Code that is highly technical like this are the typical places that initial implementations or refactors make mistakes in. Knowing the intent is an easy way to verify what is desired here.

rama-net/src/client/pool.rs
```diff
@@ -1010,4 +1044,48 @@ mod tests {
...
+        }
+
+        drop(conns);
+
+        // We should now have two connections with the same key in the pool,
+        // from most to least recently used. ([conn2, conn1]). Requesting
+        // another connection should return the first or last one depending on
+        // the reuse policy.
+        let conn = svc
+            .connect(Context::default(), String::from(""))
```
GlenDC: not that I care too much but why do you use a string type here instead of the much simpler to use `()`?

rama-net/src/client/pool.rs
```diff
@@ -472,15 +506,15 @@ where
         if result.is_err() {
             let id = &self.pooled_conn.as_ref().expect("msg").id;
             trace!(
-                "fifo connection pool: detected error result, marking connection w/ id {id:?} as failed"
+                "lru connection pool: detected error result, marking connection w/ id {id:?} as failed"
```
GlenDC: wasn't the case before either but I prefer to use capitals if it's not a word, so `LRU`. "shortened versions" of words like `id` are fine to lowercase because that can be seen as a word.

rama-net/src/client/pool.rs
```diff
@@ -231,41 +275,33 @@ where
...
+                active_slot,
+                pooled_conn: Some(pooled_conn),
+                returner: self.returner.clone(),
+                failed: false.into(),
+            }));
+        }
+
         let pool_slot = match self.total_slots.clone().try_acquire_owned() {
             Ok(permit) => PoolSlot(permit),
             Err(_) => {
```
GlenDC: You might as well include this error in your trace `: {err}` so it's not just silenced into the void

rama-net/src/client/pool.rs
```diff
@@ -231,41 +275,33 @@ where
...
             }
         }
 
+        if let Some((idx, pooled_conn)) = self.reuse_strategy.get_conn(&mut storage, id) {
+            trace!("lru connection pool: connection #{idx} found for given id {id:?}");
+            return Ok(ConnectionResult::Connection(LeasedConnection {
+                active_slot,
+                pooled_conn: Some(pooled_conn),
+                returner: self.returner.clone(),
+                failed: false.into(),
```
GlenDC: I prefer you use notations like `AtomicBool::new` here so it's clear from reading this one line what is getting created here (without having to have an IDE to do so)

rama-net/src/client/pool.rs
maartendeprez: It uses the least recently used connection wit matching id, which is returned to the front of the queue after use. That makes the pool rotate through all available connections for sequential requests, in contrast to using only the first connection for FiFo.

GlenDC: Oh ok that sounds about right then. My bad in my head we were adding to the back, didn't check that part.

rama-net/src/client/pool.rs
maartendeprez: I tested that. It returns the position from the end of the list.

```rs
fn main() {
    let conns = vec!["A", "B", "A", "C", "D", "E", "F"];
    println!("{:?}", conns.iter().rev().position(|c| *c == "A"));
    println!("{:?}", conns.iter().enumerate().rev().find_map(|(i,c)| (*c == "A").then_some(i)));
}
```

Outputs:
```
Some(4)
Some(2)
```

Of course we can use (length - idx), up to you.

GlenDC: Yes you would indeed need to compute the index based on the length. But that's what you want.
Because the only way the other one works is because it has to first store all your enumerations in a collection, such that it can than reverse it. I might be wrong but that feels like a useless allocations that can be avoided. Or perhaps it is actually somehow efficiently implemented, do tell me, haven't checked what it will boil down to.

soundofspace: Enumerate already know the exact length of the vec so it actually does this shortcut for us, so this approach is very effeciently and I do like it more also vs length-idx

soundofspace: nvm `rposition()` exists for this exact use case (just came across it somewhere else). Fyi length - idx will also be wrong by 1, so that would have to be length - idx - 1 (perfect example why functional is often safer).

rama-net/src/client/pool.rs
maartendeprez: Added a comment

rama-net/src/client/pool.rs
maartendeprez: Indeed, the off-by-one was what i wanted to avoid, but `rposition` is better

rama-net/src/client/pool.rs
maartendeprez: Uppercased in all log messages

rama-net/src/client/pool.rs
maartendeprez: Uppercased in all log messages

rama-net/src/client/pool.rs
maartendeprez: Simplified it using `()`.

rama-net/src/client/pool.rs
GlenDC: Agreed

rama-net/src/client/pool.rs
```diff
@@ -472,15 +512,15 @@ where
...
             self.mark_as_failed();
         }
         result
     }
 }
 
-impl<C, ID: Debug> Debug for FiFoReuseLruDropPool<C, ID> {
+impl<C, ID: Debug> Debug for LruDropPool<C, ID> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_list()
```
GlenDC, line 525: @soundofspace as this is your code originally. This misses debug info about properties such as timeout and now also reuse strategy

GlenDC:
Looks good to me. @soundofspace  can you review one last time? If approved by you as well it can be merged in.

rama-net/src/client/pool.rs
soundofspace: Will fix that later

soundofspace (approved):
Looks great!

PR #637, opened 2025-07-19
Tcp Connector Pool

Babbleshack:
- Add `TcpStreamConnectorPool`
  -  Add Round Robin pool
  - Add Random pool
  
 - closes: https://github.com/plabayo/rama/issues/595#event-18614331880


Babbleshack:
After some thought I have realised that my random implementation is not great. This is because I shuffle the collection of connectors on initialisation I then proceed to round-robin shuffled collection.

Therefore it isn't much different to round robin! 😅

Instead I guess I'll need to create a new rng each time.

rama-tcp/src/client/mod.rs
```diff
@@ -4,8 +4,9 @@
 pub mod service;
 
 mod connect;
+pub mod pool;
```
GlenDC: do not squeeze it between here, there is plenty of space in file. Give it its own nice place ;)

rama-tcp/src/client/mod.rs
```diff
@@ -4,8 +4,9 @@
...
 #[doc(inline)]
-pub use connect::{TcpStreamConnector, default_tcp_connect, tcp_connect};
+pub use connect::{default_tcp_connect, tcp_connect, TcpStreamConnector};
```
GlenDC: Are you still on an old Rust version or so? Try to get latest as something tells me you are on an older version, as recently this import order changed?

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,150 @@
...
+        Self {
+            idx: Arc::new(AtomicUsize::default()),
+            connectors: Self::shuffle_connectors(connectors, rng),
+        }
+    }
+
+    /// New 'Round Robin' `TcpStreamConnector`
+    pub fn new_round_robin(connectors: Vec<C>) -> Self {
+        Self {
+            idx: Arc::new(AtomicUsize::default()),
```
GlenDC: I would randomize the index at creation time so you don't always start with the same connector, WDYT?

rama-tcp/src/client/pool.rs
```diff
@@ -0,0 +1,150 @@
...
+}
+
+impl<C: TcpStreamConnector> TcpStreamConnectorPool<C> {
+    /// A `TcpStreamConnector` where each connection is chosed randomly from a pool of
+    /// `TcpStreamConnector`s
+    pub fn new_random(connectors: Vec<C>) -> Self {
+        Self::new_random_with_rng(connectors, rand::rng())
+    }
+
+    fn shuffle_connectors<Rng: rand::Rng>(mut connectors: Vec<C>, mut rng: Rng) -> Vec<C> {
```
GlenDC: why is this a method? or even a thing at all, you are basically just calling shuffle

GlenDC (changes requested):
Besides the semi-minor comments you are indeed also correct in saying that your "random" one is not really random.

You can probably work with an internal `Selector` enum or however you call it, as in:

```rust
enum Selector {
     RoundRobin(Arc<AtomicUsize>),
     Random(ThreadRng),
}
```

Something like that. Not sure if I would allow a custom Rng to be passed in. I would just leave that option out so we can just store a fixed type for now. If there's ever a good enough reason to allow custom Rngs we can still see what to do with this.

rama-tcp/src/client/pool.rs
Babbleshack: Fair point. There was an earlier version of the code where the unit test would call this function. I'll remove it.

rama-tcp/src/client/pool.rs
Babbleshack: I could do that. Am not sure what the expected behaviour the user might want. 

A part of me feels the user may want to preserve the order, but does that make sense?

Babbleshack: The reason I provide this method is so I can build unit tests -- however in this case I think there are two tests:

1. Does the impl provide a random selection from a set of connectors
2. Does the impl return more connectors than the original set

Babbleshack: In the end I left it for the user to decide. If they want a randomised index they can pass a shuffled `Vec`

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+        Self::RoundRobin(Arc::new(AtomicUsize::default()))
+    }
+
+    fn next<C: Clone>(&self, connectors: &[C]) -> Option<C> {
+        if connectors.is_empty() {
+            return None;
+        }
+        let selection = match self {
+            Selector::RoundRobin(atomic_usize) => atomic_usize.fetch_add(1, Ordering::Relaxed),
+            Selector::Random => rand::rng().next_u64() as usize,
```
Babbleshack, line 33: Maybe this should be `next_u32` for 32bit systems?

Babbleshack: I could do something like this:
```rust
            Selector::Random => {
                #[cfg(target_pointer_width = "64")]
                let selection = rand::rng().next_u64() as usize;
                #[cfg(target_pointer_width = "32")]
                let x = rand::rng().next_u32() as usize;
                selection
            }
```
However `ThreadRng` does not provide a `next_u16`, therefore we would still have the same issue.

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+};
+
+use rama_core::error::OpaqueError;
+use rand::{seq::SliceRandom, RngCore};
+
+use crate::client::TcpStreamConnector;
+
+/// Selection algorithms
+#[derive(Debug, Clone)]
+pub enum Selector {
```
Babbleshack: `ThreadRng` is `!Send` & `!Sync`, therefore I did not store it on the selector.

If creating a new `ThreadRng` each time is too much overhead we could store a [random iterator](https://rust-random.github.io/rand/rand/trait.Rng.html#method.random_iter) and then manually impl `Clone` for `Selector`. Something like

```rust

impl Clone for Selector {
   fn clone(&self) -> Self {
    match self {
      Self::Random(iter) -> Self::Random(iter.cloned())
      Self::RoundRobin(atomic_usize) -> Self::RoundRobin(atomic_usize.clone())
    }
   }
}

```

rama-tcp/src/client/pool.rs
GlenDC: Agreed, elegant and better. Thanks for standing up :)

rama-tcp/src/pool/mod.rs
GlenDC: It's fine to do u64 for now :) There's plenty of the codebase that would potentially break on 32-bit machines. For now I am only aiming for 64 bit machines. It will require a large sponsor (financially or in code contributions and maintenance) to also support 32 bit platforms.

All servers these days are 32 bit machines, and don't think you can still buy a computer, phone or tablet that is still 32 bit. Anyway, not my concern for now. Thank you for bringing it up though.

If we ever need to support such things we probably want to ensure that anything that requires a difference between 32 bit and 64 bit to be in a single module / crate somewhere so we don't need this kind of logic everywhere. For now it is what it is.

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    connectors: Vec<C>,
+}
+
+impl<C: TcpStreamConnector> TcpStreamConnectorPool<C> {
+    /// A `TcpStreamConnector` where each connection is chosed randomly from a pool of
+    /// `TcpStreamConnector`s
+    pub fn new_random(connectors: Vec<C>) -> Self {
+        let mut rng = rand::rng();
+        let mut connectors = connectors;
+        connectors.shuffle(&mut rng);
```
GlenDC: You should be able to reduce this to

```
Self {
  selector: Selector::new_random(),
  connectors,
}
```
        
According to the mathematics of combinatorics you'll have just as much chance to get connector X regardless of this initial shuffle or not. As you each time choose a random index anyway...

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+    type Error = <C as TcpStreamConnector>::Error;
+
+    async fn connect(
+        &self,
+        addr: std::net::SocketAddr,
+    ) -> Result<tokio::net::TcpStream, Self::Error> {
+        let connector = self
+            .selector
+            .next(&self.connectors)
+            .ok_or(OpaqueError::from_display(
```
GlenDC: should be a `ok_or_else` as you do some work here

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+                "TcpStreamConnector has empty connectors collection",
+            ))?;
+        connector.connect(addr).await
```
GlenDC, line 79: <https://ramaproxy.org/docs/rama/tcp/client/trait.TcpStreamConnector.html#tymethod.connect>

^ `::connect` takes `&self`... There's no reason to `Clone` the selector in your `::next` method. You can just return a `Option<&C>` instead :)

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+                    .next(connectors.as_slice())
+                    .expect("Selector could not select from empty Connections collection")
+            })
+            .collect();
+
+        assert_eq!(results, expected, "Selector returned unexpected order");
+    }
+
+    #[test]
+    fn test_selector_random() {
```
GlenDC, line 121: you might also want to do a test with a lot of samples (e.g. 1000 samples for 3 connectors) to test that you have seen all connectors...

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,182 @@
...
+        assert!(
+            results.iter().all(|connector_opt| connector_opt.is_some()),
+            "Unexpected got None from selector",
+        );
+    }
+
+    #[test]
+    fn test_empty_selector() {
+        let connectors: Vec<usize> = vec![];
+        let selector = Selector::new_round_robin();
```
GlenDC: can you add the same one for random

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+};
+
+use rama_core::error::OpaqueError;
+use rand::RngCore;
+
+use crate::client::TcpStreamConnector;
+
+/// Selection algorithms
+#[derive(Debug, Clone)]
+pub enum Selector {
```
GlenDC:
```suggestion
enum Selector {
```

Sorry didn't see this before, but this doesn't need to be public I think given  the user never needs to construct it manually

GlenDC (approved):

Babbleshack:
Good catch ill update

GlenDC:
> Good catch ill update

Ah apologies I already did it for you. Wasn't sure you would still be around.
If the CI passes I'll merge it, as it looks good to me. Thx a lot for this first time contribution, very well done.

rama-tcp/src/pool/mod.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    use rama_net::address::SocketAddress;
+
+    use crate::{
+        client::TcpStreamConnector,
+        pool::{Selector, TcpStreamConnectorPool},
+    };
+
+    #[derive(Debug, Clone)]
+    struct MockConnector;
```
Babbleshack: @GlenDC actually I think we can remove this too

Babbleshack:
> > Good catch ill update
> 
> Ah apologies I already did it for you. Wasn't sure you would still be around. If the CI passes I'll merge it, as it looks good to me. Thx a lot for this first time contribution, very well done.

No problem at all, thanks for pointers.

PR #638, opened 2025-07-22
Macro to implement common connector traits

soundofspace:


rama-net/src/client/pool/mod.rs
```diff
@@ -500,76 +480,34 @@ impl<C, ID> Drop for LeasedConnection<C, ID> {
...
+    Socket for LeasedConnection<C, ID> where {
+        C: Socket,
+        ID: Send + Sync + 'static,
+    } {
+        ref self: {self.pooled_conn.as_ref().expect("only None after drop").conn}
     }
 
-    fn peer_addr(&self) -> std::io::Result<SocketAddr> {
-        self.as_ref().peer_addr()
+    AsyncWrite for LeasedConnection<C, ID> where {
```
soundofspace, line 498: @GlenDC this PR is far from finish but already wanted to get your opinion on how this macro works

General structure is:
```rust
impl_inner_traits! {

AsyncWrite for LeasedConnection<C, ID> where {
  C: ...
} {
  ref self: {inner_target}
  &mut self: {mutable_inner_target}
}
```
The structure is like that for some reasons:
1. Specify trait name of the trait you want to implement. Multiple traits can be implement in a single block but all of them have to be mentioned separately -> This is done because we dont always want to implement all of them, or some of them have different trait bounds (in the future I might add shortcut to make it easier to implement all / a bunch in one go)
2. Where { Generic }, needs to use `{` and `}` to isolate the generics, otherwise this is giving issues. I would prefer it without but didnt find a way todo that...
3. `ref sef` and `&mut self` to specify the targets. `& self` or `&self` is not supported so `ref self` instead. `&mut self` is supported so we use that. These prefixes are used since they align mostly with what you would expect in function signatures, but we could do something else also like `target self, mut_target: self`, `ref(self) {}, mut(self) {}` ... We do need to mention `self` in these signatures in one way or another to prevent macro hygiene issues.

Im mainly asking right now do you agree with syntax, or do you prefer changes before fully committing and using this everywhere.

GlenDC: Hmm I honestly am a bit turned off by it. I feel the amount of abstraction it adds for the little benefits is not really worth it. In my opinion it should be very clear when you read code that uses a macro what exactly it does and that is not at all clear here. Like you can guess, but it's not entirely clear how it will expand to etc...

For now I'm still more convinced to just drop this and have it manually be done.

Maybe I'm missing some things why it might be anyway worth persuing though.

soundofspace: I'm currently leaning towards the same opinion, the only upside of this macro is not forgetting certain default functions in impls (when you forget to place the clippy linter rule). But don't think that's a good enough reason to introduce a macro this complex also, hence why I kinda ignored this PR this long

GlenDC: In that case let's just close it for now.
Not sure if it's worth documenting somewhere things like this that we didn't persuit but could be revisited later.

The clipper linter rule is indeed what I like more and I would hope that sufficient testing (automatic, manual and production use) detect such mistakes quick enough with the rule afterwards continuing to protect us.

PR #641, opened 2025-07-25
Add metrics support to connection pool

maartendeprez:


maartendeprez:
Rebased on main

rama-net/src/client/pool/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+                .with_description("Connection pool reused connections")
+                .build(),
+            evicted_connections: meter
+                .u64_counter(prefix_metric(prefix, CONNPOOL_EVICTED_CONNECTIONS))
+                .with_description("Connection pool evicted connections")
+                .build(),
+            reused_connection_pos: meter
+                .u64_histogram(prefix_metric(prefix, CONNPOOL_REUSED_CONNECTION_POS))
+                .with_description("Connection pool reused connection position in pool")
+                .with_boundaries(vec![0_f64, 1_f64, 2_f64, 3_f64, 4_f64, 5_f64, 6_f64])
```
soundofspace: These values will only work for very small pools, we might want something like 2, 4, 8, 16, 32 here (until exponentional histogram is supported)?

rama-net/src/client/pool/metrics.rs
```diff
@@ -0,0 +1,127 @@
...
+
+    #[must_use]
+    pub fn new_with_meter(meter: &Meter, opts: MeterOptions) -> Self {
+        let service_info = opts.service.unwrap_or_else(|| ServiceInfo {
+            name: rama_utils::info::NAME.to_owned(),
+            version: rama_utils::info::VERSION.to_owned(),
+        });
+
+        let mut attributes = opts.attributes.unwrap_or_else(|| Vec::with_capacity(2));
+        attributes.push(KeyValue::new(SERVICE_NAME, service_info.name));
```
soundofspace, line 81: Do we need to set this every time still or can we also define these globally once (might need smallish metric refactor in rama, but setting them once globally is a very nice improvement)

soundofspace (changes requested):
Two minor remarks, supporting bigger pools in metrics should be done in some way, but having global attributes can be done separately or now depending on what you prefer

rama-net/src/client/pool/metrics.rs
maartendeprez: This can be done globally by adding a resource to the meter set with `opentelemetry::global::set_meter_provider` indeed. Changing this in rama would entail updating `MeterOptions` to only allow specifying a `metric_prefix`. Can do that in a separate PR.

rama-net/src/client/pool/metrics.rs
maartendeprez: Good point, modified the bounds

rama-net/src/client/pool/metrics.rs
```diff
@@ -128,3 +125,68 @@ impl PoolMetrics {
...
+            self.reused_connection_pos_bounds = bounds;
+            self
+        }
+    }
+
+    generate_set_and_with! {
+        /// Calculate exponentially-spaced bounds for the `reused_connection_pos` metric.
+        pub fn reused_connection_pos_parametrized_bounds(mut self, max: f64, nbounds: usize) -> Self {
+            self.reused_connection_pos_bounds = (0..nbounds)
+                .map(|i| (i as f64 / ((nbounds - 1) as f64) * max.log2()).exp2())
```
soundofspace: Both of these can be simplified (sorry for the ugly math)
```
using: y ^ (x * log_y(z) ) = z ^ x

x = i as f64 / ((nbounds - 1) as f64

(x * max.log2()).exp2())  =  2 ^ ( log2(max) * x) = max ^ x = max ^ ( i as f64 / ((nbounds - 1) as f64 )

map(|i| max.powf(i as f64 / (nbounds - 1) as f64))
```

Because pc's suck at float stuff all of these variants can and will produce things like `9.9999999`, so you probably want to do some rounding here, not sure if you want decimals at all in this specific setup?

soundofspace (changes requested):
Thanks for the change, one last remark from my side, and then I will the final review to @GlenDC

rama-net/src/client/pool/metrics.rs
maartendeprez: Don't think decimals are useful here. Simplified and rounded the result.

maartendeprez: Don't think decimals are useful here. Simplified and rounded the result.

Issue #587, opened 2025-06-02
Add distributed session/application layer tests using turmoil


GlenDC:
With Turmoil (https://github.com/tokio-rs/turmoil) we can add chaos to the transport layer (UDP / TCP) and serve data for session / application layer stacks. Could be useful to add some more guarantees.

Inspiration for the kind of tests that we could do can be found in https://github.com/tokio-rs/turmoil/tree/main/tests. Whoever picks this issue up first should list the kind of tests they'll write.

I would keep these tests separate from other tests and just add them under `./tests-turmoil/`, each test its own cargo project there. This way we can add them to the workspace via the pattern `./tests-turmoil/*` in our cargo toml of the rama workspace.

The resolution to this ticket can be as small as just picking 1 or 2 use cases already. E.g. http client<->server, or with a proxy in the middle, etc.. Plenty of stuff to pick from.

It's a good first issue as one doesn't need to go very in depth in rama knowledge, but one does need sufficiently good rust knowledge as there's plenty to do here, so it isn't easy by any means. Just no difficulties specific to rama I would think.

Babbleshack:
This ticket looks interesting and I would like to have a think about it.

>  Could be useful to add some more guarantees.

I suspect these are related to proxy management, e.g., reconnection to upstream, smoothing jittery connections, etc.

Are there any specific modules/features I might consider for testing?

GlenDC:
Sure have a think about it, if you at some point think you can do it and want to do it, do let it be known and it is yours if still available.

I would test all kind of stuff with it (eventually and if possible)

- http/1.1
   - client <-> server
   - client <-> transport proxy <-> server
   - client <-> mitm proxy <-> server
- ws (skip handshake)
   - client <-> server
   - client <-> transport proxy <-> server
   - client <-> mitm proxy <-> server
- h2
   - client <-> server
   - client <-> transport proxy <-> server
   - client <-> mitm proxy <-> server

But for a first PR I would start simple. 
e.g. just a `http/1.1` client <-> server test.

Babbleshack:
Sure I will take have a go at this.

Issue #533, opened 2025-04-29
add ffi/rama-rhai: support ability to use services and layers written in rhai


GlenDC:
Similar to <https://github.com/plabayo/rama/issues/532>, but trading off flexibility for a lot of simplicity. As writing rhai scripts is a whole lot easier (for most) than writing and compiling wasm modules.

Is to be added under the `ffi` folder, meaning the crate will live under `./ffi/rama-rhai`

Initial goals:

* support for http only
   * support the ability for an layer service (meaning it will get a blackbox http service that it needs to sandwich
   * as well as the ability for an endpoint/leaf service, meaning it needs to return the content itself (response)

Will need to investigate if we can make use of nice features such as `IntoResponse` to make it easy to return common responses instead of having to build a response fully. To be investigated by the one implementing this.

An example + test of that example, as well as some unit tests are all also to be added as part of the work on this issue.

GlenDC:
@jc65536 are you still working on this? Asking as @Babbleshack has shown interest to pick it up and I rather not have these kind of issues be hogged for too long without proof of effort and progress. There will always be work to be done in future in the meanwhile I do want issues to move forward. With our guidance and help of course, always available.

jc65536:
I haven't had time to work on this recently. I can push my work so far, and Babbleshack can take over.

Babbleshack:
> I haven't had time to work on this recently. I can push my work so far, and Babbleshack can take over.

Hi @jc65536 I certainly dont want to take the work from you if your still interested. I hadnt noticed it was assigned to you when I mentioned it to @GlenDC.

I dont want to speak for @GlenDC, but from my perspective, if you would rather continue working on this, than please do!

jc65536:
I don't mind if others take over my issue. It's honestly been way too long since I've been able to spend time on this project, and things do need to move forward. If you're interested, go right ahead. If no one's interested in this issue, I'll come back to this when I have the means to.

Babbleshack:
@jc65536 Okay after I finish my current work, Ill message you here and if you still happy to hand over, then Ill take over.

GlenDC:
Great, all the same for me. Rama is anyway not gonna be a project that will be ever finished. There will always be plenty of work to be done, and even if you do not find anything that's already linked to an issue, I'm sure we can always come up with something meaningful to do which brings you joy. We're here to empower you, not only when using rama but also to develop with all of us.

jc65536:
[...]

Issue #532, opened 2025-04-29
add ffi/rama-wasm: support using services and layers written in wasm


GlenDC:
Real-world examples of systems supporting it are for example envoy:

* cc example: <https://github.com/envoyproxy/examples/tree/main/wasm-cc>
* docs:
   * intro: <https://gateway.envoyproxy.io/docs/tasks/extensibility/wasm/>
   * more: <https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/wasm-cc>

another rama-crate that can live under a new folder `ffi` that we'll use for extensions and bindings to FFI's
so the location of the crate will be `./ffi/rama-wasm`

GlenDC:
An alternative idea for the layer service interface would be to not pass the inner service at all to the ffi extension but instead have the service layer have these 2 methods : `modRequest: reqA -> reqB` and `modResponse: resA -> resB`, not sure if needed to be fallible or not.

Seems there are already systems out there like `wit` and `wasm-http` so we can probably get inspiration (if not use them) on how we want to potentially expose this.

GlenDC:
Initial support will focus on middleware proxy support of http layer

Future follow up work outside this ticket can do:

* improve that initial support
* support other protocols:
   * websocket
   * SSE
 * support leaf service providers

PR #642, opened 2025-07-25
Draft: Feature 587 turmoil http.1 client-server test

Babbleshack:
Adds a simple http.1 client - server test using tokio-turmoil

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ members = [
     "rama-udp",
     "rama-unix",
     "rama-utils",
-    "rama-ws",
+    "rama-ws", "tests-turmoil/test-client-server",
 ]
 
 [workspace.package]
--- a/tests-turmoil/test-client-server/Cargo.toml
+++ b/tests-turmoil/test-client-server/Cargo.toml
@@ -1,5 +1,6 @@
 [package]
 name = "test-client-server"
+description = "Client-Server turmoil testiong"
 version.workspace = true
 license.workspace = true
 edition.workspace = true
@@ -11,6 +12,9 @@ authors.workspace = true
 rust-version.workspace = true
 
 [dependencies]
+rama = { path = "../.." }
+turmoil = "0.6"
+tokio = { workspace = true }
 
 [lints]
 workspace = true
--- a/tests-turmoil/test-client-server/Cargo.toml
+++ b/tests-turmoil/test-client-server/Cargo.toml
@@ -12,9 +12,13 @@ authors.workspace = true
 rust-version.workspace = true
 
 [dependencies]
-rama = { path = "../.." }
+rama = { path = "../..", features = ["full"] }
 turmoil = "0.6"
 tokio = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
 
 [lints]
 workspace = true
--- a/tests-turmoil/test-client-server/Cargo.toml
+++ b/tests-turmoil/test-client-server/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "test-client-server"
+version.workspace = true
+license.workspace = true
+edition.workspace = true
+repository.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+categories.workspace = true
+authors.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+
+[lints]
+workspace = true
--- a/tests-turmoil/test-client-server/src/main.rs
+++ b/tests-turmoil/test-client-server/src/main.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("Hello, world!");
+}
--- a/tests-turmoil/test-client-server/src/main.rs
+++ b/tests-turmoil/test-client-server/src/main.rs
@@ -1,3 +1,13 @@
+use turmoil::Builder;
+
 fn main() {
-    println!("Hello, world!");
+    // NOTE:
+    // To make your tests deterministic, you can use your own seeded `rng` provider when building
+    // the simulation through `build_with_rng`.
+
+    let mut sim = Builder::new().build();
+
+    sim.host("server", || async { Ok(()) });
+
+    sim.client("client", async { Ok(()) });
 }
--- a/tests-turmoil/test-client-server/src/main.rs
+++ b/tests-turmoil/test-client-server/src/main.rs
@@ -1,13 +1,73 @@
+use rama::{
+    http::{
+        client::EasyHttpWebClient, layer::trace::TraceLayer, server::HttpServer,
+        service::web::WebService, Body, BodyExtractExt, Request,
+    },
+    net::address::SocketAddress,
+    Context, Layer, Service,
+};
+use tracing::{info, info_span, Instrument};
+use tracing_subscriber::{
+    filter::LevelFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt, EnvFilter,
+};
 use turmoil::Builder;
 
+const ADDRESS: SocketAddress = SocketAddress::default_ipv4(62004);
+
+fn setup_tracing() {
+    tracing_subscriber::registry()
+        .with(fmt::layer())
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::DEBUG.into())
+                .from_env_lossy(),
+        )
+        .init();
+}
+
+async fn start_server(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    HttpServer::http1()
+        .listen(
+            address.into(),
+            (TraceLayer::new_for_http()).into_layer(WebService::default().get("/", "Hello, World")),
+        )
+        .await
+        .map_err(|e| e as Box<dyn std::error::Error>)
+}
+
+async fn run_client(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    let client = (TraceLayer::new_for_http(),).into_layer(EasyHttpWebClient::default());
+    let resp = client
+        .serve(
+            Context::default(),
+            Request::builder()
+                .uri(format!("http://{address}/", address = address.into()))
+                .method("GET")
+                .body(Body::empty())
+                .unwrap(),
+        )
+        .await?;
+    let body = resp.try_into_string().await.unwrap();
+    assert_eq!(body, "Hello, World");
+    Ok(())
+}
+
 fn main() {
-    // NOTE:
-    // To make your tests deterministic, you can use your own seeded `rng` provider when building
-    // the simulation through `build_with_rng`.
+    setup_tracing();
+    let mut sim = Builder::new().enable_tokio_io().build();
 
-    let mut sim = Builder::new().build();
+    sim.host(ADDRESS.ip_addr(), || {
+        start_server(ADDRESS).instrument(info_span!("server"))
+    });
 
-    sim.host("server", || async { Ok(()) });
+    sim.client(
+        "client",
+        run_client(ADDRESS).instrument(info_span!("client")),
+    );
 
-    sim.client("client", async { Ok(()) });
+    sim.run().expect("Error during simulation");
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1787,6 +1787,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "libm"
+version = "0.2.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f9fbbcab51052fe104eb5e5d351cf728d30a5be1fe14d9be8a3b097481fb97de"
+
+[[package]]
 name = "libmimalloc-sys"
 version = "0.1.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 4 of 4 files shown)


Babbleshack:
@GlenDC or @soundofspace I was wondering if I could get some early feedback here.

Am interested to know if this is what you guys had in mind? 

What do you think of the package structure?

Any changes you would like to see?

Cargo.toml
```diff
@@ -48,7 +48,7 @@ members = [
     "rama-udp",
     "rama-unix",
     "rama-utils",
-    "rama-ws",
+    "rama-ws", "tests-turmoil/test-client-server",
```
GlenDC: give entries their own line please

tests-turmoil/test-client-server/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+edition.workspace = true
+repository.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+categories.workspace = true
+authors.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+rama = { path = "../..", features = ["full"] }
```
GlenDC:
```suggestion
rama = { workspace = true, features = ["full"] }
```

tests-turmoil/test-client-server/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+turmoil = "0.6"
```
GlenDC: add as a workspace dep first and than use { workspace = true }

tests-turmoil/test-client-server/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+tokio = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+tracing = { workspace = true }
```
GlenDC: use from rama-core

tests-turmoil/test-client-server/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    },
+    net::address::SocketAddress,
+};
+use tracing::{Instrument, info_span};
+use tracing_subscriber::{
+    EnvFilter, filter::LevelFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt,
+};
+use turmoil::Builder;
+
+const ADDRESS: SocketAddress = SocketAddress::default_ipv4(62004);
```
GlenDC: you made sure this port is not used by any other test?

tests-turmoil/test-client-server/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+        .init();
+}
+
+async fn start_server(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    HttpServer::http1()
+        .listen(
+            address.into(),
+            (TraceLayer::new_for_http()).into_layer(WebService::default().get("/", "Hello, World")),
```
GlenDC: useless round brackets

tests-turmoil/test-client-server/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+        )
+        .await
+        .map_err(|e| e as Box<dyn std::error::Error>)
+}
+
+async fn run_client(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    let client = (TraceLayer::new_for_http(),).into_layer(EasyHttpWebClient::default());
```
GlenDC:
```suggestion
    let client = TraceLayer::new_for_http().into_layer(EasyHttpWebClient::default());
```

tests-turmoil/test-client-server/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+            address.into(),
+            (TraceLayer::new_for_http()).into_layer(WebService::default().get("/", "Hello, World")),
+        )
+        .await
+        .map_err(|e| e as Box<dyn std::error::Error>)
+}
+
+async fn run_client(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
```
GlenDC:
```suggestion
) -> Result<(), BoxError> {
```

(from rama-core)

tests-turmoil/test-client-server/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+async fn start_server(
+    address: impl Into<SocketAddress>,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    HttpServer::http1()
+        .listen(
+            address.into(),
+            (TraceLayer::new_for_http()).into_layer(WebService::default().get("/", "Hello, World")),
+        )
+        .await
+        .map_err(|e| e as Box<dyn std::error::Error>)
```
GlenDC:
```suggestion
        .map_err(Into::into)
```

Babbleshack:
> Looks fine to me. Left some feedback already. Besides that, two questions:
> 
>     1. Why does this need to be a separate package? Can this not be just another main application in `/tests`?
> 
>     2. What is the output when you run this? Is it just binary it works or fails? are there warnings. Asking as I haven't played with turmoil myself yet.

1. I think I mis-interpretted your instructions from the ticket (#587).

> I would keep these tests separate from other tests and just add them under ./tests-turmoil/, each test its own cargo project there. This way we can add them to the workspace via the pattern ./tests-turmoil/* in our cargo toml of the rama workspace.

Now I have re-read this I think what you actually wanted was integration tests in a `/tests/turmoil` package. Ill make this change.

>     2. What is the output when you run this? Is it just binary it works or fails? are there warnings. Asking as I haven't played with turmoil myself yet.

There are no outputs unless logging is enabled. 

If the assertion fails, the binary panics as expected.

If a `Result::Err` is returned from one of the async closure than it is returned similar to `JoinHandle` behaviour.

For example if we do:

```rust
    let mut sim = Builder::new().enable_tokio_io().build();
    sim.host(ADDRESS.ip_addr(), || async {
       Err("error returned from server".into())
    });
    
    sim.client(...);
    
    sim.run().expect("Error during simulation");
    
 ```
 
 ```bash
 
 $ cargo run
 ...
 2025-07-25T19:30:18.246122Z  INFO turmoil: New nodename="0.0.0.0" addr=0.0.0.0
2025-07-25T19:30:18.247514Z  INFO turmoil: New nodename="client" addr=192.168.0.1

thread 'main' panicked at tests-turmoil/test-client-server/src/main.rs:73:23:
Error during simulation: "error returned from server"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

GlenDC:
Is there no bookkeeping that can be checked to see what inputs and parameters it used during the tests? And how to know it ran at all?

Babbleshack:
> Is there no bookkeeping that can be checked to see what inputs and parameters it used during the tests? And how to know it ran at all?

There are some functions for inspecting the simulation state:
- https://docs.rs/turmoil/latest/turmoil/fn.established_tcp_stream_count.html
- https://docs.rs/turmoil/latest/turmoil/fn.established_tcp_stream_count_on.html

The simulation ([struct Sim](https://docs.rs/turmoil/latest/turmoil/struct.Sim.html)) provide some functions for manipulating the environment.

However after some more investigation I dont believe this test is using the turmoil net types. 

```
$ RUST_LOG=INFO,turmoil=TRACE cargo test --test turmoil --features http-full
    Finished `test` profile [optimized] target(s) in 0.30s
     Running tests/turmoil/main.rs (target/debug/deps/turmoil-e998ab69135e2f65)

running 1 test
2025-07-25T21:02:52.748036Z  INFO turmoil: New nodename="0.0.0.0" addr=0.0.0.0
2025-07-25T21:02:52.748150Z  INFO turmoil: New nodename="client" addr=192.168.0.1
2025-07-25T21:02:52.748167Z TRACE turmoil: step 1
2025-07-25T21:02:52.748389Z TRACE turmoil: step 2
2025-07-25T21:02:52.748559Z TRACE turmoil: step 3
2025-07-25T21:02:52.748604Z TRACE turmoil: step 4
test http::http_1_client_server_it ... ok
```

The output should show network traffice as it sent and delivered. 

I think I need to adjust the client and server so that they are using [`turmoil::net`](https://docs.rs/turmoil/latest/turmoil/net/index.html) types. 

Something more like:

```
2023-11-29T20:23:43.277039Z TRACE node{name="client"}: turmoil: Delivered src=192.168.0.1:9999 dst=192.168.0.2:49152 protocol=TCP [...]
2023-11-29T20:23:43.277097Z TRACE node{name="client"}: turmoil: Recv src=192.168.0.1:9999 dst=192.168.0.2:49152 protocol=TCP [..]
2023-11-29T20:23:43.277324Z  INFO client: axum_example: Got response: Response { status: 200, version: HTTP/1.1, headers: {"content-type": "text/plain; charset=utf-8", "content-length": "10", "date": "Wed, 29 Nov 2023 20:23:43 GMT"}, body: b"Hello foo!" }
...

```

We could use something like [tracing_test](https://docs.rs/tracing-test/latest/tracing_test/) to scrape logs and assert specific network events.

Although its not great. I think a better solution might be to add counters to the [turmoil::Sim](https://docs.rs/turmoil/latest/turmoil/struct.Sim.html) type

tests-turmoil/test-client-server/src/main.rs
Babbleshack: `pub type BoxError = Box<dyn StdError + Send + Sync>` wheras I actually need a `Box<dyn StdError + 'static>` here

tests-turmoil/test-client-server/src/main.rs
Babbleshack: This wont work, but also here we are trying to coerce the type so that w ecan drop the `Send + Sync` trait bounds.

GlenDC:
Cool I think what you have of tracing and output is probably good enough for now, unless you think differently.

tests-turmoil/test-client-server/src/main.rs
Babbleshack: In turmoil the network stack is simulated so no need to check for collisions.

Cargo.toml
```diff
@@ -405,6 +405,7 @@ spmc = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tokio-test = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
+turmoil = "0.6"
```
GlenDC: you need to add it as a workspace dependency first (higher up in this file), this line will at that point become `turmoil = { workspace = true }`. Reason being so I only ever need to care about one specific location for _all_ dependency versions. Makes audits easier.

tests/turmoil/http.rs
```diff
@@ -0,0 +1,91 @@
...
+use http::Version;
+use rama::{
+    Context, Layer, Service,
+    http::{
+        Body, BodyExtractExt, Request, layer::trace::TraceLayer, server::HttpServer,
+        service::web::WebService,
+    },
+    net::address::SocketAddress,
+};
+use rama_http_backend::client::EasyHttpWebClientBuilder;
```
GlenDC: I know IDE's do not help with it, but please import it as `rama::http::client::EasyHttpWebClientBuilder`

tests/turmoil/http.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+async fn start_server(
+    address: impl ToSocketAddrs,
+) -> Result<(), Box<dyn std::error::Error + 'static>> {
+    let listener = turmoil::net::TcpListener::bind(address).await?;
+
+    let conn_result = tokio::time::timeout(Duration::from_secs(1), listener.accept())
+        .await
+        .map_err(|_| "accept timeout")?;
```
GlenDC: `.context("accept timeout")?` should still work, no?

GlenDC (changes requested):
Awesome, only a couple of nitpicks left. But this is coming along very nice. When you're done with these last changes I think we can merge. I'll ask a second pair of eyes from @soundofspace to be sure, but I think we're close to good to go. Thank you for this.

Babbleshack:
> Awesome, only a couple of nitpicks left. But this is coming along very nice. When you're done with these last changes I think we can merge. I'll ask a second pair of eyes from @soundofspace to be sure, but I think we're close to good to go. Thank you for this.

Thanks! I have made the requested changes, after this has merged ill start on some of the other test cases.

As a side note, some time later I would like to investigate #532 or #533.

I am hessitant to work on #532 because it seems WASM has become alot less popular, additionally I took a look at the rust wasm working group and it seems to have been disbanded.

I have no experience with WASM, and I would be using it as a learning opportunity. Whilst I think its a decent idea/use case, am unsure how much it will be used.

Do you still want to integrate with WASM for this use case? 

Finally the code base for Rama is quite a bit more complex than any other Rust code base I have worked on, as such from a arhcitectural point perspective, I might need some guidance. Is this something I would be able to expect from either yourself or _soundofspace_ ? (I understand you are both probably quite busy).

GlenDC:
Yeah both issues are still planned to be done. The WASM issue I am gonna do myself as I will use the experience to also make some proposals in the WASM working group as follow-up items.

The Roto one you can probably pick up as the previous person interested in tackling it has shown no activity since requesting to do it. We'll see what that person comes back with by the time this PR gets merged or tomorrow.

GlenDC:
And yes if you would pick up an issue like Roto both @soundofspace and I will be there for you to help figure out the right architecture. It's probably less than what you think though. Most of the work will be the learning curve of how Roto itself works as the integration AFAIK will just be Service implementations. But that's something we can talk about in a different location. It is out of scope for this PR.

Babbleshack:
Great, when I find time I'll do another pr for the other tests.

GlenDC:
Great job @Babbleshack , wonderful contribution you did. Looking forward to more tests :) I imagine they will all just pass for now but once we start doing all kind of optimizations somewhere in rama 0.4 and beyond I think checks like this might add a nice additional safety net.

Issue #621, opened 2025-07-07
Implement all JWA algorithms


soundofspace:
Currently we only support EC algorithms, goal of this issue is to also support HMAC and and RSA variants (if aws_lc_rs supports them).

This needs changes in:
- `rama-crypto/src/jose/jwa.rs` to convert to appropriate `aws_lc_rs` types
- `rama-crypto/src/jose/jwk.rs` to support working with these algorithms (encoding, decoding, utility functions such as unparsed_public_key...)

AydanPirani:
Hey, found this through the TWIR607! Do you still need help with this?

GlenDC:
> Hey, found this through the TWIR607! Do you still need help with this?

Hi @AydanPirani , welcome to the club. We keep issues like these open for the community as opportunities to (a) get you engaged into the framework and our community and (b) a learning opportunity. In return @soundofspace and I are here to mentor and guide you. So in case you are missing details, have questions, remarks, things you wish to discuss. Let us know.

I'll assign it to you in the meanwhile. If you do decide now or later to not want to pursuit it do let us know and we'll un-assign you without an issue. We understand choices can change and life can happen.

valkrypton:
Hey, Is anyone currently working on this? If not, I'd like to try this

GlenDC:
@AydanPirani is working on this I think, he can confirm to be sure. If he isn't and he is ok with it you can have it @valkrypton . But I want to leave that choice up to him. In case he doesn't respond within 10 days we can force the decision and assign it to you already if you still want it, as we have a no-hording policy.

AydanPirani:
@GlenDC  I am not - all you @valkrypton !

GlenDC:
Thank you for coming back to us @AydanPirani ! 
@valkrypton let me or @soundofspace know if you need any guidance or have any questions.
Feel free to open a draft PR as soon as you desire, to ask for (early) feedback or to help unblock you.
Best of luck.

PR #650, opened 2025-07-31
implement jwa algorithms

valkrypton:
working on #621.

@GlenDC I've written the conversions in `rama-crypto/src/jose/jwa.rs` file. Let me know if they are correct. However, I'd like some guidance on what changes needs to be made in the `rama-crypto/src/jose/jwk.rs` file.

### DIFF
--- a/rama-crypto/src/jose/constants.rs
+++ b/rama-crypto/src/jose/constants.rs
@@ -35,6 +35,18 @@ mod rsa_algorithm_identifier {
     const NULL_TAG: u8 = 0x05;
     const LENGTH_NULL: u8 = 0x00;
 
+    /// This entire thing has 3 layers, and the final goal is to
+    /// get the der algorithm identifier for rsa encoding
+    ///
+    /// 1. The identifier oid: 1.2.840.113549.1.1.1 defined in RFC 8017
+    /// 2. RFC 3279 (Algorithms and Identifiers for the Internet X.509 PKI)
+    ///    specifies tag is needs to be NULL. The general structure is
+    ///    IDENTIFIER, PARAMETER, but for rsa here we dont have PARAMETER
+    ///    so NULL instead
+    /// 3. We need to combine everything in DER encode
+    ///   - 3.1 (IDENTIFIER, PARAMETER) is a sequence, so SEQUENCE_TAG, LENGHT of what follows
+    ///   - 3.2 OBJECT_IDENTIFIER_TAG to specify OID, LENGTH of OID,and actual oid
+    ///   - 3.3 NULL_TAG and LENGTH_NULL to encode null value
     pub(crate) const RSA_ALGORITHM_IDENTIFIER: [u8; 15] = [
         SEQUENCE_TAG,
         LENGTH,


GlenDC:
I've asked @soundofspace to do the initial reviews and guidance here. He'll come back to you.

rama-crypto/src/jose/jwa.rs
```diff
@@ -99,3 +104,36 @@ impl TryFrom<JWA> for &'static EcdsaVerificationAlgorithm {
...
+            JWA::HS384 => Ok(&HMAC_SHA384),
+            JWA::HS512 => Ok(&HMAC_SHA512),
+            _ => Err(OpaqueError::from_display(
+                "Non-Hmac algorithm cannot be converted to hmac types",
+            )),
+        }
+    }
+}
+
+impl TryFrom<JWA> for Box<&dyn RsaEncoding> {
```
soundofspace: Any specific reason to box this and not use `&'static RsaSignatureEncoding` instead?

soundofspace:
For changes in `rama-crypto/src/jose/jwk.rs` I would start with RSA as it behaves quite similar to EC curves.

Places that need updating:
- Add `JWK::new_from_rsa_keypair` using `RsaKeyPair` from `awc_lc_rs`. Here you create key_type `JWKType::RSA` with base64 encoded values for n, e (using `RsaKeyPair.public_key().modules/exponent`)
- In `JWK::unparsed_public_key` implement the reverse so we can get an UnparsedPublicKey 
- Add `RsaKey` very similar to `EcdsaKey` but modify where needed to use RSA stuff
- Rename `EcdsaKeySigningHeaders` to `SigningHeaders` so you can also use it to sign for RSA
- Implement `Signer` for `RsaKey`, this logic can probably be copied from the implementation for `EcdsaKey` but might need small tweaks
- Add some tests to verify that everything works like expected. These can be hardcoded to some expected output for a given input, there are lots of other projects online that have already a very nice test suite for this if you need inspiration
- If you see opportunities to extend the current (very minimal) test suite please also do, it would appreciated a lot

Also don't be afraid to look at the [jwk rfc](https://datatracker.ietf.org/doc/html/rfc7517) it's a decently sized one but it's split up in a lot of chapters and it also contains some examples

And thanks for looking into this!

rama-crypto/src/jose/jwa.rs
valkrypton: `RsaSignatureEncoding` is not exposed outside of the crate.

soundofspace: Missed that it was not exposed, in that case this is indeed a good approach

valkrypton: Btw my i've been busy with my day job so this issue slipped my mind. I'll make a complete PR in a few days

soundofspace: No worries, no rush here

GlenDC: > Missed that it was not exposed, in that case this is indeed a good approach

@soundofspace do you think that's a mistake of that crate? If so can you perhaps check with those crate authors to get that improved so we can do perhaps better here in future?

soundofspace: @GlenDC already did that a few hours ago

rama-crypto/src/jose/jwk.rs
```diff
@@ -156,7 +159,48 @@ impl JWK {
...
+                let mut final_sequence = Vec::new();
+                final_sequence.extend_from_slice(&algorithm_identifier);
+                final_sequence.extend_from_slice(&bit_string);
+
+                let final_len = utils::encode_der_length(final_sequence.len());
+                let mut result = vec![0x30];
+                result.extend_from_slice(&final_len);
+                result.extend_from_slice(&final_sequence);
+
+                Ok(signature::UnparsedPublicKey::new(self.alg.try_into()?, result))
```
valkrypton: @soundofspace 
The `UnparsedPublicKey::new` takes trait objects of type `VerificationAlgorithm` as argument. And `JWA.alg` is of type `dyn RsaEncoding`. So I couldn't figure out a way to convert our `alg` field to that type. Any guidance would be appreciated.

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +349,118 @@ impl Signer for EcdsaKey {
...
+    }
+
+    fn sign(&self, data: &str) -> Result<Self::Signature, Self::Error> {
+        let mut sig: Vec<u8> = Vec::new();
+        let alg = <Box<&dyn RsaEncoding> as TryFrom<JWA>>::try_from(self.alg)?;
+        self.inner
+            .sign(*alg, self.rng(), data.as_bytes(), &mut sig)
+            .context("sign protected data")?;
+        // problem here
+        Ok(sig.into())
```
valkrypton: The sign method for `RsaKeyPair` does not return the `Signature` and the `Signature` type does not expose it's `new` method. So there is no way to construct that type. What should we do here?

rama-crypto/src/jose/jwa.rs
```diff
@@ -99,3 +106,36 @@ impl TryFrom<JWA> for &'static EcdsaVerificationAlgorithm {
...
```
soundofspace: Can you change this to `&'static dyn RsaEncoding` so no allocations are needed

rama-crypto/src/jose/jwa.rs
```diff
@@ -99,3 +106,36 @@ impl TryFrom<JWA> for &'static EcdsaVerificationAlgorithm {
...
+            JWA::RS512 => Ok(Box::new(&RSA_PKCS1_SHA512)),
+            JWA::PS256 => Ok(Box::new(&RSA_PSS_SHA256)),
+            JWA::PS384 => Ok(Box::new(&RSA_PSS_SHA384)),
+            JWA::PS512 => Ok(Box::new(&RSA_PSS_SHA512)),
+            _ => Err(OpaqueError::from_display(
+                "Non-RSA algorithm cannot be converted to rsa types",
+            )),
+        }
+    }
+}
```
soundofspace, line 156: If you also add these it should solve the issue you have with converting

```rust
impl TryFrom<JWA> for &'static dyn VerificationAlgorithm {
    type Error = OpaqueError;

    fn try_from(value: JWA) -> Result<Self, Self::Error> {
        match value {
            JWA::RS256 => Ok(&RSA_PKCS1_2048_8192_SHA256),
```

soundofspace: And once that is is added, the logic from this one, should also be placed in there
```impl TryFrom<JWA> for &'static EcdsaVerificationAlgorithm {```

Something along these lines:
```rust
           JWA::PS512 => Ok(&RSA_PSS_2048_8192_SHA512),
            JWA::ES256 | JWA::ES384 ... => {
                let signing_algo: &'static EcdsaSigningAlgorithm = value.try_into()?;
                Ok(signing_algo.deref())
            }
```
Otherwise there might be issues with try_into overlapping for the use case of verification

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +349,118 @@ impl Signer for EcdsaKey {
...
+    }
+
+    #[must_use]
+    pub fn rng(&self) -> &SystemRandom {
+        &self.rng
+    }
+}
+
+impl Signer for RsaKey {
+    type Signature = Signature;
```
soundofspace: For Rsa signing this is a different signature

```suggestion
    type Signature =  Vec<u8>;
```
So no need to do extra try_into in the sign method

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +349,118 @@ impl Signer for EcdsaKey {
...
+
+mod utils {
+    pub(super) fn encode_der_length(len: usize) -> Vec<u8> {
+        if len < 128 {
+            vec![len as u8]
+        } else {
+            let mut len_bytes = len.to_be_bytes().to_vec();
+            while len_bytes[0] == 0 {
+                len_bytes.remove(0);
+            }
```
soundofspace, line 444:
```suggestion
            len_bytes.retain(|&x| x != 0);
```

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +349,118 @@ impl Signer for EcdsaKey {
...
+
+    pub(super) fn encode_integer(mut bytes: Vec<u8>) -> Vec<u8> {
+        if bytes[0] & 0x80 != 0 {
+            bytes.insert(0, 0x00);
+        }
+        let len_bytes = encode_der_length(bytes.len());
+        let mut result = vec![0x02];
+        result.extend_from_slice(&len_bytes);
+        result.extend(bytes.into_iter());
+        result
```
soundofspace: For efficiency its better if you can allocate the vector up front (and also avoid the bytes.insert), should be possible here with a small refractor, something like

```rust
let needs_leading_zero = bytes[0] & 0x80 != 0;
let value_len = bytes.len() + needs_leading_zero as usize;
let len_bytes = encode_der_length(value_len);
let mut result = Vec::with_capacity(1 + len_bytes.len() + value_len);
//push and extend all parts
```

soundofspace (changes requested):
Nice progress, and thanks for looking into this!

rama-crypto/src/jose/jwk.rs
```diff
@@ -156,7 +159,51 @@ impl JWK {
...
+                let n_bytes = BASE64_URL_SAFE_NO_PAD
+                    .decode(n)
+                    .context("decode RSA modulus (n)")?;
+                let e_bytes = BASE64_URL_SAFE_NO_PAD
+                    .decode(e)
+                    .context("decode RSA exponent (e)")?;
+                let n_der_encoded = utils::encode_integer(n_bytes);
+                let e_der_encoded = utils::encode_integer(e_bytes);
+
+                let mut rsa_public_key_sequence = Vec::new();
```
soundofspace: It should be possible to create all the vec's in this block with `Vec::with_capacity` to prevent extra allocations and moves

rama-crypto/src/jose/jwk.rs
```diff
@@ -156,7 +159,51 @@ impl JWK {
...
+                rsa_public_key_sequence.extend_from_slice(&n_der_encoded);
```
soundofspace: Since you don't need these "original" vectors later you can you `extend` instead which takes ownership and is more optimized

rama-crypto/src/jose/jwk.rs
```diff
@@ -180,6 +227,27 @@ impl JWK {
...
+
+    /// Creates a new [`JWK`] from a given [`RSAKeyPair`]
+    pub fn new_from_rsa_key_pair(rsa_key_pair: &RsaKeyPair, alg: JWA) -> Result<Self, OpaqueError> {
+        let n = rsa_key_pair.public_key().modulus();
+        let e = rsa_key_pair.public_key().exponent();
+        Ok(Self {
+            alg,
+            key_type: JWKType::RSA {
+                n: String::try_from(n.big_endian_without_leading_zero().to_vec())
+                    .map_err(|e| OpaqueError::from_display(format!("{:?}", e)))?,
```
soundofspace: n and e should be base64 encoded bytes which is not the same as `String::try_from`, that one does utf8 encoding, instead use

```
 BASE64_URL_SAFE_NO_PAD.encode(n)
```

soundofspace: Can you also try to add some unit tests which should detect mistakes like this

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +352,120 @@ impl Signer for EcdsaKey {
     }
 }
 
+pub struct RsaKey {
```
soundofspace, line 327: This one will also have to import/exported in `mod.rs`

When you run linting/clippy/tests
```
just qa
```
or only linting/clippy without tests
```
just qq
```
It should also detect these things, same for unused imports

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +352,120 @@ impl Signer for EcdsaKey {
...
+        alg: JWA,
+        rng: SystemRandom,
+    ) -> Result<Self, OpaqueError> {
+        let key_pair = RsaKeyPair::from_pkcs8(pkcs8_der).context("create RSAKeyPair from pkcs8")?;
+
+        Self::new(key_pair, alg, rng)
+    }
+
+    /// Create pkcs8 der for the current [`RsaKeyPair`]
+    pub fn pkcs8_der(&self) -> Result<(JWA, Pkcs8V1Der<'_>), OpaqueError> {
```
soundofspace:
```suggestion
    pub fn pkcs8_der(&self) -> Result<(JWA, Pkcs8V1Der<'static>), OpaqueError> {
```
'static is more explicit here and its also what happens internally

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +352,120 @@ impl Signer for EcdsaKey {
...
+        protected_headers.try_set_headers(SigningHeaders {
+            alg: jwk.alg,
+            jwk: &jwk,
+        })?;
+        Ok(())
+    }
+
+    fn sign(&self, data: &str) -> Result<Self::Signature, Self::Error> {
+        let mut sig: Vec<u8> = Vec::new();
+        let alg = <&'static dyn RsaEncoding as TryFrom<JWA>>::try_from(self.alg)?;
```
soundofspace:
```suggestion
        let alg = self.alg.try_into()?;
```
Should be possible to just do this?

rama-crypto/src/jose/jwk.rs
```diff
@@ -284,6 +352,120 @@ impl Signer for EcdsaKey {
...
+                len_bytes.remove(0);
+            }
+            let first_byte = 0x80 | len_bytes.len() as u8;
+            let mut result = vec![first_byte];
+            result.extend_from_slice(&len_bytes);
+            result
+        }
+    }
+
+    pub(super) fn encode_integer(mut bytes: Vec<u8>) -> Vec<u8> {
```
soundofspace: Can you add comment that this function should only be used for parsing JWK encoded rsa values. This function does not work for general asn1 encoding which is fine for this use case, but a comment should prevent other people from later extending this module and using this function for wrong use cases.

Some things that this function does not handle:
- empty bytes vec
- negative numbers
- assumes input is already in minimal format, which is not always the case in general asn 1

Again fine for this use case, but warrants a comment

soundofspace (changes requested):
Nice progress, some minor remarks but looks like it's almost ready

rama-crypto/src/jose/jwk.rs
valkrypton: Could you suggest me some resources where I could learn more about these specs?

valkrypton:
@soundofspace could you review this?

soundofspace:
@valkrypton I waited with reviewing since tests were failing. And seems there is a merge conflict now also.

Can you fix both, I will review after? Let me know if you need help with the merge conflict.

You can run the same tests ci-cd does by running `just qa` locally, if that command succeeds ci-cd should also succeed

valkrypton:
Fixed @soundofspace

rama-crypto/src/jose/jwk.rs
soundofspace: Generally reading RFC's (but this is very involved), or checking how other projects do it, doesn't even have to be in the same language.

soundofspace:
LGTM for me, thanks for this! @GlenDC do you still want to review?

GlenDC:
> LGTM for me, thanks for this! @GlenDC do you still want to review?

Yes I'll review tonight and merge it if I'm also okay with it.
Can you @soundofspace please resolve the threads with a comment if they are resolved as there are still open (review/comment) threads in this PR.

soundofspace:
> > LGTM for me, thanks for this! @GlenDC do you still want to review?
> 
> Yes I'll review tonight and merge it if I'm also okay with it. Can you @soundofspace please resolve the threads with a comment if they are resolved as there are still open (review/comment) threads in this PR.

Resolved them, thought I resolved all of them, but seems like Github didn't load all of them

rama-crypto/src/jose/jwk.rs
```diff
@@ -181,6 +217,70 @@ impl JWK {
...
+        bit_string
+    }
+
+    fn create_final_sequence(algorithm_identifier: [u8; 15], bit_string: Vec<u8>) -> Vec<u8> {
+        let mut final_sequence = Vec::with_capacity(algorithm_identifier.len() + bit_string.len());
+        final_sequence.extend(algorithm_identifier);
+        final_sequence.extend(bit_string);
+        let final_sequence_len = utils::encode_der_length(final_sequence.len());
+        let mut result = Vec::with_capacity(1 + final_sequence_len.len() + final_sequence.len());
+        result.push(0x30);
```
GlenDC: @valkrypton @soundofspace can you please create constants for all these magic constants. They seem to indicate a type, I suppose related to ASN.1 tags or something like it? Please put these constants somewhere and use the constant here.

The constants should:

* have clear names
* have a proper description (doc strings)
* also have official RFC spec links where more can be learned about the tag

This is very important as it is very easy to make mistakes like this. And more so, any time I see this kind of code in future I would have to again play Sherlock Holmes and find out what the heck these numbers mean... Please be kind to future us.

rama-crypto/src/jose/jwk.rs
```diff
@@ -157,7 +160,40 @@ impl JWK {
...
+
+                let der_rsa_key = Self::create_der_rsa_key(rsa_key_len, rsa_public_key_sequence);
+
+                let bit_string_payload = Self::create_bit_string_payload(der_rsa_key);
+                let bit_string_len = utils::encode_der_length(bit_string_payload.len());
+
+                let bit_string = Self::create_bit_string(bit_string_len, bit_string_payload);
+
+                let algorithm_identifier = [
+                    0x30, 0x0d, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x01,
```
GlenDC: Create a constant for this, which will be a constant array from constant bytes, as each of these bytes have a meaning resulting in your identifier... The way it is now it is completely unreadable.... We are not robots, please be kind to humans.

rama-crypto/src/jose/jwk.rs
```diff
@@ -181,6 +217,70 @@ impl JWK {
...
+
+            r#use: Some(JWKUse::Signature),
+            key_ops: None,
+            x5c: None,
+            x5t: None,
+            x5t_sha256: None,
+        })
+    }
+
+    fn create_public_key_sequence(n_der_encoded: Vec<u8>, e_der_encoded: Vec<u8>) -> Vec<u8> {
```
GlenDC: Can you please include at the very least in the module docstring links to the RFC's which are used to implement this logic? As I assume you didn't just invent the exact algorithms used here to construct all these

rama-crypto/src/jose/jwk.rs
```diff
@@ -157,7 +160,40 @@ impl JWK {
...
+                let n_bytes = BASE64_URL_SAFE_NO_PAD
+                    .decode(n)
+                    .context("decode RSA modulus (n)")?;
+                let e_bytes = BASE64_URL_SAFE_NO_PAD
+                    .decode(e)
+                    .context("decode RSA exponent (e)")?;
+                let n_der_encoded = utils::encode_integer(n_bytes);
+                let e_der_encoded = utils::encode_integer(e_bytes);
+
+                let rsa_public_key_sequence =
```
GlenDC: @soundofspace @valkrypton you are allocating new Vec's over and over again in these kind of functions, while you just re-write them in a new vec each time... Is there no way to just allocate a single vec in which you write everything??? As in the end you also now have a result vec as output from `final_sequence`, seems like you might just be able to write everything in such a final vec buffer to begin with? I imagine you would even be able to up-front allocate the entire buffer at once as you know how big it is?

rama-crypto/src/jose/jwk.rs
```diff
@@ -286,6 +386,123 @@ impl Signer for EcdsaKey {
...
+        self.inner
+            .sign(self.alg.try_into()?, self.rng(), data.as_bytes(), &mut sig)
+            .context("sign protected data")?;
+        Ok(sig)
+    }
+}
+
+mod utils {
+    pub(super) fn encode_der_length(len: usize) -> Vec<u8> {
+        if len < 128 {
```
GlenDC: please provide explanation for these algorithms. Feel free to copy-paste the relevant docs (and shorten/summarise where applicable) from the RFC's used to implement with "see more links" to the RFC's itself... Makes it easer to verify the correctness of these algorithm implementations

rama-crypto/src/jose/jwk.rs
```diff
@@ -286,6 +386,123 @@ impl Signer for EcdsaKey {
...
+            vec![len as u8]
+        } else {
+            let mut len_bytes = len.to_be_bytes().to_vec();
+            while len_bytes[0] == 0 {
```
GlenDC: this is a very expensive operation, is there a way to make this a bit smarter? Why are there 0's in the first place, why not move it all at once? etc,..

rama-crypto/src/jose/jwk.rs
```diff
@@ -286,6 +386,123 @@ impl Signer for EcdsaKey {
...
+            result
+        }
+    }
+
+    /// This function should only be used for parsing JWK encoded RSA values.
+    /// The function should **not** be used for general ASN1 encoded values.
+    /// The function assumes the input is in minimal form, not empty and is a
+    /// positive integer.
+    pub(super) fn encode_integer(bytes: Vec<u8>) -> Vec<u8> {
+        let needs_leading_zero = bytes[0] & 0x80 != 0;
```
GlenDC: try to work with buffers please, you are allocating all over the place

rama-crypto/src/jose/jwk.rs
```diff
@@ -286,6 +386,123 @@ impl Signer for EcdsaKey {
...
+        if needs_leading_zero {
+            result.push(0);
+        }
+        result.extend(bytes);
+        result
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
GlenDC, line 409: I imagine the rfc's which you used to implement provide some more examples of byte sequences and their meaning, might be good to use them here as tests to check that you encode and decode it all correctly

GlenDC (changes requested):
Looks pretty good, but I believe you can do better. Please address or ask for more feedback in case I'm unclear. Of course where I'm wrong do correct me.

GlenDC:
In case there is something nog clear @valkrypton , do ask @soundofspace or myself for more explanation.
Issues like these come with mentoring so do make use of it where you need and want it.

valkrypton:
> In case there is something nog clear @valkrypton , do ask @soundofspace or myself for more explanation. Issues like these come with mentoring so do make use of it where you need and want it.

Yes I will definitely ask for help if i need it. I've been busy with my day job. I'll take a look at this over the weekend

rama-crypto/src/jose/jwk.rs
valkrypton: I am doing this to reduce `len` to it's minimal form, since it's type is `usize`. So a value of say 235 can be represented with the first byte only. So, this removes the zero bytes of the `usize` and sends only minimum necessary bytes.

rama-crypto/src/jose/jwk.rs
valkrypton: I could not find any example byte sequence in any of the RFCs I used. However, I generated a byte sequence of the JWK in this [rfc](https://datatracker.ietf.org/doc/html/rfc7517). I used python's `cryptography` library to do this. However, this seems a bit hacky to me, so if you have any suggestions to better this I'd be happy to do it.

valkrypton:
@GlenDC could you take a look at the new changes?

rama-crypto/src/jose/constants.rs
```diff
@@ -0,0 +1,52 @@
...
+pub(crate) use rsa_algorithm_identifier::RSA_ALGORITHM_IDENTIFIER;
+mod der_encoding_tags {
+    /// Identifier tag for a DER encoded integer.
+    /// Defined in [ITU X.680](https://www.itu.int/rec/T-REC-X.690/).
+    pub(crate) const DER_TAG_INTEGER: u8 = 0x02;
+    /// Identifier tag for a DER encoded bit string.
+    /// Defined in [ITU X.680](https://www.itu.int/rec/T-REC-X.690/).
+    pub(crate) const DER_TAG_BIT_STRING: u8 = 0x03;
+    /// Identifier tag for a DER encoded sequence.
+    /// Defined in [ITU X.680](https://www.itu.int/rec/T-REC-X.690/).
```
GlenDC: you link to x690 but talk about x680, which is it? And what is the last integer in that version?
https://www.itu.int/rec/T-REC-X.680/en is the link for 680 but what about 681 or 682, are those just patch versions?

Your value is correct FWIW but just wonder what you shoud link to here and why you talk about one thing but link to another

rama-crypto/src/jose/constants.rs
```diff
@@ -0,0 +1,52 @@
...
+    /// Maximum length of a DER encoded length in short form.
+    /// Defined in [ITU X.690](https://www.itu.int/rec/T-REC-X.690/).
+    pub(crate) const DER_LENGTH_SHORT_FORM_MAX: usize = 127;
+    /// Octet that indicates that no unused bits are present in a bit string.
+    /// Defined in section 8.6 of [ITU X.690](https://www.itu.int/rec/T-REC-X.690/).
+    pub(crate) const BIT_STRING_NO_UNUSED_BITS: u8 = 0x00;
+}
+
+/// Byte representation of RSA encryption algorithm identifier.
+/// See appendix C of [RFC 8017](https://datatracker.ietf.org/doc/rfc8017/)
```
GlenDC: Seems like an odd resource to link. Yes some if it is vaguely appears in these RFCs... but not all and even the ones that are are just by sheer luck of an example... Is there no authorive table for these???? I feel like I have to be sherlock holmes to try to find it. How do you find this info while implementing it????

rama-crypto/src/jose/constants.rs
GlenDC: In general though these itu docs while authoritive, are not very useful.... are there no easy tables? E.g. with tls cryptographic stuff there are usually quite good authoritive sources where you can find tables of these values...

rama-crypto/src/jose/constants.rs
```diff
@@ -0,0 +1,52 @@
...
+mod rsa_algorithm_identifier {
+    const SEQUENCE_TAG: u8 = 0x30;
+    const LENGTH: u8 = 0x0d;
+    const OBJECT_IDENTIFIER_TAG: u8 = 0x06;
+    const LENGTH_OID: u8 = 0x09;
+    const NULL_TAG: u8 = 0x05;
+    const LENGTH_NULL: u8 = 0x00;
+
+    pub(crate) const RSA_ALGORITHM_IDENTIFIER: [u8; 15] = [
```
GlenDC, line 50: source??

rama-crypto/src/jose/jwk.rs
GlenDC: If that's best you can find at least document that process in comments so it can be produced

valkrypton: I have already mentioned in code comments how i produced the byte sequence

GlenDC: @soundofspace in case this is an important byte sequence it would be good if you can help find one or more good sources for this more than just how to reproduce (which I do also like to keep in, so good job on that).

rama-crypto/src/jose/constants.rs
valkrypton: The link to X.690 also has the link to X.680, that's why linked it with X.680 as well. Are you talking about a table like this? 
https://www.oss.com/asn1/resources/asn1-made-simple/asn1-quick-reference/asn1-tags.html

GlenDC: yes except that oss.com is a pretty random source. Would prefer something more authoritive,
for example for most TLS constants I can easily find them in sources like https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml

Even better in rFC's but as these are often "extensions" not standarized in rfcs I think a source like IANA is good enough.

worst worst case I refer to constants as defined in source code of an authoritive implementation. E.g. for web implementations something like an open source browser engine or curl can be acceptable. But that's worst case.

valkrypton: I could only find random resources like the one i linked. Even iana search is just giving me links to rfcs. even this [crate](https://docs.rs/der/latest/der/index.html) mentions the same doc i mentioned

soundofspace: Tags are defined in the original x680 one 
https://www.itu.int/ITU-T/studygroups/com17/languages/X.680-0207.pdf
x680 focusses on ASN notation and the basic syntax

x690 then focussen on encoding rules such as BER, CER, DER
https://www.itu.int/ITU-T/studygroups/com17/languages/X.690-0207.pdf

Please update references to point to these sources and this part should be good

rama-crypto/src/jose/constants.rs
soundofspace: This entire thing has 3 layers, and the final goal is to get the der algorithm identifier for rsa encoding
1. The identifier oid: 1.2.840.113549.1.1.1 defined in RFC 8017
2. RFC 3279 (Algorithms and Identifiers for the Internet X.509 PKI) specifies tag is needs to be NULL. The general structure is IDENTIFIER, PARAMETER, but for rsa here we dont have PARAMETER so NULL instead
3. We need to combine everything in DER encode 
3.1 (IDENTIFIER, PARAMETER) is a sequence, so SEQUENCE_TAG, LENGHT of what follows
3.2 OBJECT_IDENTIFIER_TAG to specify OID, LENGTH of OID,and actual oid
3.3 NULL_TAG and LENGTH_NULL to encode null value

rama-crypto/src/jose/jwk.rs
soundofspace: I plan to add more examples to the entire jose module down the line, but it's true that the rfc doesn't really have examples so will add some of my own then, until then i think this example suffices

soundofspace (changes requested):
Some changes also from my end on the docs part, but otherwise this looks good to merge after that

valkrypton:
@GlenDC @soundofspace I've updated the docs, please have a look

soundofspace:
@valkrypton ci/cd is failing can you fix it and run `just qa` locally to test if everything is working

valkrypton:
`just qa` runs successfully for me

GlenDC:
> `just qa` runs successfully for me

rebase main branch in your branch, you’ll see the errors once you did that.

We no longer allow usage of expext/unwrap unless you explicitly allow it. Most of the times however you want to actually avoid it

valkrypton:
rebased and pushed

rama-crypto/src/jose/constants.rs
```diff
@@ -0,0 +1,59 @@
...
+///
+mod rsa_algorithm_identifier {
+    const SEQUENCE_TAG: u8 = 0x30;
+    const LENGTH: u8 = 0x0d;
+    const OBJECT_IDENTIFIER_TAG: u8 = 0x06;
+    const LENGTH_OID: u8 = 0x09;
+    const NULL_TAG: u8 = 0x05;
+    const LENGTH_NULL: u8 = 0x00;
+
+    pub(crate) const RSA_ALGORITHM_IDENTIFIER: [u8; 15] = [
```
GlenDC, line 50:
```suggestion
    /// This entire thing has 3 layers, and the final goal is to
    /// get the der algorithm identifier for rsa encoding
    ///
    /// 1. The identifier oid: 1.2.840.113549.1.1.1 defined in RFC 8017
    /// 2. RFC 3279 (Algorithms and Identifiers for the Internet X.509 PKI)
    ///    specifies tag is needs to be NULL. The general structure is
    ///    IDENTIFIER, PARAMETER, but for rsa here we dont have PARAMETER
    ///    so NULL instead
    /// 3. We need to combine everything in DER encode
    ///   - 3.1 (IDENTIFIER, PARAMETER) is a sequence, so SEQUENCE_TAG, LENGHT of what follows
    ///   - 3.2 OBJECT_IDENTIFIER_TAG to specify OID, LENGTH of OID,and actual oid
    ///   - 3.3 NULL_TAG and LENGTH_NULL to encode null value
    pub(crate) const RSA_ALGORITHM_IDENTIFIER: [u8; 15] = [
```

PR #651, opened 2025-08-01
Migrate to Fn from FnOnce + Clone

soundofspace:


examples/http_conn_state.rs
```diff
@@ -137,7 +137,9 @@ async fn main() {
             .expect("bind TCP Listener")
             .serve_graceful(
                 guard,
-                MapStateLayer::new(|app: AppState| {
+                MapStateLayer::new(move |app: AppState| {
+                    let alive = alive.clone();
```
GlenDC, line 141: fwiw I think you could also have done it inline (`alive: alive.clone()`) but I guess there is also clarity in having on top, so all good for me (this is just a comment, you can close as-is if this is on purpose)

GlenDC (approved):
Awesome, turned out to be a lot smaller than I imagined it to be. Great work, as always.

soundofspace:
> Awesome, turned out to be a lot smaller than I imagined it to be. Great work, as always.

Thanks, and yea was expecting this one to be a LOT bigger

examples/http_conn_state.rs
soundofspace: Yea was what i did first, but kinda liked the pattern of always doing the cloning first thing

PR #652, opened 2025-08-01
Save headers from CONNECT response in HttpProxyConnectResponseHeaders extension.

maartendeprez:
This can be used, for example, to copy headers from the CONNECT response to proxied responses.

### DIFF
--- a/rama-http-backend/src/client/proxy/layer/proxy_connector/mod.rs
+++ b/rama-http-backend/src/client/proxy/layer/proxy_connector/mod.rs
@@ -12,4 +12,4 @@ pub use layer::HttpProxyConnectorLayer;
 
 mod service;
 #[doc(inline)]
-pub use service::{HttpProxyConnector, HttpProxyConnectResponseHeaders};
+pub use service::{HttpProxyConnectResponseHeaders, HttpProxyConnector};


rama-http-backend/src/client/proxy/layer/proxy_connector/connector.rs
```diff
@@ -103,9 +104,12 @@ impl InnerHttpProxyConnector {
             }
         };
 
+        let headers = response.headers().clone();
```
GlenDC: no need to clone, you should be able to change `upgrade::on(response)` to `upgrade::on(&mut response)` and then later do `into_parts` to move the headers out of the response, this makes it something free to do, because if we would have to clone this will have to be opt-in given this use case is probably very specialized

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -262,3 +265,21 @@ where
         })
     }
 }
+
+#[derive(Clone)]
+pub struct HttpProxyHeaders {
```
GlenDC: I'm fine with this, but please:

- add docs
- also explain some use cases why this can be useful (here in "extra" docs of struct, but also in PR description, please always use cases especially in PR description / motivation)
- also impl `Debug` always

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -262,3 +265,21 @@ where
...
+    headers: Arc<HeaderMap>,
+}
+
+impl HttpProxyHeaders {
```
GlenDC: I think I would just implement this is as

```
struct HttpProxyConnectResponseHeaders(Arc<HeaderMap>)
```

and then impl `Deref => target = HeaderMap` as well as `AsRef`

GlenDC (changes requested):
Looks nice to me, please do address comments, once that is done we can merge this swiftly, def. do not forget about motivations

rama-http-backend/src/client/proxy/layer/proxy_connector/connector.rs
maartendeprez: Nice! I had tried using `into_parts`, but missed that `upgrade::on` can work on a mutable reference.

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
maartendeprez: Done!

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -262,3 +265,38 @@ where
...
+    }
+}
+
+impl AsRef<HeaderMap> for HttpProxyConnectResponseHeaders {
+    fn as_ref(&self) -> &HeaderMap {
+        &self.0
+    }
+}
+
+impl AsMut<HeaderMap> for HttpProxyConnectResponseHeaders {
```
GlenDC: I would drop the `AsMut` and `DerefMut`.

in Rust "as" and "deref" are really meant to be zero-cost operations.

Arc's function is called "make_mut" (not as_mut) because it will clone on write in case there are other references. Also there is probably no reason to anyway need to mutate this headermap directly as there is nothing really useful you can do with it. If you ever want to use it in your own `Response` (bad idea anyway) you would need to anyway clone it out of an arc (first). If you do that you can mutate without having to do it here.

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
maartendeprez: Added documentation.

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
maartendeprez: Removed the impls

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -267,6 +267,11 @@ where
 }
 
 #[derive(Clone, Debug)]
+/// Extension added to the [`Context`] by [`HttpProxyConnector`] to record the
+/// headers from a successful CONNECT response.
+///
+/// This can be used, for example, to copy headers from the CONNECT response to
```
GlenDC: It’s clear that you might want to extract or derive information from headers. The real question is, what are some practical use cases for this? Sharing examples could inspire others to look for this information, even if it's just for some of their own use cases. No need for sensitive details here—just general ideas would help others realize its value.

maartendeprez: Sorry for the delay; i had other obligations to tend to... Something like this? Or would you like a longer explanation?

GlenDC (changes requested):
LGTM @maartendeprez ; once CI passes it can be merged

GlenDC (approved):

PR #656, opened 2025-08-05
correctly handle insecure request over http proxies

soundofspace:


rama-http-backend/src/client/svc.rs
```diff
@@ -326,3 +331,170 @@ fn sanitize_client_req_header<S, B>(
...
+    ctx: &Context<State>,
+    req: &Request<Body>,
+) -> bool {
+    let is_http_proxy = ctx
+        .get::<ProxyAddress>()
+        .and_then(|proxy| proxy.protocol.as_ref())
+        .map(|protocol| protocol.is_http())
+        .unwrap_or_default();
+
+    let is_insecure_request = req
```
GlenDC: what if the scheme is not there? Not better to use `RequestContext` so we can check it using the `protocol` which also checks the port as a fallback and in future possible other things?

GlenDC (changes requested):
Looks good to me, just 1 remark

rama-http-backend/src/client/svc.rs
```diff
@@ -166,16 +166,29 @@ fn sanitize_client_req_header<S, B>(
         return Err(OpaqueError::from_display("missing host in CONNECT request").into());
     }
 
+    let is_http_proxy = ctx
```
GlenDC:
```suggestion
    let uses_http_proxy = ctx
```

rama-http-backend/src/client/svc.rs
```diff
@@ -166,16 +166,29 @@ fn sanitize_client_req_header<S, B>(
...
+        .and_then(|proxy| proxy.protocol.as_ref())
+        .map(|protocol| protocol.is_http())
+        .unwrap_or_default();
+
     let request_ctx = ctx
         .get_or_try_insert_with_ctx::<RequestContext, _>(|ctx| (ctx, &req).try_into())
         .context("fetch request context")?;
 
+    let is_insecure_request_over_http_proxy =
+        (request_ctx.protocol == Protocol::HTTP) && is_http_proxy;
```
GlenDC: I would use `!request_ctx.protocol.is_secure()` as to also make sure to catch `ws`. Another option is to manually match on protocols if you do not want to risk weird edge cases with `Custom` catch-all

PR #659, opened 2025-08-06
Issue 498/support auto proxy connector http+https+socks5+socks5h

soundofspace:
- Adds socks5(h) support to EasyWebClient
  - Add proxy connector that supports both http and socks5 proxies and delegates to a specific connector depending on proxy protocol
  - Kept this connector specific (read: not generic) to our HttpProxyConnector/Socks5ProxyConnector since this is used for EasyWebClientBuilder specifically and saw no real value in making this generic. For special use cases we already support bringing a custom proxy connector
  - Another option was also. Currently you can configure existing ProxyConnectors in `optional` mode which forwards traffic to inner one if proxy is not configured. I could have modified this to also forward traffic to inner svc if Proxy protocol was wrong instead of returning error like we do now, but this felt ugly and error prone in the long run
  - Right now I focussed on http(s) + socks5(h) only, not a generic amount of proxy connectors. Once we add one more I have some "scalable" ideas for this but didn't want to overcomplicate the current setup for something we might not support for quite some time
- Move EasyWebClient to rama crate as now its not specific to http anymore
- Move autobahn example to rama examples so it can use EasyWebClient


tests/integration/cli/cli_tests/http_echo.rs
```diff
@@ -1,5 +1,5 @@
 use rama::Context;
-use rama_http_backend::client::EasyHttpWebClient;
+use rama::http::client::EasyHttpWebClient;
 use rama_ws::handshake::client::HttpClientWebSocketExt;
```
GlenDC: this should also use `rama::`, not on you, but stupid IDE's :P If you do bump into it do fix it please

src/http/client/proxy_connector.rs
```diff
@@ -0,0 +1,194 @@
...
+                let EstablishedClientConnection { ctx, req, conn } =
+                    self.inner.connect(ctx, req).await.map_err(Into::into)?;
+                Ok(EstablishedClientConnection {
+                    ctx,
+                    req,
+                    conn: Either3::A(conn),
+                })
+            }
+            Some(proto) => match *proto {
+                Protocol::SOCKS5 | Protocol::SOCKS5H => {
```
GlenDC: Why not use something like `proto.is_http() { ... } else if proto.is_socks5() { ... } else { ... }` here? You are anyway doing a catch-all in the end.

src/http/client/proxy_connector.rs
```diff
@@ -0,0 +1,194 @@
...
+    stream::Stream,
+    transport::TryRefIntoTransportContext,
+};
+use rama_socks5::{Socks5ProxyConnector, Socks5ProxyConnectorLayer};
+
+/// Proxy connector which supports http(s) and socks5(h) proxy address
+///
+/// Connector will look at [`ProxyAddress`] to determine which proxy
+/// connector to use if one is configured
+pub struct ProxyConnector<S> {
```
GlenDC, line 21: Would have called this `AutoProxyConnector`. But I guess the auto doesn't have that much value here

src/http/client/proxy_connector.rs
```diff
@@ -0,0 +1,194 @@
+use std::sync::Arc;
+
+use rama_core::{
```
GlenDC: please prefer to use `crate::` imports here for everything. I know that in the cli modules I have some direct imports as well by accident, that is once again fault of the ide... It makes it all a lot easier to manager over time if we use the single export location

src/http/client/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    async fn serve(
+        &self,
+        ctx: Context<State>,
+        req: Request<Body>,
+    ) -> Result<Self::Response, Self::Error> {
+        let uri = req.uri().clone();
+
+        let EstablishedClientConnection { ctx, req, conn } = self.connector.serve(ctx, req).await?;
+        // NOTE: stack might change request version based on connector data,
```
GlenDC, line 156: Why add that note here? Is there anything that needs to be done for it?

src/http/client/mod.rs
```diff
@@ -0,0 +1,169 @@
...
+    Context, Service,
+    error::{BoxError, ErrorContext, OpaqueError},
+    http::{Request, Response, dep::http_body},
+    net::client::EstablishedClientConnection,
+    service::BoxService,
+    telemetry::tracing,
+};
+
+#[doc(inline)]
+pub use ::rama_http_backend::client::*;
```
GlenDC, line 18: use crate imports only please

src/http/client/builder.rs
```diff
@@ -0,0 +1,589 @@
...
+    }
+
+    /// Add support for usage of a http(s) [`ProxyAddress`] to this client
+    ///
+    /// Note that a tls proxy is not needed to make a https connection
+    /// to the final target. It only has an influence on the initial connection
+    /// to the proxy itself
+    ///
+    /// [`ProxyAddress`]: rama_net::address::ProxyAddress
+    pub fn with_http_only_proxy_support(
```
GlenDC: Personally I would have named it:

```
- with_http_proxy_support
- with_socks5_proxy_support
- with_proxy_support
```

The only is clear by the fact that they can `http_proxy` and that they can no longer call `socks5_proxy` after this

src/http/client/builder.rs
```diff
@@ -0,0 +1,589 @@
...
+    /// Add support for usage of a http(s) [`ProxyAddress`] to this client
+    ///
+    /// Note that a tls proxy is not needed to make a https connection
+    /// to the final target. It only has an influence on the initial connection
+    /// to the proxy itself
+    ///
+    /// Note to also enable socks proxy support enable feature `socks5`
+    ///
+    /// [`ProxyAddress`]: rama_net::address::ProxyAddress
+    pub fn with_proxy_support(self) -> EasyHttpWebClientBuilder<HttpProxyConnector<T>, ProxyStage> {
```
GlenDC, line 260: I would only define this if you enable both `http` and `socks5`... otherwise this is very confusing. If that's to confusing name it `auto_proxy` or so but I would not have both `http` and `socks5` .

But perhaps you have good counter arguments

src/http/client/builder.rs
```diff
@@ -0,0 +1,589 @@
+use super::{HttpConnector, http_inspector::HttpVersionAdapter};
+use rama_core::{
+    Layer, Service,
+    error::{BoxError, OpaqueError},
+};
+use rama_dns::DnsResolver;
```
GlenDC: crate imports only please (or super) but not rama_core etc directly

src/http/client/mod.rs
soundofspace: No, that was a remark that was already there by you, removed this by accident a while ago and then you asked me te place it again.

I think the reason you mentioned then was: connector might change version (eg alpn) which is expected technically but might surprise people, hence why this remark is here

src/http/client/proxy_connector.rs
soundofspace: Yea started with that but it felt unneeded so removed it

src/http/client/mod.rs
soundofspace: Not possible for this one? here we re-export it so how would it already be available as crate import

src/http/client/builder.rs
```diff
@@ -0,0 +1,585 @@
...
+use crate::{http::client::proxy_connector::ProxyConnector, proxy::socks5::Socks5ProxyConnector};
+
+/// Builder that is designed to easily create a [`super::EasyHttpWebClient`] from most basic use cases
+#[derive(Default)]
+pub struct EasyHttpWebClientBuilder<C = (), S = ()> {
+    connector: C,
+    _phantom: PhantomData<S>,
+}
+
+pub struct TransportStage;
```
GlenDC, line 39: Btw something I only notice now. But you usually want to make these marker types both: `non_exhaustive` and `Debug`.

Especially the first is important as we do not really want people to be able to construct this. What matters only is that the type itself is public.

src/http/client/proxy_connector.rs
```diff
@@ -0,0 +1,193 @@
...
+
+/// Proxy connector which supports http(s) and socks5(h) proxy address
+///
+/// Connector will look at [`ProxyAddress`] to determine which proxy
+/// connector to use if one is configured
+pub struct ProxyConnector<S> {
+    inner: S,
+    socks: Socks5ProxyConnector<S>,
+    http: HttpProxyConnector<S>,
+    required: bool,
```
GlenDC, line 25: This does mean that you can make the parent 'non-required' but one of them 'still required'. I guess that's ok but it's funny to realize none the less. Feel free to close this without any action if this is fine in your eyes, as I can live with it as well.

tests/integration/examples/example_tests/proxy_connectivity_check.rs
```diff
@@ -47,11 +44,7 @@ async fn test_http_client_over_socks5_proxy_connect() {
         "local servers up and running",
     );
 
-    // TODO: once we have socks5 support in Easy http web client
-    // we can probably simplify this by using the interactive runner client
-    // instead of having to do this manually...
-
-    let client = HttpConnector::new(Socks5ProxyConnector::required(TcpConnector::new()));
+    let client = EasyHttpWebClient::default();
```
GlenDC: What I meant is that in all these socks5 tests we shouldn't even have a need to create our own Client, right? We should be able to just run `runner.get(...).send(...)` no? Just need to make sure that our default runner client is making use of the auto proxy support, which I guess it already is anyway?

GlenDC (approved):
LGTM, couple of last remarks but these are details that you can probably resolve without any controversy and merge it yourself afterwards if you have the confidence.

tests/integration/examples/example_tests/proxy_connectivity_check.rs
soundofspace: ah missed that intent, indeed much better that

src/http/client/proxy_connector.rs
soundofspace: yea thought about that aswel but didnt really find a better solution. My first solution was to integrate Socks5ProxyConnector directly into this connector, but i then moved to providing them as layers as that gives extra flexibility if a user wants to use this proxyConnector directly and provide these layers with custom options (eg socks dns), but yea its weird for the required part

PR #660, opened 2025-08-07
Fix broken tracing with multiple layers

soundofspace:
Tracing doesn't really support trace events inside Debug implementations. This debug impl has a trace event quite deep in the debug logic. Easy option is to remove that trace event in `SettingOrder::push` but to me this shows that this Debug impl was too complex, so I simplified it. It should still contain all the necessary info, just a bit more manual when you need to interpret it

### DIFF
--- a/rama-http-types/src/proto/h2/frame/settings.rs
+++ b/rama-http-types/src/proto/h2/frame/settings.rs
@@ -318,11 +318,6 @@ impl fmt::Debug for Settings {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut builder = f.debug_struct("Settings");
         builder.field("flags", &self.flags);
-        builder.field("setting_order", &self.config.setting_order);
-        builder.field(
-            "setting_order defaults (appended if not present)",
-            &SettingOrder::default_settings(),
-        );
         builder.field("config", &self.config);
         builder.finish()
     }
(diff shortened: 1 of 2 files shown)


rama-http-types/src/proto/h2/frame/settings.rs
```diff
@@ -319,37 +319,11 @@ impl fmt::Debug for Settings {
...
-            }
-            SettingId::NoRfc7540Priorities => {
-                builder.field("no_rfc7540_priorities", &setting.value);
-            }
-            SettingId::Unknown(id) => {
-                builder.field(&format!("unknown_unknown_setting_{id}"), &setting.value);
-            }
-        });
-
+        builder.field(
```
GlenDC: what is the goal of this defaults? Also spaces in a name (and brackets) is not something that we do anywhere else

rama-http-types/src/proto/h2/frame/settings.rs
```diff
@@ -319,37 +319,11 @@ impl fmt::Debug for Settings {
         let mut builder = f.debug_struct("Settings");
         builder.field("flags", &self.flags);
         builder.field("setting_order", &self.config.setting_order);
```
GlenDC: why still do this if you already expose config directly as a field?

GlenDC (changes requested):

GlenDC (approved):

PR #668, opened 2025-08-19
Default http proxy connector when no protocol is set on proxy

soundofspace:


src/http/client/proxy_connector.rs
```diff
@@ -108,25 +108,30 @@ where
...
                     let EstablishedClientConnection { ctx, req, conn } =
                         self.socks.connect(ctx, req).await?;
                     Ok(EstablishedClientConnection {
                         ctx,
                         req,
                         conn: Either3::B(conn),
                     })
-                } else if proto.is_http() {
+                } else {
+                    tracing::trace!("using http proxy connector");
```
GlenDC, line 131: this is not he logic you want

I think you want high level:

- protocol = protocol or (log that we default to http + use http)
- if protocol.is_http() => use http
- else if protocol.is_socks() => use socks
- else => error

GlenDC (changes requested):

src/http/client/proxy_connector.rs
soundofspace: agree

GlenDC (approved):

PR #670, opened 2025-08-20
Use MaybeProxiedConnection wrapper in proxy connector instead of heavy Either3

soundofspace:
This changes the generic to a single one instead of 3 different heavy branches.

### DIFF
--- a/src/http/client/proxy_connector.rs
+++ b/src/http/client/proxy_connector.rs
@@ -11,7 +11,13 @@ use crate::{
     proxy::socks5::{Socks5ProxyConnector, Socks5ProxyConnectorLayer},
 };
 use pin_project_lite::pin_project;
-use std::sync::Arc;
+use rama_core::combinators::Either;
+use std::{
+    fmt::Debug,
+    pin::Pin,
+    sync::Arc,
+    task::{self, Poll},
+};
 use tokio::io::{AsyncRead, AsyncWrite};
 
 /// Proxy connector which supports http(s) and socks5(h) proxy address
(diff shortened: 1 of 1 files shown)


src/http/client/proxy_connector.rs
```diff
@@ -133,6 +132,102 @@ where
     }
 }
 
+pin_project! {
+    #[non_exhaustive]
```
GlenDC: useless tag if your properties are private anyway

src/http/client/proxy_connector.rs
```diff
@@ -133,6 +132,102 @@ where
...
+    /// A connection which will be proxied if a [`ProxyAddress`] was configured
+    ///
+    /// This is wrapper around the actual connection to prevent the use of
```
GlenDC: don't think this comment needs to be here,  as it's not just for that. It triggered us to look at it, but it's also fact that this is a detail that shouldn't be exposed either

src/http/client/proxy_connector.rs
```diff
@@ -133,6 +132,102 @@ where
...
+        inner: Connection<S>,
+    }
+}
+
+pin_project! {
+    #[project = ConnectionProj]
+    enum Connection<S> {
+        Direct{ #[pin] conn: S },
+        Socks{ #[pin] conn: S },
+        Http{ #[pin] conn: rama_core::combinators::Either<S, rama_http::io::upgrade::Upgraded> },
```
GlenDC: you can just import this at top, bit ugly and not needed to have these here

src/http/client/proxy_connector.rs
```diff
@@ -133,6 +132,102 @@ where
...
+    ) -> std::task::Poll<Result<(), std::io::Error>> {
+        match self.project().inner.project() {
+            ConnectionProj::Direct { conn } | ConnectionProj::Socks { conn } => conn.poll_flush(cx),
+            ConnectionProj::Http { conn } => conn.poll_flush(cx),
+        }
+    }
+
+    fn poll_shutdown(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
```
GlenDC: it's a bit noisy with these full qualifiers. for stuff like Pin just import it. For stuff like context I for now usually do it as `task::Context` in case it otherwise conflicts with rama COntext. Once rama context is gone also that potential issue is gone

src/http/client/proxy_connector.rs
```diff
@@ -133,6 +132,102 @@ where
...
 
+pin_project! {
+    #[non_exhaustive]
+    /// A connection which will be proxied if a [`ProxyAddress`] was configured
+    ///
+    /// This is wrapper around the actual connection to prevent the use of
+    /// `Either3<NoProxyConnector, HttpProxyConnector, SocksProxyConnector>`
+    /// which is a really heavy generic if placed in the middle of a connector
+    /// stack.
+    pub struct MaybeProxiedConnection<S> {
```
GlenDC, line 143: do not forget about debug for these 2 types :)

GlenDC (approved):

Issue #658, opened 2025-08-06
add http-to-https "upgrade" Redirect (SS) service


GlenDC:
A very simplistic impl of this would be:

```rust
async |req: Request| {
    let uri = req.uri().clone().into_parts();
    let Some(authority) = uri.authority else {
        return Ok(StatusCode::BAD_GATEWAY.into_response());
    };

    let loc = match uri.path_and_query {
        Some(paq) => format!("https://{authority}?{paq}"),
        None => format!("https://{authority}"),
    };

    Ok(([(header::LOCATION, loc)], StatusCode::PERMANENT_REDIRECT).into_response())
}
```

However:

- this falsely assumes that the authority is always there in Uri which it is not, so we probably want to rely on `RequestContext` instead
- beyond that I think it's probably ok, except that we do not need to clone uri as we do not need ownership over it

All in all a simple service which can even be a `ZST`.

Do double check the relevant RFCs in case I miss something here.

GlenDC:
@soundofspace your opinion is welcome here as well :)

GlenDC:
This is the most recent version I use in some places for this:

```rust
async |mut ctx: Context<State>, req: Request| {
    let req_ctx: &mut RequestContext =
        match ctx.get_or_try_insert_with_ctx(|ctx| (ctx, &req).try_into()) {
            Ok(req_ctx) => req_ctx,
            Err(err) => {
                tracing::error!(
                    "failed to get request ctx for insecure incoming req: {err}"
                );
                return Ok(StatusCode::BAD_GATEWAY.into_response());
            }
        };

    let host = &req_ctx.authority.host();
    let paq = req
        .uri()
        .path_and_query()
        .map(|paq| paq.as_str())
        .unwrap_or("/");
    let loc = format!("https://{host}{paq}");

    Ok(([(header::LOCATION, loc)], StatusCode::PERMANENT_REDIRECT)
        .into_response())
}
```

lauk20:
Hi! I would like to give this a go as we discussed. 😄 

Would appreciate some code pointers/guidance to get started!

GlenDC:
> Hi! I would like to give this a go as we discussed. 😄 
> 
> Would appreciate some code pointers/guidance to get started!

All yours. Welcome to the community. Let me know if you need more explanation or guidance. Feel free to open a Draft PR as soon as you want, are stuck or need guidance.

PR #678, opened 2025-08-24
add http-to-https 'upgrade' Redirect (SS) service

lauk20:
Addressing #658 
add http-to-https 'upgrade' Redirect (SS) service.

### DIFF
--- a/rama-http/src/service/upgrade.rs
+++ b/rama-http/src/service/upgrade.rs
@@ -1,0 +1,60 @@
+//! Service that redirects all HTTP requests to HTTPS
+
+use crate::Request;
+use crate::{Response, header};
+use crate::service::web::response::IntoResponse;
+use crate::StatusCode;
+use rama_core::{Context, Service, telemetry::tracing};
+use rama_net::http::RequestContext; 
+use std::{convert::Infallible, fmt};
+
+/// Service that redirects all HTTP requests to HTTPS
+pub struct Upgrade;
+
+impl<State, Body> Service<State, Request<Body>> for Upgrade
+where
+    State: Clone + Send + Sync + 'static,
+    Body: Send + 'static,
+{
+    type Response = Response;
+    type Error = Infallible;
+
+    async fn serve(
+        &self,
+        mut ctx: Context<State>,
+        req: Request<Body>,
+    ) -> Result<Self::Response, Self::Error> {
+        let req_ctx: &mut RequestContext =
+            match ctx.get_or_try_insert_with_ctx(|ctx| (ctx, &req).try_into()) {
+                Ok(req_ctx) => req_ctx,
+                Err(err) => {
+                    tracing::error!(
+                        "failed to get RequestContext for insecure incoming req: {err}"
+                    );
+                    return Ok(StatusCode::BAD_GATEWAY.into_response());
+                }
+            };
+        let host = &req_ctx.authority.host();
+        let paq = req
+            .uri()
+            .path_and_query()
+            .map(|paq| paq.as_str())
+            .unwrap_or("/");
+        let loc = format!("https://{host}{paq}");
+
+        Ok(([(header::LOCATION, loc)], StatusCode::PERMANENT_REDIRECT).into_response())
+    }
+}
+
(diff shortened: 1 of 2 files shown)


rama-http/src/service/upgrade.rs
```diff
@@ -0,0 +1,60 @@
...
+use crate::Request;
+use crate::{Response, header};
+use crate::service::web::response::IntoResponse;
+use crate::StatusCode;
+use rama_core::{Context, Service, telemetry::tracing};
+use rama_net::http::RequestContext; 
+use std::{convert::Infallible, fmt};
+
+/// Service that redirects all HTTP requests to HTTPS
+pub struct Upgrade;
```
GlenDC:
```suggestion
#[derive(Debug, Clone, Default)]
#[non_exhaustive]
pub struct Upgrade;
```

- pretty much always you want debug and clone, at times needed to be done manually
- in this case default is also applicable
- non_exhaustive to make sure others cannot construct it except via constructors we provide (e.g. `default()`)

rama-http/src/service/upgrade.rs
```diff
@@ -0,0 +1,60 @@
...
+            match ctx.get_or_try_insert_with_ctx(|ctx| (ctx, &req).try_into()) {
+                Ok(req_ctx) => req_ctx,
+                Err(err) => {
+                    tracing::error!(
+                        "failed to get RequestContext for insecure incoming req: {err}"
+                    );
+                    return Ok(StatusCode::BAD_GATEWAY.into_response());
+                }
+            };
+        let host = &req_ctx.authority.host();
```
GlenDC, line 50: might want to also check check specifically on the protocol found in req ctx

- if ws => upgrade to wss
- if http => upgrade to https
- if something else return bad gateway as that's unexpected

rama-http/src/service/upgrade.rs
```diff
@@ -0,0 +1,60 @@
...
+        let paq = req
+            .uri()
+            .path_and_query()
+            .map(|paq| paq.as_str())
+            .unwrap_or("/");
+        let loc = format!("https://{host}{paq}");
+
+        Ok(([(header::LOCATION, loc)], StatusCode::PERMANENT_REDIRECT).into_response())
```
GlenDC: status_code is probably something we wish to allow to be customised. The default one can be `PERMANENT_REDIRECT`, that means the struct no longer needs a non_exhaustive as it will contain a `StatusCode`. You can make a setter (set/with) for it using the `generate_set_and_with!` macro (see codebase on how to import and use it)

rama-http/src/service/upgrade.rs
```diff
@@ -0,0 +1,60 @@
...
+    }
+}
+
+impl fmt::Debug for Upgrade {
```
GlenDC: no need to manually implement these for `Upgrade`. Only needed manually if you want to prevent some automatic behaviour (e.g. do not want to enforce the bounds on a generic where it isn't needed)

GlenDC (changes requested):
Thank you for this PR. Left some reviews. Once those are resolved we can probably think about moving this forward and merge. In case you do not understand some feedback or do not agree, do let me know.

lauk20:
> Thank you for this PR. Left some reviews. Once those are resolved we can probably think about moving this forward and merge. In case you do not understand some feedback or do not agree, do let me know.

Thanks for the feedback! I have updated the PR with the suggested changes. Let me know if any edits should be made.

GlenDC (approved):
Looks great. Once the CI passes it can be merged in. Thank you for this swift and great contribution!

GlenDC:
@lauk20 please run `just qq`, and ensure it all passes.

lauk20:
> @lauk20 please run `just qq`, and ensure it all passes.

@GlenDC Got it, updated PR with `just qq` changes 👍

GlenDC:
no need to run `just qq` btw if you already run `just qa`, as `qa` contains all that `qq` contains + `test` :) `qq` is just there for folks who want to have quick feedback as `test` takes a while on some machines.

GlenDC (approved):

Issue #462, opened 2025-03-17
revisit Service API (core) design in 0.3 milestone


GlenDC:
As rama is "maturing", we start to get more end user experience and also collect more feedback from other users. This didn't make the 0.2.x cut as it's an impactful change that will require some iteration to get right. But for now this is more or less how I see it.

```rust
pub trait Service<Request>: Sized + Send + Sync + 'static {
    type Response: Send + 'static;
    type Error: Send + Sync + 'static;

    fn serve(
        &self,
        req: Request,
    ) -> impl Future<Output = Result<Self::Response, Self::Error>> + Send + '_;

    fn boxed(self) -> BoxService<Request, Self::Response, Self::Error> {
        BoxService {
            inner: Box::new(self),
        }
    }
}
```

in contrast to <https://docs.rs/rama/latest/rama/trait.Service.html>. The change here is that `Context` would be removed, if we indeed go through with this. I want to be clear this isn't a certainty yet. And this issue is mostly to start the discussion and iteration of ideas on how we might want to tackle this and anything related in 0.3.

## So why would we want to get rid of `Context`, if at all?

This would make rama compatible with a future version of tower. And while rama might always have its own services, it might make it at least possible for us to make it possible to use tower services in rama and rama services in tower context is today used for three things.

### 1. State

This is the big one, but as rama starts to become more mature and I have more experience with it as an end user I start to think that we pay a high price for that static state. Sure it's nice, but it's also very limited in use. The real killer is to have it in web endpoint services, which is a very specific use case. And other case where I use it is to have static stuff for creating services downstream. For the first we might be able to make it still work but more the axum-approach where we perhaps tie that directly into the http server somehow and still allow it to be accessed as an extractor. Not sure how that would work though, would need some figuring out. The latter I am not convinced we really need static types for that. Might as well just put that stuff in extensions if needed or use a more functional approach by using global functions unique to the app to get access to this information. At the other hand the price for this state is really high even though in 80% of places we do not use it, you do always have to take it into account.

#### Why was state introduced?

[...]

PR #685, opened 2025-08-26
Issue 462/revisit service api (core) design in 0.3 milestone# removing state from context

soundofspace:
Part of #462, this PR removes State from Context. Where state is still needed it has been moved to `ctx.extensions` and in most cases been added with `AddExtensionLayer`.

Logic for with_state has also been removed in favor of `AddExtensionLayer`. Once we have a solution for fully typed (localized) state again we could probably add this again, but until then it has been removed so we can start with a clean slate.


docs/book/src/intro/context.md
```diff
@@ -1,56 +1,39 @@
...
 the `Extensions` type/data also available in an `http Request`. However it provides no means
-of typesafe `State`, executors, spawning etc. On top of that it would make it more awkward to
+of typesafe executors, spawning etc. On top of that it would make it more awkward to
 also freely pass all this data between services, especially those operating
 across different layers of the network.
 
 ## State
 
-`rama` supports two kinds of states:
+`rama` currently only support one type of state:
```
GlenDC: I do think it makes sense here that where one wants typed state they can already do so easily and require no support from rama for this. 

1. add as part of struct, good old regular property
2. closures for things like service functions

examples/unix_socket.rs
```diff
@@ -48,18 +49,15 @@ mod unix_example {
...
 
         graceful.spawn_task_fn(async |guard| {
-            async fn handle(
-                ctx: Context<ShutdownGuard>,
-                mut stream: impl Stream + Unpin,
-            ) -> Result<(), BoxError> {
+            async fn handle(ctx: Context, mut stream: impl Stream + Unpin) -> Result<(), BoxError> {
                 let mut buf = [0u8; 1024];
-                let guard = ctx.state();
+                let guard = ctx.get::<ShutdownGuard>().unwrap();
```
GlenDC, line 63: Can you add a Giant TODO Here that we want to clean this up in near future?

1. Executor is now part of context but will move into regular extensions
2. That has some kind of guard there should be an easy way to get a future from that somehow that either is when guard cancels or just something that never cancels (if desired)

that will make this code better and something that works in general, and not only because this example has to inject it themselves.

When you make a graceful server it will inject that kind of stuff in extensions, or so is the goal.

rama-core/src/context/mod.rs
```diff
@@ -2,25 +2,9 @@
 //!
 //! # State
 //!
-//! [`rama`] supports two kinds of states:
+//! [`rama`] currently supports one type of state:
```
GlenDC: same comment as in the regular .MD file

rama-core/src/layer/add_extension.rs
```diff
@@ -39,7 +37,7 @@
 //! };
 //!
 //! let mut service = (
-//!     // Share an `Arc<State>` with all requests.
+//!     // Share an `Arc` with all requests.
```
GlenDC:
```suggestion
//!     // Share an `Arc<State>` with all requests.
```

rama-http-core/src/h2/proto/connection.rs
```diff
@@ -557,7 +557,7 @@ where
                 tracing::trace!("recv GOAWAY: {frame:?}");
                 // This should prevent starting new streams,
                 // but should allow continuing to process current streams
-                // until they are all EOS. Once they are, State should
+                // until they are all EOS. Once they are should
```
GlenDC: revert please, this is different kind of state :P

rama-http/src/service/web/k8s.rs
```diff
@@ -4,26 +4,24 @@ use crate::{
     Request, Response, StatusCode, matcher::HttpMatcher,
     service::web::endpoint::response::IntoResponse,
 };
+use core::fmt;
```
GlenDC: revert this please, it was part of std, which is fine.

rama-unix/src/unix/server/listener.rs
```diff
@@ -24,62 +23,30 @@ use crate::UnixSocketAddress;
...
-}
+pub struct UnixListenerBuilder;
 
-impl UnixListenerBuilder<()> {
+impl UnixListenerBuilder {
     /// Create a new `UnixListenerBuilder` without a state.
     #[must_use]
     pub fn new() -> Self {
-        Self { state: () }
+        Self {}
```
GlenDC:
```suggestion
        Self
```

GlenDC (changes requested):
Looks good. Much cleaner and less generics. Love it.
Step 1 pretty much complete, once you resolved the open feedback we can queue this for merge. Will do by end of week, after the release of alpha.3; which I'll do between friday and sunday.

GlenDC:
0.3.0-alpha.3 was released, you can now finish up this PR

PR #646, opened 2025-07-27
basic har layer

MeerKatDev:
Deals with #357 

 - Missing example
 - Missing internal logic

PR #694, opened 2025-09-02
support HAR exporter (http) layer in rama

### LINKED ISSUE #357: support HAR exporter (http) layer in rama
By default it exports to a file of your choice,
but we can also support to provide custom writers I suppose :)

Most likely we need to do the actual writing over a channel as to ensure there's only ever one active writer to the file though, given a (rama) service is `Send+Sync`.

HAR File wikipedia: https://en.wikipedia.org/wiki/HAR_(file_format)

GlenDC:
Closes #357

Continued based on the great work by @MeerKatDev
in PR https://github.com/plabayo/rama/pull/646


GlenDC:
FYI @MeerKatDev the initial HAR support is now finished

PR #696, opened 2025-09-04
Issue 692/integrate / fork http crate into rama#partially

soundofspace:
Goal of this PR:
- Integrate http request/response modules into rama so we can use our Extension type
- Provide From/Into to convert between hyper and rama types
- Reorganize imports and exports in such a way that if we later decide to fork more modules/types all imports should remain the same

Not done in this PR:
- Reorganize our Extension module
- Add new logic / utilities (except from) so only focus is integrating


examples/http_mitm_proxy_boring.rs
```diff
@@ -115,6 +115,7 @@ use rama::{
 };
 
 use itertools::Itertools;
+use rama_http::Body;
```
GlenDC: move up

rama-http-core/src/proto/h1/mod.rs
```diff
@@ -69,7 +68,7 @@ pub(crate) struct ParseContext<'a> {
     h1_max_headers: Option<usize>,
     h09_responses: bool,
     on_informational: &'a mut Option<crate::ext::OnInformational>,
-    encoded_request_extensions: &'a mut Option<http::Extensions>,
+    encoded_request_extensions: &'a mut Option<rama_core::context::Extensions>,
```
GlenDC: you can just import extensions, no?

rama-http-core/src/proto/h1/role.rs
```diff
@@ -275,7 +274,7 @@ impl Http1Transaction for Server {
             return Err(Parse::transfer_encoding_invalid());
         }
 
-        let mut extensions = http::Extensions::default();
+        let mut extensions = rama_core::context::Extensions::default();
```
GlenDC: might as well import Extensions as there should be no more conflict anyway?

rama-http-types/src/extensions.rs
```diff
@@ -0,0 +1,27 @@
+#[derive(Default, Clone)]
+pub struct RamaExtensions(pub rama_core::context::Extensions);
```
GlenDC: Why do we make these two wrappers? Can we not live without??? Seems silly?

1. we can just use core extensions directly, no?
2. we shouldn't need that ext wrapper and can instead just create convertors directly?

You won't be able to create `From` due to orphan rule, but don't think we need it, probably just can do this directly within the request/response convertor types. There you can add inner utility for avoiding duplication, but no need to ugly wrap I would thing... seems stupid and not something I think we need to publicly expose / polute for?

rama-http-types/src/lib.rs
```diff
@@ -133,13 +91,66 @@ pub mod header {
...
-#[doc(inline)]
-pub use self::dep::http::uri::{Scheme, Uri};
-#[doc(inline)]
-pub use self::dep::http::version::Version;
+pub mod dep {
+    //! Dependencies for rama http modules.
+    //!
+    //! Exported for your convenience.
+
+    pub mod hyperium {
```
GlenDC: should we still expose this if we already have these crates and their types all exposed nicely without dep?

rama-http/src/lib.rs
```diff
@@ -47,5 +48,5 @@ pub mod dep {
     pub use rama_core as core;
 
     #[doc(inline)]
-    pub use ::rama_http_types::dep::{http, http_body, http_body_util, mime, mime_guess};
+    pub use ::rama_http_types::dep::{hyperium, mime, mime_guess};
```
GlenDC: again, not sure we still need/want the hyperium dep here?

tests/http-core/support/mod.rs
```diff
@@ -476,7 +476,7 @@ async fn async_test(cfg: __TestConfig) {
                 func(res.headers());
             }
 
-            let body = res.collect().await.unwrap().to_bytes();
+            let body = res.into_body().collect().await.unwrap().to_bytes();
```
GlenDC: why the need to turn into body now first??? or you just prefer to do it explicitly even if not needed

rama-http-types/src/extensions.rs
soundofspace: Yea migrated. The plan here was always to hopefully add a new api to hyper http crate so we could transfer extensions directly and this was only a temp solution.
(And also me fighting the orphan rule again...)

rama-http-types/src/lib.rs
soundofspace: I kept them exposed for use cases such as otel where you still need access to hyper original types but want the same version rama uses. This will become more important also if we integrate more types I think

GlenDC: I think that's a detail though. For example if you write your own otel exporter you can just as well import hyper's http directly, just like you also import http.

I don't think within our API we ever expose our require user to need to use those types.
So unless you feel strongly about it I would remove it and keep it a detail for us. Where all types are ours (re-exported as you already do) and if people need access to original types they can import hyper's http crates directly. THey are stable enough that it should be ok.

We can always revisit this if it causes issues, but I think it should be ok, wdyt?

tests/http-core/support/mod.rs
soundofspace: Forgot to migrate two implementations (thought it was a type issues, but was just something I forgot)

rama-http-types/src/lib.rs
soundofspace: Removed the re-export but kept pub(crate) since I quite like this import structure in this crate to clearly see what comes from where

Issue #566, opened 2025-05-23
add rama-pac: initial implementaiton with `PacConnector` as the main API support


GlenDC:
PAC (Proxy Auto Configuration) is legacy but still used. It is dynamic approach to decide what traffic to proxy how, with fallback support etc. The dynamic part is achieved by running a small javascript file (the PAC file) within an embedded js runtime which outputs a line such as:

```
PROXY gateway.proxy.example.com:1080; DIRECT
```

Which means proxy the traffic via an HTTP proxy (PROXY gateway.proxy.example.com:1080) or if that connection fails, do a direct connection instead. This line is the return line in a much bigger pac file, that looks for example like:

```javascript
function FindProxyForURL(url, host) {
    host = host.toLowerCase();
    if (host == "example.com") {
       return "PROXY gateway.proxy.example.com:1080; DIRECT";
    }
    return "DIRECT";
}
```

Within our API in rust we would have a service named `PacConnector<C, W = EasyWebClient>` which wraps another connector (similar to other connectors already in rama). It can be created from the environment or via a hardcoded url. It is also possible once we have system config support (https://github.com/plabayo/rama/issues/540) to load this from system settings as depending on the platform this PAC url can be set in all kind of ways.

As the connector service needs to download that file it will also need a web client. By default the `EasyWebClient` can be used, but a custom one should be possible.

For a runtime we could make use of `boa_runtime` which is more than good enough probably for what we need. Let's make that an abstraction detail that is hidden so we can still swap it out if needed/desired.

`rama-pac` will be exposed in `rama` under `rama::proxy::pac` when the `pac` feature is enabled. It is not to be part of `proxy-full` as it is a pretty heavy one . It can however be part of `full`.

---

Resources:

* Proxy PAC or WPAD Functions overview: https://pactester.brdbnt.com/pacfunctions.html
* MDN Web Docs on Proxy Auth Configuration (PAC) file: https://developer.mozilla.org/en-US/docs/Web/HTTP/Proxy_servers_and_tunneling/Proxy_Auto-Configuration_PAC_file
* PAC File best practices: https://www.websense.com/content/support/library/web/v76/pac_file_best_practices/pac_file_best_practices.pdf
* ZScaler docs on PAC FILES: https://help.zscaler.com/zia/understanding-pac-file 
* Cloudflare docs on PAC Files
[...]

PR #698, opened 2025-09-06
[WIP] Implement rama-pac

nagxsan:
# Details
## Issue number
#566 
## Summary
- Adds a `rama-pac` crate that will handle logic of creating a PAC Connector, getting and reading the PAC file for the outgoing request URL proxy options and route the request through first successful proxy.

### DIFF
--- a/rama-pac/src/connector/mod.rs
+++ b/rama-pac/src/connector/mod.rs
@@ -1,0 +1,55 @@
+use crate::{pac_fetcher::fetch_pac, pac_parser::parse_pac_file};
+use rama_core::{Context, Service, error::BoxError};
+use rama_http::{Uri, service::client::HttpClientExt};
+use rama_net::{
+    client::{ConnectorService, EstablishedClientConnection},
+    stream::{Socket, Stream},
+};
+use std::fmt::Debug;
+
+pub struct PACConnector<S, W> {
+    pub service: S,
+    pub web_client: W,
+    pub pac_uri: Uri,
+}
+
+impl<S, W> PACConnector<S, W>
+where
+    W: HttpClientExt + Send + Sync + 'static,
+{
+    pub fn new(service: S, web_client: W, pac_uri: Uri) -> Self {
+        Self {
+            service,
+            web_client,
+            pac_uri,
+        }
+    }
+}
+
+impl<S, W, Request> Service<Request> for PACConnector<S, W>
+where
+    S: ConnectorService<Request, Connection: Stream + Socket + Unpin, Error: Into<BoxError>>
+        + Send
+        + Sync
+        + 'static,
+    W: HttpClientExt + Send + Sync + 'static,
+    Request: Send + 'static,
+    <W as HttpClientExt>::ExecuteResponse: Into<String>,
+    <W as HttpClientExt>::ExecuteError: Debug
+{
+    type Response = EstablishedClientConnection<S::Connection, Request>;
+    type Error = BoxError;
+
+    async fn serve(&self, ctx: Context, req: Request) -> Result<Self::Response, Self::Error> {
+        // first get the pac file
+        let pac_file = fetch_pac(&self.web_client, self.pac_uri.clone())
+            .await
+            .unwrap();
+
+        // parse and get back result for the uri form the pac file
+        let proxy_options = parse_pac_file(&pac_file);
+
+        // based on the result, either connect to the uri or return error
+        // Ok(self.service.connect(ctx, req).await.unwrap())
+    }
+}
--- a/rama-pac/src/lib.rs
+++ b/rama-pac/src/lib.rs
@@ -1,0 +1,3 @@
+pub mod connector;
+pub mod pac_fetcher;
+pub mod pac_parser;
--- a/rama-pac/src/pac_fetcher/mod.rs
+++ b/rama-pac/src/pac_fetcher/mod.rs
@@ -1,0 +1,23 @@
+use std::fmt::Debug;
+
+use rama_core::{Context, error::BoxError};
+use rama_http::{Body, Uri, dep::http::Request, service::client::HttpClientExt};
+
+pub async fn fetch_pac<W>(web_client: &W, pac_uri: Uri) -> Result<String, BoxError>
+where
+    W: HttpClientExt + Send + Sync + 'static,
+    <W as HttpClientExt>::ExecuteResponse: Into<String>,
+    <W as HttpClientExt>::ExecuteError: Debug,
+{
+    let request = Request::builder()
+        .method("GET")
+        .uri(pac_uri)
+        .body(Body::empty())
+        .unwrap();
+    let response = web_client
+        .execute(Context::default(), request)
+        .await
+        .unwrap();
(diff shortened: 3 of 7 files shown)


rama-pac/src/connector/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    type Error = BoxError;
+
+    async fn serve(
+        &self,
+        ctx: Context,
+        req: Request<Body>,
+    ) -> Result<Self::Response, Self::Error> {
+        let pac_file = fetch_pac(&self.web_client, &self.pac_uri)
+            .await
+            .unwrap();
```
GlenDC, line 44: given it's wip this is ok, but obviously in the final code errors like this should be handled or bubbled up. no unwraps please unless you have a very good reason

rama-pac/src/connector/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                    *attempt.uri_mut() = proxy_uri;
+                }
+                ProxyDirective::Socks4(hp) => {
+                    let proxy_uri: Uri = format!("socks4://{hp}")
+                        .parse()
+                        .map_err(|e| format!("bad socks4 uri {hp}: {e}"))?;
+                    *attempt.uri_mut() = proxy_uri;
+                }
+                ProxyDirective::Socks5(hp) => {
+                    let proxy_uri: Uri = format!("socks5://{hp}")
```
GlenDC, line 81: The proxy info will have to be added in the extensions (context for now, but in some days part of request), so that the inner connectors will use it as proxy

rama-pac/src/connector/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                        .parse()
+                        .map_err(|e| format!("bad socks5 uri {hp}: {e}"))?;
+                    *attempt.uri_mut() = proxy_uri;
+                }
+            }
+            match self.service.serve(ctx.clone(), attempt).await {
```
GlenDC, line 87: Inner service is a connector service so might as well call it connector. E.g. is going to be something like an AutoConnector wrapped around tcp etc.

rama-pac/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+pub mod connector;
+pub mod pac_fetcher;
```
GlenDC, line 2: no need to make all modules public if you only use 1 or 2 types it's better to keep the public module structure flat.

rama-pac/src/pac_fetcher/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+    W::ExecuteError: std::error::Error + Send + Sync + 'static,
+    W::ExecuteResponse: Into<String>,
+{
+    let request = Request::builder()
+        .method("GET")
+        .uri(pac_uri)
+        .body(Body::empty())
+        .map_err(Into::<BoxError>::into)?;
+
+    let response = web_client.execute(Context::default(), request).await?;
```
GlenDC, line 16: given you already use `HttpClientExt` you can make this simpler by doing `web_client.get(pac_ure).send(Context::default())` or something like it

rama-pac/src/pac_fetcher/mod.rs
```diff
@@ -0,0 +1,21 @@
+use rama_core::{Context, error::BoxError};
+use rama_http::{Body, Request, Uri, service::client::HttpClientExt};
+
+pub async fn fetch_pac<W>(web_client: &W, pac_uri: &Uri) -> Result<String, BoxError>
+where
+    W: HttpClientExt,
```
GlenDC, line 6: We do not trait bound on `HttpClientExt` instead trait bound on regular `Service`. `HttpClientExt` is implemented for `Service`.

rama-pac/src/pac_parser/mod.rs
```diff
@@ -0,0 +1,90 @@
+use boa_engine::{Context as JsContext, JsString, JsValue, Source, js_string, property::Attribute};
+use boa_runtime::Console;
+use rama_core::error::BoxError;
+use rama_http::Uri;
+
+const PAC_STDLIB: &str = r#"
+// spec-ish helpers
+function dnsDomainIs(host, domain) {
```
GlenDC, line 8: Instead of using these js interpreted functions, use rust host functions please, so we can use std rama code and it will be easier to keep it straight

rama-pac/src/pac_parser/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+function localHostOrDomainIs(host, hostdom) { return host === hostdom || hostdom.startsWith(host + "."); }
+function dnsResolve(host) { return host; }          // stub; swap for real resolver if needed
+function myIpAddress() { return "127.0.0.1"; }      // stub
+function isInNet(ip, pattern, mask) { return false; } // stub
+"#;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ProxyDirective {
+    Direct,
+    Proxy(String),
```
GlenDC, line 27: We can probably do better, type-wise, no? As in a SocketAddress, or something like that?

rama-pac/src/pac_fetcher/mod.rs
```diff
@@ -0,0 +1,21 @@
+use rama_core::{Context, error::BoxError};
+use rama_http::{Body, Request, Uri, service::client::HttpClientExt};
+
+pub async fn fetch_pac<W>(web_client: &W, pac_uri: &Uri) -> Result<String, BoxError>
```
GlenDC, line 4: why is this public? This is just a detail of your connector I would think

rama-pac/src/connector/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+{
+    type Response = S::Response;
+    type Error = BoxError;
+
+    async fn serve(
+        &self,
+        ctx: Context,
+        req: Request<Body>,
+    ) -> Result<Self::Response, Self::Error> {
+        let pac_file = fetch_pac(&self.web_client, &self.pac_uri)
```
GlenDC, line 42: this should be cached in your connector I think with an ArcSwap together with a duration and only refetch it after some cache duration, default pretty high but can be overwritten.

rama-pac/src/connector/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+            pac_uri,
+        }
+    }
+}
+
+impl<S, W, Body> Service<Request<Body>> for PACConnector<S, W>
+where
+    S: Service<Request<Body>, Error: Into<BoxError> + Send + Sync + 'static>,
+    W: HttpClientExt + Send + Sync + 'static,
+    Body: Clone + Send + Sync + 'static,
```
GlenDC, line 30: Body is never clone. Neither should it. Your connector is not dealing with the actual Http request anyway, what you need to implement is a service as a connection service. For any request which has a target uri, but that doesn't have to be a http request perse. This way can also work for any other protocol that has uris and again it's; not to already serve the input but to just establish connection. The inner service is the actual connector stack establishing the connection.

GlenDC:
Closing this as inactive. I will pickup PAC soon myself.

PR #706, opened 2025-09-25
Issue 462/revisit service api (core) design in 0.3 milestone#2 moving extensions

soundofspace:
What this PR does (for #462):
- Add extensions to all Request/Response/Connection types
- Move ctx extensions to these extensions instead and remove extensions from context
- ExtensionsRef, and ExtensionsMut trait so we can implement this generically
- `debug-extensions` feature that implement a debug impl for extensions that can print all type_names stored in extensions, this made debugging extensions issues a lot easier
- ChainableExtensions -> this is something i want to improve (see next points) but for now it does help already
- Extractor to get extensions, currently this clones them, but we can probably do this similar to how we consume body. To be checked in followup pr also
- Add ExtensionsMut + Send bound on ConnectorService's connection and Send + 'static bounds on Error. This is similar to how we set bounds on Service trait and makes usage easier all around

What this Pr doesn't do
- Removing ctx. Some very minor things need to be migrated, but removing ctx from all API has a pretty big impact and I want to combine this with thinking how we expose/make these APIs as I believe we can do better. This is partially related to the never caching remark and making traits for that
- Huge refactors to be consistent. This will happen in follow up PR so we can focus specifically on that instead of getting lost in this big PR. Some refactoring already happened where needed. This includes renaming things, making this private vs public...
- A better design to transfer extensions. Right now we always transfer all of them, but depending on the place/service this can be very tricky and error prone (especially around upgrades). I'm already playing with some ideas to make this much nicer, but will implement this in a follow up PR. In general I believe we can make extensions easier and less error prone to work with

Unanswered questions that will require some input / more playing time:
- Right now we are able to add extensions to pretty much all connection types except the inner http-core types. Adding and passing them all through the stack there is pretty complex, but it might be worth it down the line. If extensions are everywhere it's pretty transparent how they are passed, right now we need to remember on which levels to pass them and where they were skipped. On the other hand, not having to think about extensions in places makes the code easier there and more focussed. Curious about what you think @GlenDC , but either way I still believe we can do much better in general, so might be a question we answer once all the other improvements are done

### DIFF
--- a/src/cli/service/ip.rs
+++ b/src/cli/service/ip.rs
@@ -15,10 +15,10 @@ use crate::{
         headers::{Accept, HeaderMapExt},
         layer::{
             forwarded::GetForwardedHeaderLayer, required_header::AddRequiredResponseHeadersLayer,
-            trace::TraceLayer, ua::UserAgentClassifierLayer,
+            trace::TraceLayer,
         },
         server::HttpServer,
-        service::web::response::{Html, IntoResponse, Json},
+        service::web::response::{Html, IntoResponse, Json, Redirect},
     },
     layer::{ConsumeErrLayer, LimitLayer, TimeoutLayer, limit::policy::ConcurrentPolicy},
     net::forwarded::Forwarded,
@@ -178,6 +178,12 @@ impl Service<Request> for HttpIpService {
     type Error = BoxError;
 
     async fn serve(&self, _ctx: Context, req: Request) -> Result<Self::Response, Self::Error> {
+        let norm_req_path = req.uri().path().trim_matches('/');
+        if !norm_req_path.is_empty() {
+            tracing::debug!("unexpected request path '{norm_req_path}', redirect to root");
+            return Ok(Redirect::permanent("/").into_response());
+        }
+
         let peer_ip = req
             .extensions()
             .get::<Forwarded>()
@@ -393,7 +399,6 @@ impl<M> IpServiceBuilder<M> {
         let http_service = (
             TraceLayer::new_for_http(),
             AddRequiredResponseHeadersLayer::default(),
-            UserAgentClassifierLayer::new(),
             ConsumeErrLayer::default(),
             http_forwarded_layer,
         )


examples/haproxy_client_ip.rs
```diff
@@ -56,11 +57,16 @@ async fn main() {
         let tcp_http_service = HttpServer::auto(Executor::graceful(guard.clone())).service(
             AddRequiredResponseHeaders::new(Router::new().get(
                 "/",
-                async |ctx: Context| -> Result<String, (StatusCode, String)> {
-                    let client_ip = ctx
+                async |req: Request| -> Result<String, (StatusCode, String)> {
+                    let client_ip = req
+                        .extensions()
                         .get::<Forwarded>()
```
GlenDC, line 63: This is probably something where a nice high level perhaps stand alone function could be nice, to easily extract (stateless) the peer client IP without having to remember to check these different sources one self

examples/http_connect_proxy.rs
```diff
@@ -137,12 +136,12 @@ async fn main() {
                                     "lucky_number": path.number,
                                 }))
                             },
-                            HttpMatcher::get("/*") => async move |ctx: Context, req: Request| {
+                            HttpMatcher::get("/*") => async move |_ctx: Context, req: Request| {
```
GlenDC, line 139: you cannot just remove `_ctx: Context` all together?

examples/http_key_value_store.rs
```diff
@@ -142,16 +144,16 @@ async fn main() {
                             HttpMatcher::method_get().or_method_head().and_path("/item/:key"),
                             // only compress the get Action, not the Post Action
                             CompressionLayer::new()
-                                .into_layer((async |Path(params): Path<ItemParam>, method: Method, ctx: Context| {
+                                .into_layer((async |Path(params): Path<ItemParam>, method: Method, req: Request| {
```
GlenDC, line 147: You can probably add an `Extensions` FromRequest` impl, as that would make it clear we only need extensions here instead of full request

GlenDC: Of course in future we would actually user typed state for this specific situation I would think, but the proposal still stands for similar code in general

examples/http_mitm_proxy_boring.rs
```diff
@@ -368,11 +368,12 @@ where
 
     let target_version = req.version();
     tracing::debug!("forcing egress http connection as {target_version:?} to ensure WS upgrade");
-    ctx.insert(TargetHttpVersion(target_version));
+    let mut extensions = Extensions::new();
```
GlenDC, line 376: need to perhaps add a TODO here, as I think we need to pass this from the mitm_websocket input parameters, no?

examples/http_user_agent_classifier.rs
```diff
@@ -44,8 +45,8 @@ async fn main() {
         .unwrap();
 }
 
-async fn handle(ctx: Context, _req: Request) -> Result<Response, Infallible> {
-    let ua: &UserAgent = ctx.get().unwrap();
+async fn handle(_ctx: Context, req: Request) -> Result<Response, Infallible> {
```
GlenDC, line 48: same question here, do we still need `_ctx: Context` to make it compile or so? Why?

justfile
```diff
@@ -56,11 +56,11 @@ hack:
 	@cargo install cargo-hack
 	cargo hack check --each-feature --no-dev-deps --workspace
 
-test:
-	cargo test --all-features --workspace
+test *ARGS:
```
GlenDC, line 59: I understand why you did it and agree with the change. Just out of curiosity, why did you use it? To filter on a specific test?

rama-core/src/context/mod.rs
```diff
@@ -1,128 +1,40 @@
...
-        &mut self.0
-    }
-}
-
 #[derive(Debug, Default, Clone)]
 /// Context passed to and between services as input.
 ///
 /// See [`crate::context`] for more information.
 pub struct Context {
     executor: Executor,
```
GlenDC, line 11: The executor also need to move into the extension, so you can do `get_or_default()` to easily spawn tasks, I suppose you'll do that in a next PR with the removal of context?

rama-core/src/extensions.rs
```diff
@@ -185,11 +261,193 @@ impl Extensions {
 }
 
 impl fmt::Debug for Extensions {
+    #[cfg(feature = "debug-extensions")]
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
GlenDC, line 265: What if we just enforce all extensions to be `fmt::Debug`, wouldn't that allow this to be always possible? In general there should anyway be little reason to not support fmt::Debug, cannot think of one at all. Or am I missing something else?

rama-core/src/extensions.rs
```diff
@@ -185,11 +261,193 @@ impl Extensions {
...
+}
+
+crate::combinators::impl_either!(impl_extensions_either);
+
+pub trait ChainableExtensions {
+    fn contains<T: Send + Sync + 'static>(&self) -> bool;
+    fn get<T: Send + Sync + 'static>(&self) -> Option<&T>;
+}
+
+impl<S, T> ChainableExtensions for (S, T)
```
GlenDC, line 417: there is a macro afaik in rama to make this easier to do for all the different kind of tuples up to a certain length

rama-core/src/extensions.rs
```diff
@@ -185,11 +261,193 @@ impl Extensions {
...
+                    $(crate::combinators::$id::$param(s) => s.extensions_mut(),)+
+                }
+            }
+        }
+    };
+}
+
+crate::combinators::impl_either!(impl_extensions_either);
+
+pub trait ChainableExtensions {
```
GlenDC, line 412: this deserves a bit more explanation though. Like why is this ever useful? Or at least useful enough to provide here as something so explicit

rama-core/src/extensions.rs
```diff
@@ -221,6 +479,49 @@ impl Clone for Box<dyn AnyClone + Send + Sync> {
     }
 }
 
+#[derive(Debug, Clone)]
+/// Wrapper type that can be injected into the dynamic extensions of a "Response",
+/// in order to preserve the [`Context`]'s extensions of the _Request_
+/// which was used to produce the _Response_.
+pub struct RequestContextExt(Extensions);
```
GlenDC, line 488: Are we still planning to replace this with something better in a follow up PR?

rama-core/src/generic_request.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// A generic request that implements all rama traits conditionally
+    ///
+    /// This request type can be used in places where you need a request/stream
+    /// that implements things such as [`ExtensionsRef`] without having to
+    /// create a custom type for it.
+    ///
+    /// This is mainly useful for testing or less import request types. In most
+    /// cases you should create a new type that implements all the needed traits,
+    /// but that is focussed specifically on that use case.
+    pub struct GenericRequest<T> {
```
GlenDC: I would call it `ExtensibleRequest` or something in that spirit, describes it much better than "generic"... which is.... generic?

rama-core/src/generic_request.rs
```diff
@@ -0,0 +1,97 @@
...
+        #[pin]
+        pub request: T,
+        pub extensions: Extensions,
+    }
+}
+
+impl<T> GenericRequest<T> {
```
GlenDC: Make `Debug` conditional only if `T` is debug, and also implement Default conditionally if T implements Default.

(FWIW, as you love doing these things, it is probably trivial to make a macro_rules to automatically implement stuff like Default, Debug, Clone when generics are also that, seems like the kind of PR you would throw at me on some random foggy night)

rama-core/src/stream/peek.rs
```diff
@@ -28,11 +46,12 @@ impl<P, S> PeekStream<P, S> {
     /// peek [`AsyncRead`] and inner [`Stream`].
     ///
     /// [`Stream`]: super::Stream
-    pub fn new(peek: P, inner: S) -> Self {
+    pub fn new_with_fresh_extensions(peek: P, inner: S) -> Self {
         Self {
             done_peek: false,
             peek,
             inner,
```
GlenDC: that does mean inner still has its extensions whereas with `fn new` the extensions are consumed. Why the difference at all. Why not just always use the extensions of inner (and also leave it within inner, just delegate, let the parent/callee wipe these if it wishes to do so?)

rama-core/src/stream/rewind.rs
```diff
@@ -1,15 +1,27 @@
...
 #[derive(Debug)]
 pub struct Rewind<T> {
     pre: Option<Bytes>,
     inner: T,
+    extensions: Extensions,
+}
+
+impl<T: ExtensionsMut> Rewind<T> {
+    pub fn new_buffered(mut io: T, buf: Bytes) -> Self {
+        let extensions = io.take_extensions();
```
GlenDC: same remark as with PeakStream

rama-haproxy/src/client/layer.rs
```diff
@@ -215,13 +216,16 @@ where
...
-            .and_then(|f| f.client_socket_addr())
-            .or_else(|| ctx.get::<SocketInfo>().map(|info| *info.peer_addr()))
-            .ok_or_else(|| {
-                OpaqueError::from_display("PROXY client (v1): missing src socket address")
-            })?;
+        let src = {
+            let ext_chain = (&conn, &req);
+            ext_chain
+                .get::<Forwarded>()
+                .and_then(|f| f.client_socket_addr())
```
GlenDC, line 223: I hope you keep a list for all these kind of chains, as ideally we abstract hem away behind easy to use abstractions (be it traits or functions, depends on the situation I suppose). This way we can easily extend them if we know of more possible sources / ext econtrols ,and ensure the order etc is correct.

rama-haproxy/src/client/layer.rs
```diff
@@ -627,39 +656,40 @@ mod tests {
                     })
                 },
             ));
-            svc.serve(input_ctx, ()).await.unwrap();
+            svc.serve(Context::default(), ext).await.unwrap();
```
GlenDC: That's very confusing, I would use your GenericRequest<()> in that case, as using extensions directly might look weird if this is the first code someone reads

rama-haproxy/src/client/layer.rs
```diff
@@ -627,39 +656,40 @@ mod tests {
...
+                ext.insert(Forwarded::new(ForwardedElement::forwarded_for(
                     NodeId::try_from("127.0.0.1:80").unwrap(),
                 )));
-                ctx
+                ext
             },
         ] {
             let svc = HaProxyLayer::udp().with_payload(vec![42]).layer(service_fn(
-                async move |ctx, req| {
+                async move |ctx, req: Extensions| {
```
GlenDC: same same, GenericRequest<()> is probably better here

rama-haproxy/src/server/layer.rs
```diff
@@ -242,7 +243,7 @@ where
 
 #[cfg(test)]
 mod test {
-    use rama_core::service::service_fn;
+    use rama_core::{generic_request::GenericRequest, service::service_fn};
```
GlenDC: FWIW we probably should pub use `GenericRequest` directly into rama_core given it is I imagine anyway the only type of `generic_request` mod that is pub exposed. Avoids the duplicate `foo::Foo`

rama-http-backend/src/client/conn.rs
```diff
@@ -234,6 +244,7 @@ where
                 let svc = HttpClientService {
                     sender: SendRequest::Http1(Mutex::new(sender)),
                     http_req_inspector: self.http_req_inspector_svc.clone(),
+                    extensions: conn_extensions,
```
GlenDC, line 247: what's the expected / possible use to put the conn extensions into the HttpClientService struct? And should it have no relationship / influence on the req extensions?

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -302,13 +299,41 @@ pin_project! {
...
         inner: Connection<S>,
+        extensions: Extensions
+    }
+}
+
+impl<S: ExtensionsMut + Unpin + Stream> MaybeHttpProxiedConnection<S> {
+    fn direct(mut conn: S) -> Self {
+        let extensions = conn.take_extensions();
+        Self {
+            inner: Connection::Direct { conn },
```
GlenDC, line 308: I still do not understand the motivation to take it out of inner instead of just delegating ext usage to inner for wrapper cases like these

rama-http-backend/src/client/svc.rs
```diff
@@ -70,6 +71,8 @@ where
             }
         }
 
+        req.extensions_mut().extend(self.extensions.clone());
```
GlenDC: these (`self.extensions`) come from the connection, right? I wonder if one expects these in the request extensions, as it kinda is already upstream (prior to response).

Then again you add them as req extensions for the response, so perhaps is ok, dunno, something to discuss perhaps

rama-http-backend/src/server/service.rs
```diff
@@ -3,6 +3,7 @@
 use super::HttpServeResult;
 use super::hyper_conn::HttpCoreConnServer;
```
GlenDC, line 4: probably should rename this mod at point

rama-http-core/src/client/conn/http1.rs
```diff
@@ -555,7 +555,7 @@ mod upgrades {
                 Ok(proto::Dispatched::Shutdown) => Poll::Ready(Ok(())),
                 Ok(proto::Dispatched::Upgrade(pending)) => {
                     let Parts { io, read_buf } = self.inner.take().unwrap().into_parts();
-                    pending.fulfill(Upgraded::new(io, read_buf));
+                    pending.fulfill(Upgraded::new_with_fresh_extensions(io, read_buf));
```
GlenDC: in an ideal world there are no special cases. When there are special cases they should have a good explanation. As such, why do we not want to preserve the original connection extensions?

rama-http-core/src/server/conn/auto.rs
```diff
@@ -105,12 +105,12 @@ impl Builder {
     {
         let state = match self.version {
             Some(Version::H1) => {
-                let io = Rewind::new_buffered(io, Bytes::new());
+                let io = Rewind::new_buffered_with_fresh_extensions(io, Bytes::new());
```
GlenDC: Same here, what's the footgun here that we need to have this (what feel like a) hack?

rama-http/src/layer/har/spec.rs
```diff
@@ -400,7 +401,7 @@ impl TryFrom<Request> for crate::Request {
 
 impl Request {
     pub fn from_http_request_parts(
-        ctx: &Context,
+        _ctx: &Context,
```
GlenDC, line 404: why still have it as a parameter if you anyway do not use it, not like you are implementing a trait or so

rama-http/src/layer/header_from_str_config.rs
```diff
@@ -313,8 +316,8 @@ mod test {
             .unwrap();
 
         let inner_service =
-            rama_core::service::service_fn(async |ctx: Context, _req: Request<()>| {
-                let id: &usize = ctx.get().unwrap();
+            rama_core::service::service_fn(async |_ctx: Context, req: Request<()>| {
```
GlenDC, line 319: have remarked it in an earlier file, but these should be able to be declared (already) as `            rama_core::service::service_fn(async |req: Request<()>|`, no?

rama-http/src/service/client/ext.rs
```diff
@@ -575,6 +577,18 @@ where
         }
     }
 
+    /// Get mutable access to the underlying [`Extensions`]
+    ///
+    /// This function will return None if [`Extensions`] are not available,
+    /// or if this builder is in an error state
+    pub fn extensions_mut(&mut self) -> Option<&mut Extensions> {
```
GlenDC, line 584: given you did it for request/response as well. Should you not impl this for ExtensionsRef and ExtensionsMut instead?

rama-http/src/service/web/endpoint/extract/extensions.rs
```diff
@@ -0,0 +1,15 @@
+use super::FromRequestContextRefPair;
+use crate::request::Parts;
+use rama_core::{Context, extensions::Extensions};
+use std::convert::Infallible;
+
+impl FromRequestContextRefPair for Extensions {
```
GlenDC, line 6: I think we want to better implement this not by ref but by value

rama-net/src/client/conn.rs
```diff
@@ -38,9 +40,9 @@ impl<S: Clone, Request: Clone> Clone for EstablishedClientConnection<S, Request>
 /// but from a Rama POV it is mostly used for UX trait bounds.
 pub trait ConnectorService<Request>: Send + Sync + 'static {
     /// Connection returned by the [`ConnectorService`]
-    type Connection;
+    type Connection: Send + ExtensionsMut;
```
GlenDC, line 43: Do we need to enforce this here? (extensionMut that is) Is there no legit use case where we do not want this?

rama-net/src/client/pool/mod.rs
```diff
@@ -112,6 +113,7 @@ pub struct LeasedConnection<C, ID> {
     active_slot: ActiveSlot,
     returner: ConnReturner<C, ID>,
     failed: AtomicBool,
+    extensions: Extensions,
```
GlenDC: Why does leasedConnection require extensions itself, instead of it just being part of pooled_conn?

rama-net/src/stream/layer/opentelemetry.rs
```diff
@@ -370,14 +371,14 @@ mod tests {
...
-        .with_attributes(|size_hint: usize, _ctx: &Context| {
+        .with_attributes(|size_hint: usize, _ext: &Extensions| {
             let mut attributes = Vec::with_capacity(size_hint + 1);
             attributes.push(KeyValue::new("test", "attribute_fn"));
             attributes
         })
         .into_layer(());
 
-        let attributes = svc.compute_attributes(&Context::default());
+        let attributes = svc.compute_attributes(&Extensions::default());
```
GlenDC, line 381: I wonder if compute_attributes shouldn't take an `Option<&Extensions>` instead

rama-net/src/stream/layer/tracker/bytes.rs
```diff
@@ -37,6 +40,19 @@ pin_project! {
         written: Arc<AtomicUsize>,
         #[pin]
         stream: S,
+        extensions: Extensions,
```
GlenDC: I'm still not convinced that we need to take the extensions each time instead of just delegating where nothing custom

rama-net/src/stream/layer/tracker/bytes.rs
```diff
@@ -50,20 +66,25 @@ impl<S: fmt::Debug> fmt::Debug for BytesRWTracker<S> {
     }
 }
 
-impl<S> BytesRWTracker<S> {
+impl BytesRWTracker<()> {
```
GlenDC: why did you make this change?

rama-socks5/src/server/udp/relay.rs
```diff
@@ -2,6 +2,8 @@ use std::io::ErrorKind;
 
 use rama_core::bytes::{Bytes, BytesMut};
 use rama_core::error::{BoxError, ErrorExt, OpaqueError};
+#[cfg(feature = "dns")]
+use rama_core::extensions::Extensions;
```
GlenDC: why add a separate line for this if you already have a feature dns gate below for imports?

rama-tcp/src/lib.rs
```diff
@@ -20,5 +20,5 @@
 pub mod client;
 pub mod pool;
 pub mod server;
-
-pub use tokio::net::{TcpSocket, TcpStream};
+pub mod stream;
+pub use stream::{TcpStream, TokioTcpStream};
```
GlenDC, line 24: seems confusing to have both TcpSteeam and TokioTcpStream... ALso not sure we need to expose that one given we consier tokio to be stable anyway Before was ok as it was only one, but now having these both, dunno how I feel about that

rama-tcp/src/stream.rs
```diff
@@ -0,0 +1,109 @@
...
+use pin_project_lite::pin_project;
+use rama_core::{
+    extensions::Extensions,
+    extensions::{ExtensionsMut, ExtensionsRef},
+};
+use rama_net::stream::Socket;
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+pub use tokio::net::TcpStream as TokioTcpStream;
+
+pin_project! {
```
GlenDC, line 17: so far it doesn't seem to add anything on top of `GenericRequest`? Except the `From` impl, which you could just as easily add to `GenericRequest`?

rama-tls-boring/src/client/mod.rs
```diff
@@ -9,6 +9,8 @@ pub use tls_stream_auto::AutoTlsStream;
 mod tls_stream;
 pub use tls_stream::TlsStream;
 
+pub use rama_boring_tokio::SslStream as BoringTlsStream;
```
GlenDC, line 12: Same question as tcp I suppose, does a user ever have a need to access this inner one directly?

rama-tls-boring/src/server/tls_stream.rs
```diff
@@ -0,0 +1,115 @@
...
+use rama_boring_tokio::SslStream;
+use rama_core::{
+    extensions::Extensions,
+    extensions::{ExtensionsMut, ExtensionsRef},
+    stream::Stream,
+};
+use tokio::io::{AsyncRead, AsyncWrite};
+
+pin_project! {
+    /// A stream which can be either a secure or a plain stream.
```
GlenDC: copy paste comment?

rama-tls-boring/src/server/tls_stream.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn poll_write_vectored(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+        bufs: &[std::io::IoSlice<'_>],
+    ) -> std::task::Poll<Result<usize, std::io::Error>> {
+        self.project().inner.poll_write_vectored(cx, bufs)
+    }
+
+    fn is_write_vectored(&self) -> bool {
+        false
```
GlenDC: why do you not delegate this?

rama-tls-rustls/src/client/tls_stream.rs
```diff
@@ -0,0 +1,98 @@
+use super::RustlsTlsStream;
+use pin_project_lite::pin_project;
+use rama_core::extensions::{Extensions, ExtensionsMut, ExtensionsRef};
+use std::{
+    io,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
```
GlenDC, line 9: same same, seems very similar to GenericRequest

src/http/client/proxy_connector.rs
```diff
@@ -155,17 +147,80 @@ pin_project! {
     pub struct MaybeProxiedConnection<S> {
         #[pin]
         inner: Connection<S>,
+        extensions: Extensions,
+    }
+}
+
+impl<S: ExtensionsMut> MaybeProxiedConnection<S> {
```
GlenDC, line 153: still not convinced of neither the "fresH" requirement, neither of the taking extensions like that

tests/turmoil/stream.rs
```diff
@@ -0,0 +1,108 @@
...
+
+pin_project! {
+    pub struct TcpStream {
+        #[pin]
+        pub stream: TurmoilTcpStream,
+        pub extensions: Extensions,
+    }
+}
+
+impl TcpStream {
```
GlenDC, line 25: Given this is for a test... do we really need to impl some new type instead of using your Generic one?

tests/integration/ua_emulation.rs
```diff
@@ -455,8 +450,10 @@ where
     type Response = EstablishedClientConnection<MockSocket, Request>;
 
     async fn serve(&self, ctx: Context, req: Request) -> Result<Self::Response, Self::Error> {
-        let (client_socket, server_socket) = new_mock_sockets();
-
+        // Store request extensions also on our server socket so we have acces to things like state
```
GlenDC: not better make a clean solution then instead of doing this here?

GlenDC (changes requested):

examples/haproxy_client_ip.rs
soundofspace: Yep agree I'm keeping todo of all these things but wanted to solved them in separate PRs so its easier to discuss and review

examples/http_connect_proxy.rs
soundofspace: I can but the plan was to remove context in the next PR everywhere, so didnt really invest time into doing it now partially, so the focus of this pr was really to migrate extensions

examples/http_key_value_store.rs
soundofspace: Yea agree, that was part of the improvements I was thinking about + quite some other ones. If good for you I would again do this in followup

examples/http_user_agent_classifier.rs
soundofspace: I just completely ignored it and did as little refactoring as needed, so it could all be done in a single pr

justfile
soundofspace: Yea to run special combinations of things not something that deserves a different task but that was still handy to be able to just use this target with extra args

examples/http_mitm_proxy_boring.rs
soundofspace: Added todo and add it to my list of things that need follow up

rama-core/src/context/mod.rs
soundofspace: yes

rama-core/src/extensions.rs
soundofspace: Extensions don't need to be Debug even. This logic is just to print the types which are stored in extensions, not the content of them. In my personal debugging having the extensions (names) that are available was already enough. Having this also print all the values of these extensions would probably way to overkill and too much info.

I feature gated this as for this logic to work it always does type_name when adding new things to extensions which also needs to allocated a string, for something like extensions which could be in very hot loops I thought that might be overkill for the very rare occasions where you do need this logic (to figure out why it was / or wasn't there, mainly if you forgot to transfer extensions across some boundary)

GlenDC: I am a bit puzzled why we need a string and separate for this to work though. Could you perhaps put a playground together and show me why it doesn't work without that explicit extra info?

rama-core/src/extensions.rs
soundofspace: Saw it indeed, but I think I will be replacing this trait with a better option, so only implement the bare minimum for now to make this work. Good again if I leave it as is?

rama-core/src/extensions.rs
soundofspace: yes

rama-core/src/extensions.rs
soundofspace: The main use case is right now you want to get some info out of extensions, and you can get it from the connection or the request, and you dont really know or care where it comes from. By chaining them it's easier to get this. I do plan to make this more of first class citizen problem and that should replace this trait completely

rama-core/src/generic_request.rs
soundofspace: Is indeed something on my list, but stayed away from macros on purpose for now a bit as i tend to spend too much time on them when i play with them (but will happen soon)

rama-core/src/stream/peek.rs
soundofspace: Something i wasn't completely sure of but options we have are these
1. Leave extensions in inner and delegate, but require inner to always implement this.
  - Pro: Only one constructor
  - Con: Cannot be used with inner that doesn't implement this. Might be fine since we see extensions as first class
2. Leave extensions in inner and conditionally delegate if inner implements them
  - Pro: more flexible
  - Con: types more complex and you cannot assume if we have `PeekStream` that we extensions. As all rama native types now implement extensions, this feels wrong?
3. Move extensions from inner to this, but also support constructor with fresh extensions
  - Pro: Once we have a rama type, here PeekStream, we always have extensions, no matter what the inner type is/was
  - Pro: Can be used to convert external types to rama native types with explicit different constructor (eg easy to search all usages of these)
  - Con: Two constructors
4. Move extensions from inner to this and require inner to implement extensions
  - Pro: only one constructor which makes it more predictable
  - Con: inner type already needs to implement extensions, so already needs to be a rama specific type

I wanted to streamline this approach between all components so PeakStream logic should be very similar to BoringTlsStream for example. So I went with option 3 because delegating extensions to inner is quite a bit more complex and has extra difficulties, whilst just moving extensions to top level type made it really clear that we own extensions and that they travelled across a boundary (and were transferred). I also should to always create extensions for rama-native types, so inside rama we really treat extensions as a mandatory part of inputs/outputs/requests all the way.

I do agree that for things like Peakstream we might implement different logic, but for now I thought streamlining had more benefits, but definitely open to discuss this in more detail on how you see this and how we want to proceed

rama-haproxy/src/client/layer.rs
soundofspace: yes doing that

rama-haproxy/src/client/layer.rs
soundofspace: aha thought i changed all of them, but forgot some

rama-core/src/extensions.rs
GlenDC: Why do we need it at all, now?

rama-http-backend/src/client/conn.rs
soundofspace: A client can be used multiple times so we need to store extensions so we can also set these in follow up requests. The way I see it you want connector extensions also available for requests (be it in a better format in the near future, eg similar like we do reqContextExt).

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
soundofspace: It could be possible, see points above

rama-http-backend/src/server/service.rs
soundofspace: agree

rama-http-backend/src/client/svc.rs
soundofspace: yep I think this problem will become much clearer ones we defines boundaries and different logic for them.
Eg now we just copy everything, but we could this in a lot of different ways

rama-http-core/src/client/conn/http1.rs
soundofspace: This is example of inner type which is not yet rama-native (meaning it implements things we expect it to, here specifically extensionsMut). This is what I meant in the PR description of it didnt implement extensions for all inner http-core types. I could definitely do this, but the types in there are quite complex already and didnt really see benefit of adding it vs the extra complexity. But I also see the added complexity it brings by not doing it, so open for all feedback here on how you see this...

rama-http-core/src/server/conn/auto.rs
soundofspace: Same remark as above, open for feedback

rama-http/src/layer/har/spec.rs
soundofspace: When i removed it in some places unexpected issues came up, and wanted to postpone this to next PR where I would also rethink all these APIs as removing ctx from them opens up new possibilities of how to do it (sometimes in a much cleaner way)

rama-http/src/layer/header_from_str_config.rs
soundofspace: yes, but later, unless you really think it should happen now

rama-http/src/service/web/endpoint/extract/extensions.rs
soundofspace: Agree, but have also different ideas of how todo it, but this felt like a good enough solution to get this pr through

rama-net/src/client/conn.rs
soundofspace: The way i see it, Extensions is now a first class citizen and to be available everywhere inside rama-native types. So added this bound here as it makes Connectors bounds way more ergonomic (eg this ExtensionsMut and Send was places everywhere). Was thinking about adding it to Service also but there its more complex so didnt do it there

rama-net/src/client/pool/mod.rs
soundofspace: To be able to isolate extensions between runs
1. Pooled connection extensions: extensions that were stored on connection when it was added to pool
2. Leased connections extensions: clone of pooled_conn as starting point
3. Other connectors can add more extensions
4. Connection goes back to pool -> we start from pooled_conn_extensions again so we dont leak extensions between runs

rama-net/src/stream/layer/opentelemetry.rs
soundofspace: Added to my list of todos

rama-net/src/stream/layer/tracker/bytes.rs
soundofspace: Yea maybe agree here, not sure yet

rama-net/src/stream/layer/tracker/bytes.rs
soundofspace: Bound was on inner new fn, moved it to external bound

rama-socks5/src/server/udp/relay.rs
soundofspace: missed it during refactor

rama-tcp/src/lib.rs
soundofspace: I kinda exposed them on purpose to make sure the naming would be consistent everywhere vs relying on import renaming everywhere. This will make it easier if we go down the line of always creating a tokioStream first and then convert it to a rama one, but yea open for other approaches here

rama-tcp/src/stream.rs
soundofspace: My lazy side also wanted to do this everywhere. But having a separate type is much easier in compiler errors, allows use to add more constructors specific to this (skipped this for now, but probably something I want to do very soon), and some other things, but yea very fair remark. But in my view having separates types is much more flexible towards the future, with only very little work (up front). But definitely open to switch this approach if you have other remarks about it

rama-tls-boring/src/client/mod.rs
soundofspace: I thought that kinda alligned with the vision of exposing all individual blocks so user could customize / swap components easily, if this goes to far, let me know

rama-tls-boring/src/server/tls_stream.rs
soundofspace: yes

rama-tls-boring/src/server/tls_stream.rs
soundofspace: Mistake and then copy paste probably, fixed it

rama-tls-rustls/src/client/tls_stream.rs
soundofspace: Same remark as above also

src/http/client/proxy_connector.rs
soundofspace: Kinda agree here, main reason was streamline it everywhere, but we might indeed have different approaches depending on the type

tests/turmoil/stream.rs
soundofspace: No really, main issue was `GenericRequest` being so generic would really love an assist for a better name. That is not only focussed on Extensions and Request. Something like that signal `GenericRamaInput` or `WrappedInput`...

tests/integration/ua_emulation.rs
soundofspace: fair, done

examples/http_user_agent_classifier.rs
GlenDC: Just makes me wonder how you are doing these code changes. Using tooling I suppose, because in my eyes if you would do it manually: changing `ctx` to `_ctx` is same work as removing it all together. But I suppose you auto fix this somehow if that is the approach you choose to take?

rama-core/src/generic_request.rs
GlenDC: > Make Debug conditional only if T is debug, and also implement Default conditionally if T implements Default.

^This is something you should tackle still now though.

The other bit is not meant for this PR indeed.

rama-core/src/stream/peek.rs
GlenDC: We need to have a talk about this. Just feels like something you can easily shoot yourself in the foot with... Because what if you turn it into "inner" again. The extension would have been lost? Also in case the inner was relying on the extension it will no longer have access to it after having been wrapped...

rama-http-backend/src/client/conn.rs
GlenDC: hmmmm ok, we need to discuss how we see this evolve

rama-http-backend/src/server/service.rs
GlenDC: add it to your backlog please, is out of scope for this PR anyway, but for sure one day

rama-http/src/service/web/endpoint/extract/extensions.rs
GlenDC: let's discuss

rama-net/src/client/conn.rs
GlenDC: I think either it is first class and it should be on service trait, or we do not do it at alll. Only putting it on ConnectorService feels half baked.

Then again, not sure we really want this. But I suppose we do...

rama-net/src/client/pool/mod.rs
GlenDC: mhmmm ok we need to discuss this, I think you are right, but at the very least I think we need these kind of flows and logic documented somewhere (as an abstract concept). Easier to refer to in future as well

rama-net/src/stream/layer/tracker/bytes.rs
GlenDC: Sure but why? Feels weird to me to do as we anyway return <S> as the only function in this impl block?

rama-tcp/src/lib.rs
GlenDC: Yeah Same as boring I think I agree with the intent. I am just worried that it isn't clear really from the name what is a `TcpStream` vs a `TokioTcpStream`. For same reason that that meaning is also lost in the `GenericRequest` name.

rama-tcp/src/stream.rs
GlenDC: Fair enough. Perhaps you are right. Still not convinced, even though I find your arguments convincing and I am inclined to agree.

rama-tls-boring/src/client/mod.rs
GlenDC: Yes, but the naming is not clear enough.

rama-core/src/generic_request.rs
soundofspace: Already pushed that

rama-core/src/extensions.rs
soundofspace: Used in things such as haproxy to make it easier will be replaced by proper chaining

rama-http/src/service/client/ext.rs
soundofspace: No this builder is special as its returns option, to handle the case if builder is in error state

rama-http/src/io/upgrade.rs
```diff
@@ -126,6 +132,7 @@ impl Upgraded {
     {
         Self {
             io: Rewind::new_buffered(Box::new(io), read_buf),
+            extensions: Extensions::new(),
```
GlenDC, line 135: In future PR this will need the extensions from IO as its parent in a frozen state. Similar to how `BytesMut` has `freeze` we probably also need something like that for Extensions, which should basically clone `self` and return a new extensions with self as a arc or something such that the fallback to parents also work (as deep as possible).

Just leaving it as a note, you can resolve this as-is once you read it.

rama-net/src/client/pool/mod.rs
```diff
@@ -125,6 +126,26 @@ impl<C, ID> LeasedConnection<C, ID> {
...
+            .conn
+            .extensions()
+    }
+}
+
+impl<C: ExtensionsMut, ID> ExtensionsMut for LeasedConnection<C, ID> {
+    fn extensions_mut(&mut self) -> &mut Extensions {
+        self.pooled_conn
+            .as_mut()
+            .expect("only None after drop")
```
GlenDC, line 143: FYI, for your backlog, not for this PR, so just read/ack and close as-is, but there are also unsafe primitives that allow you this in a more explicit manner. So we do not need to do expect but instead use unsafe and acknowledge that this is fine. Forget what the primitive is, but there's something in there which does basically the same but which is more clear/explicit than an option and intended for use cases like this

GlenDC (approved):

rama-http/src/io/upgrade.rs
soundofspace: Yep agree there but here io is tricky especially if its a h2 connection. But yea the frozen concept will work nicely here

GlenDC: Not really. Shouldn't make a difference for h2. What happens with h2 is that just that stream is tunneling. That stream is what I called pseudo connections. In case of h2 your hierarchy is

```
tcp
   |
  stream
     |
     http req 
```

So your IO Extensions will in that case be `stream`, with the tcp one already frozen and shared among all streams. That's something that `http` core should be able to facilitate.

GlenDC: (again outside this PR, just letting you know the direction that I think it should go towards, open ofc for feedback)

PR #715, opened 2025-10-08
Issue 462/revisit service api (core) design in 0.3 milestone#3 no remove from extensions

soundofspace:
- No more support for `remove` on extensions, this makes them much more predictable and will leave a trace of info. Especially once we add support for "parent" extensions
- We can remove `clear_extensions_safely`. The underlying original logic in hyper was needed to prevent a deadlock that got triggered because extensions were dropped that included a lock in the drop, while holding that same lock in these functions. Because we now always keep extensions around we are not affected by that problem
- Refactor upgrade::on to upgrade::handle_upgrade where we never have support None upgrades. Instead we throw error if we did an unexpected upgrade, or we already handled the upgrade
- Remove legacy tls logic that should have been removed already

For separate PR:
- Add parent extensions (this includes the todo from setting request extensions as parent extensions for response)
- On server side we use RamaHttpService to store extensions from io, and then reapply them. Should be pretty doable to do this in http-core where we have io access and create the actual request. Should also make this less error prone. But I has to potential to be a quite invasive change, so wont do that here

### DIFF
--- a/rama-core/src/extensions.rs
+++ b/rama-core/src/extensions.rs
@@ -169,42 +169,29 @@ impl Extensions {
     /// Returns true if the `Extensions` or parents contains the given type.
     #[must_use]
     pub fn contains<T: Send + Sync + 'static>(&self) -> bool {
-        if self.contains_without_chaining::<T>() {
-            true
-        } else {
-            self.parent_extensions
-                .as_ref()
-                .map(|parent| parent.contains::<T>())
-                .unwrap_or_default()
-        }
-    }
-
-    /// Returns true if the `Extensions` contains the given type.
-    #[must_use]
-    pub fn contains_without_chaining<T: Send + Sync + 'static>(&self) -> bool {
         self.map
             .as_ref()
             .map(|map| map.contains_key(&TypeId::of::<T>()))
-            .unwrap_or_default()
+            .unwrap_or_else(|| {
+                self.parent_extensions
+                    .as_ref()
+                    .map(|parent| parent.contains::<T>())
+                    .unwrap_or_default()
+            })
     }
 
     /// Get a shared reference to a type previously inserted on this `Extensions` or any of the parents
     #[must_use]
     pub fn get<T: Send + Sync + 'static>(&self) -> Option<&T> {
-        self.get_without_chaining().or_else(|| {
-            self.parent_extensions
-                .as_ref()
-                .and_then(|parent| parent.get())
-        })
-    }
-
-    /// Get a shared reference to a type previously inserted on this `Extensions`.
-    #[must_use]
-    pub fn get_without_chaining<T: Send + Sync + 'static>(&self) -> Option<&T> {
         self.map
             .as_ref()
             .and_then(|map| map.get(&TypeId::of::<T>()))
             .and_then(|boxed| (**boxed).as_any().downcast_ref())
+            .or_else(|| {
+                self.parent_extensions
+                    .as_ref()
+                    .and_then(|parent| parent.get())
+            })
     }
 
     /// Get an exclusive reference to a type previously inserted on this `Extensions`.
--- a/rama-http/src/io/upgrade.rs
+++ b/rama-http/src/io/upgrade.rs
@@ -319,11 +319,7 @@ mod sealed {
     fn handle_upgrade<T: ExtensionsRef>(
         obj: T,
     ) -> impl Future<Output = Result<Upgraded, OpaqueError>> + 'static {
-        let on_upgrade = match obj
-            .extensions()
-            .get_without_chaining::<OnUpgrade>()
-            .cloned()
-        {
+        let on_upgrade = match obj.extensions().get::<OnUpgrade>().cloned() {
             Some(on_upgrade) => {
                 trace!("upgrading this: {:?}", on_upgrade);
                 if on_upgrade.has_handled_upgrade() {
--- a/tests/integration/ua_emulation.rs
+++ b/tests/integration/ua_emulation.rs
@@ -259,7 +259,6 @@ async fn test_ua_emulation() {
         }
 
         async fn server_svc_fn(req: Request) -> Result<Response, Infallible> {
-            println!("extensions: {:?}", req.extensions());
             let state = req.extensions().get::<State>().unwrap();
             let expected = state.expected;
             let description = state.description;
@@ -463,7 +462,6 @@ where
         let (client_socket, mut server_socket) = new_mock_sockets();
 
         if let Some(extensions) = req.extensions().get::<ServerExtensions>() {
-            println!("extensions for server: {:?}", extensions.0);
             *server_socket.extensions_mut() = extensions.0.clone();
         }
 


soundofspace:
@GlenDC this is what happens when remove `Extensions.remove()`. Old logic to chain TLS config has also been removed because it has been replaced and was actually not in use anywhere (anymore). This PR is not ready and is currently a pretty mechanical change, but should serve as a good starting point to discuss if we want to commit to this

rama-http-core/src/h2/proto/streams/recv.rs
```diff
@@ -337,8 +337,9 @@ impl Recv {
         // frame or the user violated the contract.
         match stream.pending_recv.pop_front(&mut self.buffer) {
             Some(Event::Headers(peer::PollMessage::Client(mut response))) => {
-                if let Some(mut request_ext) = stream.encoded_request_extensions.take() {
-                    if let Some(request_headers) = request_ext.remove::<RequestHeaders>() {
+                if let Some(request_ext) = stream.encoded_request_extensions.take() {
+                    // TODO instead of this manual stuff do we just see request extensions as parent extensions for response?
```
GlenDC, line 341: yes

rama-http/src/io/upgrade.rs
```diff
@@ -316,31 +316,35 @@ mod sealed {
...
+                .cloned()
                 .unwrap_or_else(OnUpgrade::none)
         }
     }
 
     impl<B> CanUpgrade for &'_ mut Request<B> {
         fn on_upgrade(self) -> OnUpgrade {
             self.extensions_mut()
-                .remove::<OnUpgrade>()
+                .get::<OnUpgrade>()
```
GlenDC: OnUpgrade we need to better mark that it is already upgraded, so that evidence remains there but we change the status

rama-http-core/src/h2/proto/streams/recv.rs
soundofspace: See description: but this will be done in separate pr focussed on parent chaining and tools around it

soundofspace: Which now means the PR that will introduce `Vec<extensions>`

rama-http/src/io/upgrade.rs
```diff
@@ -93,31 +100,37 @@ pub struct Parts<T> {
...
 ///
 /// - `http::Request<B>`
 /// - `http::Response<B>`
-/// - `&mut rama_http::Request<B>`
-/// - `&mut rama_http::Response<B>`
-pub fn on<T: sealed::CanUpgrade>(msg: T) -> OnUpgrade {
-    msg.on_upgrade()
+/// - `&rama_http::Request<B>`
+/// - `&rama_http::Response<B>`
+pub fn handle_upgrade<T: sealed::HandleUpgrade>(
```
soundofspace, line 104: With this change in logic this seemed like a better name, do you agree or do you think we should keep the original one?

rama-core/src/extensions.rs
```diff
@@ -86,6 +88,21 @@ impl Extensions {
             map: None,
             #[cfg(debug_assertions)]
             type_map: None,
+            parent_extensions: None,
+        }
+    }
+
+    #[must_use]
+    /// Convert [`Extensions`] into a frozen/readonly variant
+    pub fn into_frozen_extensions(self) -> Arc<Self> {
```
GlenDC, line 97: not sure why the internals are exposed here. Is there a legit use case in even needing to know how it freezes?

I would think that something like

```
fn freeze(self) -> Extensions
```

is good enough and that it would return you a new extensions object with itself as the parent extensions?

rama-core/src/extensions.rs
```diff
@@ -143,18 +166,41 @@ impl Extensions {
         }
     }
 
-    /// Returns true if the `Extensions` contains the given type.
+    /// Returns true if the `Extensions` or parents contains the given type.
     #[must_use]
     pub fn contains<T: Send + Sync + 'static>(&self) -> bool {
+        if self.contains_without_chaining::<T>() {
```
GlenDC: why the if else? You can just chain them with `||`, no?

rama-core/src/extensions.rs
```diff
@@ -143,18 +166,41 @@ impl Extensions {
...
+            self.parent_extensions
+                .as_ref()
+                .map(|parent| parent.contains::<T>())
+                .unwrap_or_default()
+        }
+    }
+
+    /// Returns true if the `Extensions` contains the given type.
+    #[must_use]
+    pub fn contains_without_chaining<T: Send + Sync + 'static>(&self) -> bool {
```
GlenDC: feels like a design smell if one ever needs to care about this, I want in general chaining to be pretty much invisible to anything, is there a legit use case in this PR for it?

rama-core/src/extensions.rs
```diff
@@ -143,18 +166,41 @@ impl Extensions {
...
+        self.get_without_chaining().or_else(|| {
+            self.parent_extensions
+                .as_ref()
+                .and_then(|parent| parent.get())
+        })
+    }
+
+    /// Get a shared reference to a type previously inserted on this `Extensions`.
+    #[must_use]
+    pub fn get_without_chaining<T: Send + Sync + 'static>(&self) -> Option<&T> {
```
GlenDC: same question as the containss variant

rama-http-core/src/h2/proto/streams/recv.rs
GlenDC: Not sure it will be a Vec, but yeah ok I get what you mean

rama-http-core/src/service.rs
```diff
@@ -19,12 +21,23 @@ pub trait HttpService<ReqBody>: sealed::Sealed<ReqBody> {
...
+    pub fn new(svc: S) -> Self {
+        Self {
+            svc,
+            parent_extensions: None,
+        }
+    }
+
+    generate_set_and_with! {
+        /// Set the parent Extensions that will be applied by [`RamaHttpService`] on each [`Request`]
+        pub fn parent_extensions(mut self, frozen_extensions: Option<Arc<Extensions>>) -> Self {
```
GlenDC, line 37: Feels odd, why this has to be here? what's the use case?

tests/integration/ua_emulation.rs
```diff
@@ -259,6 +259,7 @@ async fn test_ua_emulation() {
         }
 
         async fn server_svc_fn(req: Request) -> Result<Response, Infallible> {
+            println!("extensions: {:?}", req.extensions());
```
GlenDC: even within tests best to use trace calls, not print statements

tests/integration/ua_emulation.rs
```diff
@@ -459,6 +463,7 @@ where
         let (client_socket, mut server_socket) = new_mock_sockets();
 
         if let Some(extensions) = req.extensions().get::<ServerExtensions>() {
+            println!("extensions for server: {:?}", extensions.0);
```
GlenDC: same same

GlenDC (changes requested):
Looks good in general, couple of small remarks, but huge improvement already. Looking forward to making extensions even better after this, we're getting there. The entire progress feels amazing. Let's grab some lunch once we have the entire redesign + extensions 2.0 work behind our back!

tests/integration/ua_emulation.rs
soundofspace: should have been removed, but missed it

tests/integration/ua_emulation.rs
soundofspace: same

rama-core/src/extensions.rs
soundofspace: Yes added these just in case, but since our last talk just forgot about them, but will already remove them

rama-core/src/extensions.rs
soundofspace: there are places such as in HttpConnector where request already has extensions but you want to configure the parent. Not exposes these building blocks make that kinda hard/hacky. Do you still feel this should be refactored or is it part of `vec<Extensions>`?

rama-http-core/src/service.rs
soundofspace: For places where you need to set it later and not on creation (eg HttpConnector)

rama-core/src/extensions.rs
GlenDC: Yes I saw that, but I don't understand the need. So far parent extensions were in my mind natural borders

Transport to Application, Request To Response. Retry Layer. That kind of stuff. that's why calling it parent can be a bit confusing and keeping it to frozen in public API is probably less confusing.

So what's the legit use case for what you're describing?

Also let's make sure to have a call before you start that work what you call Vec<Extennsions>. To make sure we're aligned.

soundofspace: Here we go from connector to request which feel like a natural border also?

Once we have vec i just see this as RequestPreConnectionExt -> TcpExt -> ... -> HttpConn ext -> Request Ext -> Response ext...

I can already introduce that concept here but it feels more hacky then this setup. Especially if you keep in mind that im planning todo vec very soon (maybe still this week)

(yes to that call)

GlenDC: > Here we go from connector to request which feel like a natural border also?

What?

rama-http-core/src/service.rs
GlenDC: Can you give me a less abstract example?

rama-core/src/extensions.rs
soundofspace: > "Yes I saw that, but I don't understand the need. So far parent extensions were in my mind natural borders"

> there are places such as in HttpConnector 

Here we go from connector to request, but request already has extensions so we just need to set parent directly

```
        let EstablishedClientConnection { mut req, mut conn } =
            self.inner.connect(req).await.map_err(Into::into)?;

        let parent_extensions = std::mem::take(conn.extensions_mut()).into_frozen_extensions();
        req.extensions_mut()
            .set_parent_extensions(parent_extensions);
 ```

GlenDC: Hmm I think the flow there is a bit weird...
Connection is established in function of request... Somehow that connection sits in between request and response, and the connection can also be shared between requests... Not entirely certain what to do with that, but feels a bit weird here...

soundofspace: Yea, I think the solution here is good enough. But for the vec approach we really want to figure this one out

PR #717, opened 2025-10-12
advanced redirect/rewrite/forward http svc/layer

### LINKED ISSUE #716: create http redirect service layer
We already have a network and proxy forward service. We have not yet an http one.

The goal of this new service is to make it easier to create a rule tree (including a single rule) to forward one url to another url (using a redirect resolution, e.g. status code 302).

Currently we already do this in some private services built with rama, but each time manually crafted using a custom service developed "from scratch". This has been done sufficiently amount of times, that clearly we can better make a service for it to make this easier.

Wildcard examples

* forward `www.example.com/*` to `example.com/*`
* forward `*.example.org/*` to `example.com/*`

We could however also go fancier and allow to use path parameters and also make those available... etc.
To be figured out...

This could then be used to create a forwarder, with as a fallback your regular router.

GlenDC:
Closes #716


rama-http/src/matcher/mod.rs
```diff
@@ -485,29 +486,46 @@ impl<Body> HttpMatcher<Body> {
...
+        self.and(Self::uri_wildcard(wc))
     }
 
     /// Create a [`UriMatcher`] matcher to match as an alternative to the existing set of [`HttpMatcher`] matchers.
     ///
     /// See [`UriMatcher`] for more information.
     #[must_use]
-    pub fn or_uri(self, re: impl AsRef<str>) -> Self {
-        self.or(Self::uri(re))
+    pub fn or_uri_wildcard(self, wc: Wildcard<'static>) -> Self {
```
GlenDC, line 535: missing or for regex

rama-http/src/service/redirect/mod.rs
```diff
@@ -6,6 +6,9 @@ use crate::{Response, header};
 use rama_core::Service;
 use std::{convert::Infallible, fmt, marker::PhantomData};
 
+mod r#static;
+pub use r#static::RedirectStatic;
+
 /// Service that redirects all requests.
 pub struct Redirect<ResBody> {
```
GlenDC: remove this one

Issue #686, opened 2025-08-27
add initial version of new `rama-tls-schannel` crate


GlenDC:
- client + server support
- only active on `target_os = "windows"`
  - and when so: also re-exported in `rama::tls` and option in `EasyHttpWebClient`
  - impl also other support if possible, e.g. ACME tls challenge

Not urgent at all. Can be picked up if someone has a lot of bandwidth for it and the stamina to see it all the way through without months of delay in between. But can also be left open for a long time. Not assigned to any milestone as it is mostly here just for me to get it out of my head without an actual desire to plan it in anytime soon.

Sponsors do feel free to reach out to me if you do have a need for it and we can prioritize it where possible.

PR #718, opened 2025-10-13
Implement octet stream response

### LINKED ISSUE #647: support application/octet-stream responses
To be added to https://ramaproxy.org/docs/rama/http/service/web/response/index.html

relevant rfcs:

* octet-stream: https://datatracker.ietf.org/doc/html/rfc2046#section-4.5.1
* byte ranges: https://datatracker.ietf.org/doc/html/rfc7233#section-2.1

The `application/octet-stream` MIME type is used in HTTP to represent arbitrary binary data. It's the default fallback when the server doesn’t know the specific content type of a file. When this type is returned, most clients (especially browsers) will treat the response as a downloadable file instead of attempting to render it. Servers often pair this with the `Content-Disposition: attachment` header to prompt a "Save as" dialog, optionally suggesting a filename. This content type carries no format-specific semantics — it's just raw bytes.

For large files or streaming scenarios, HTTP supports byte-range requests using the `Range` header. A client can request specific byte offsets (e.g. `Range: bytes=0-499`), and the server responds with `206 Partial Content` along with the `Content-Range` header. This mechanism enables download resumption, chunked streaming, and efficient media seeking. When multiple ranges are requested, the server can reply with a `multipart/byteranges` response. All of this works independently of the actual content type — including `application/octet-stream` — since range semantics are handled at the HTTP protocol level.

Reference implementations:

* https://docs.rs/axum-extra/latest/axum_extra/response/file_stream/struct.FileStream.html (byte range)
* https://docs.rs/axum-extra/latest/axum_extra/extract/multipart/index.html (multipart)

airycanon:
close #647

### DIFF
--- a/rama-http/src/service/web/endpoint/response/octet_stream.rs
+++ b/rama-http/src/service/web/endpoint/response/octet_stream.rs
@@ -143,7 +143,7 @@ impl<S> OctetStream<S> {
     ///
     /// # fn example() -> Result<(), rama_http_headers::Error> {
     /// // Serving first 5 bytes of "Hello, World!" (13 bytes total)
-    /// let partial_data = b"Hello".to_vec();
+    /// let partial_data = b"Hello";
     /// let cursor = std::io::Cursor::new(partial_data);
     /// let stream = ReaderStream::new(cursor);
     ///


rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,121 @@
...
+/// ```
+///
+/// ## Tuple struct syntax (simple use case)
+///
+/// ```
+/// use rama_http::service::web::response::{IntoResponse, OctetStream};
+///
+/// async fn handler() -> impl IntoResponse {
+///     let data = vec![0x48, 0x65, 0x6C, 0x6C, 0x6F];
+///     OctetStream(data)  // Equivalent to OctetStream::new(data)
```
GlenDC: this doesn't work, right? how would that even work? Anyway not a use case we need I think

rama-http-headers/src/common/content_disposition.rs
```diff
@@ -56,17 +56,15 @@ impl ContentDisposition {
         Self(HeaderValue::from_static("inline"))
     }
 
-    /*
     pub fn attachment(filename: &str) -> ContentDisposition {
-        let full = Bytes::from(format!("attachment; filename={}", filename));
-        match ::HeaderValue::from_maybe_shared(full) {
+        let full = format!("attachment; filename={}", filename);
```
GlenDC: make a trace call here and perhaps turn it into a regular attachment `HeaderValue` (static, "attachment"), as fallback, with the error being debug-traced

should rarely happen but at least it's a nice fallback over a panic

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,121 @@
...
+        Self {
+            data,
+            filename: None,
+        }
+    }
+
+    /// Create a new `OctetStream` response with `Content-Disposition: attachment` header.
+    ///
+    /// This will trigger a file download in browsers with the specified filename.
+    pub fn attachment(data: T, filename: impl Into<String>) -> Self {
```
GlenDC: I would rename this to `file_name` as octed-stream is treated as an attachment by default anyway.
And that is in the end what this is about. You can use rama's util macro for this to define in 1 defintion a `set_` and `with_` variant.


I would also just take a `String` for now as that makes it more clear an allocation is happening here.

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,121 @@
...
+        Self {
+            data,
+            filename: Some(filename.into()),
+        }
+    }
+}
```
GlenDC, line 297: I would also provide a `try_into_range_response(self, ...)` method here to consume it into a Response as a range, which would fail if we go out of range, but at least it would allow you to also return a response where we also define the `content-range` header.

airycanon:
@GlenDC Could you please take another look? If everything looks good, I'll continue implementing the ServeDir part.

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,174 @@
...
+        range: impl RangeBounds<u64>,
+        complete_length: u64,
+    ) -> Result<Response, Error>
+    where
+        T: Into<Body>,
+    {
+        let content_range =
+            ContentRange::bytes(range, complete_length).map_err(|_| Error::invalid())?;
+
+        let body = self.data.into();
```
GlenDC: Add a comment please that it is up to the creator of this `OctetStream` object that the `stream` is indeed of the specified range. It is not checked here and that is by design.

GlenDC:
> @GlenDC Could you please take another look? If everything looks good, I'll continue implementing the ServeDir part.

I would first try to get this PR merged in though and do any work regarding ServeDir in a follow up PR. 
Multipart would be also still another PR, but could also be picked up by someone else I suppose.

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,174 @@
...
+/// }
+/// ```
+///
+/// ## Download with filename
+///
+/// ```
+/// use rama_http::service::web::response::{IntoResponse, OctetStream};
+///
+/// async fn download() -> impl IntoResponse {
+///     let data = b"file contents".to_vec();
```
GlenDC:
```suggestion
///     let data = b"file contents";
```

Pretty certain this should work as well

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,174 @@
...
+///     OctetStream::new(data).with_file_name("data.bin".to_string())
+/// }
+/// ```
+///
+pub struct OctetStream<T> {
+    data: T,
```
GlenDC: This is. to be a `stream`, not data. Same same, just different name. But the trait bound below for it also has to change. More in other comment.

Next to that it is important to also add an optional `content_size: Option<usize>` property as not all streams will have this information. Next to that you can also provide a setter to optionally set the content-size.

Would also be nice if you can provide an alternative (extra) constructor to easily create an `OctetStream` from a file on disk, something like:

```
pub async fn try_from_path(path: impl AsRef<Path>) -> io::Result<OctetStream<ReaderStream<File>>>
```

This will provide the content_size and file name in case it is available from the metadata (and can be read, be graceful here)

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,174 @@
...
+        Self {
+            data: self.data.clone(),
+            filename: self.filename.clone(),
+        }
+    }
+}
+
+impl<T> IntoResponse for OctetStream<T>
+where
+    T: Into<Body>,
```
GlenDC: These trait bounds will instead have to be something like this:

```
T: TryStream<Ok: Into<Bytes>, Error: Into<BoxError>> + Send + 'static,
```

Haven't tested it, so might miss some bounds or have a mistake.

Btw when taking a generic "Stream" we usually out of convention us the letter `S` instead of `T`.

rama-http/src/service/web/endpoint/response/octet_stream.rs
GlenDC: You will no longer have a body (see other comment) so instead you create a body `from_stream`

rama-http/src/service/web/endpoint/response/octet_stream.rs
GlenDC: Also make a second version of it for range responses:

```
pub async fn try_range_response_from_path(path: impl AsRef<Path>, start: usize, end: usize) -> io::Result<Response>
```

examples/http_octet_stream.rs
```diff
@@ -0,0 +1,58 @@
...
+                    .get("/download", serve_download),
+            ),
+        )
+        .await
+        .expect("failed to run service");
+}
+
+/// Example 1: Simple binary response
+async fn serve_binary_data() -> impl IntoResponse {
+    let data = vec![0x48, 0x65, 0x6C, 0x6C, 0x6F]; // "Hello" in bytes
```
GlenDC:
```suggestion
    let data = b"Hello";
```

examples/http_octet_stream.rs
```diff
@@ -0,0 +1,58 @@
...
+    let cursor = std::io::Cursor::new(data);
+    let stream = ReaderStream::new(cursor);
+    OctetStream::new(stream)
+}
+
+/// Example 2: Binary download with Content-Disposition
+async fn serve_download() -> impl IntoResponse {
+    let data = b"Binary file content".to_vec();
```
GlenDC:
```suggestion
    let data = b"Binary file content";
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+
+/// An octet-stream response for serving arbitrary binary data.
+///
+/// Will automatically get `Content-Type: application/octet-stream`.
+///
+/// This is commonly used for:
+/// - Serving unknown or arbitrary binary files
+/// - Downloadable content that doesn't fit other MIME types
+/// - Raw binary data streams
+///
```
GlenDC, line 22:
```suggestion
///
/// User Agents often treat it as if the `Content-Disposition` header was set to attachment,
/// and propose a "Save As" dialog.
///
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+/// # Examples
+///
+/// ## Basic binary response
+///
+/// ```
+/// use rama_http::service::web::response::{IntoResponse, OctetStream};
+/// use rama_core::stream::io::ReaderStream;
+///
+/// async fn handler() -> impl IntoResponse {
+///     let data = vec![0x48, 0x65, 0x6C, 0x6C, 0x6F]; // "Hello"
```
GlenDC:
```suggestion
///     let data = b"Hello";
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+/// ```
+///
+/// ## Download with filename and size
+///
+/// ```
+/// use rama_http::service::web::response::{IntoResponse, OctetStream};
+/// use rama_core::stream::io::ReaderStream;
+///
+/// async fn download() -> impl IntoResponse {
+///     let data = b"file contents".to_vec();
```
GlenDC:
```suggestion
///     let data = b"file contents";
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+///
+/// ## Partial content (range request)
+///
+/// ```
+/// use rama_http::service::web::response::OctetStream;
+/// use rama_core::stream::io::ReaderStream;
+///
+/// # fn example() -> Result<(), rama_http_headers::Error> {
+/// // Serving first 5 bytes of "Hello, World!" (13 bytes total)
+/// let partial_data = b"Hello".to_vec();
```
GlenDC:
```suggestion
/// let partial_data = b"Hello";
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+        ///
+        /// This will trigger a file download in browsers with the specified filename.
+        pub fn file_name(mut self, filename: String) -> Self {
+            self.filename = Some(filename);
+            self
+        }
+    }
+
+    rama_utils::macros::generate_set_and_with! {
+        /// Set the content size for `Content-Length` header.
```
GlenDC:
```suggestion
        /// Set the content size for `Content-Length`- or as part of the `Content-Range` header.
```

rama-http/src/service/web/endpoint/response/octet_stream.rs
```diff
@@ -0,0 +1,476 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use rama_http::service::web::response::OctetStream;
+    /// use rama_core::stream::io::ReaderStream;
+    ///
+    /// # fn example() -> Result<(), rama_http_headers::Error> {
+    /// // Serving first 5 bytes of "Hello, World!" (13 bytes total)
+    /// let partial_data = b"Hello".to_vec();
```
GlenDC:
```suggestion
    /// let partial_data = b"Hello";
```

airycanon:
@GlenDC Hi, just saw your message—sorry for the slow reply. I’m happy to take on MultiPart and would appreciate your guidance. The only catch is I’m tied up with work at the moment and may not be able to finish within 1–2 weeks. If the timing’s flexible, I can start and provide regular updates. If not, maybe you take MultiPart and I’ll help on something else. Let me know what works best.

GlenDC:
> @GlenDC Hi, just saw your message—sorry for the slow reply. I’m happy to take on MultiPart and would appreciate your guidance. The only catch is I’m tied up with work at the moment and may not be able to finish within 1–2 weeks. If the timing’s flexible, I can start and provide regular updates. If not, maybe you take MultiPart and I’ll help on something else. Let me know what works best.

I'll take up the multipart issue myself, as there are still I'll have to fiddle with. It would be a frustrating experience for the other party (you) when delegated.

If you like some tls work that is non-prio and you are open for some work you pickup

- https://github.com/plabayo/rama/issues/687 if you are on linux
- or https://github.com/plabayo/rama/issues/686 if you are on windows

Pick one if you want by reacting on the specific issue.
But please do so only if you are willing to work on it on a regular base and also finish it.
It's not a time pressure issue so I'm fine it taking many weeks, but I would still prefer to see some progress every week.

First step (and can be only step you execute) would be to just make that separate crate, e.g. rama-tls-ktls, and implement the client connector and server acceptor. That with an example and test. Once we have that as baseline we can see what kind of configurations we can expose (step 2, later). And only much later can we do step (3) and add also an option for it to stuff like the `EasyHttpWebClient` to use ktls / schannel as connector.

PR #719, opened 2025-10-13
Issue 517/support akamai h2 passive fingerprint

### LINKED ISSUE #517: support akamai h2 passive fingerprint and expose in echo + fp services
Paper: <https://www.blackhat.com/docs/eu-17/materials/eu-17-Shuster-Passive-Fingerprinting-Of-HTTP2-Clients-wp.pdf>

Example implementation: <https://tls.peet.ws/>

Can be added to https://github.com/plabayo/rama/tree/main/rama-net/src/fingerprint with implementation in `./akamai/h2.rs`

You can look at <https://github.com/plabayo/rama/blob/main/rama-net/src/fingerprint/ja4/http.rs> for an example

In a file like <https://github.com/plabayo/rama/blob/main/src/cli/service/echo.rs> you should be able to find out how to get access to all required info.

Please also add unit tests with inline docs

sim-hash:
closes #517

Implement passive HTTP2 client fingerprinting using Akamai methodology.

  - Added rama-net/src/fingerprint/akamai/h2.rs with AkamaiH2 implementation
  - Integrated into FP service and echo service
  - Added test suite

### DIFF
--- a/rama-cli/src/cmd/fp/endpoints.rs
+++ b/rama-cli/src/cmd/fp/endpoints.rs
@@ -12,7 +12,7 @@ use rama::{
     error::{ErrorContext, OpaqueError},
     extensions::ExtensionsRef,
     http::{
-        Body, BodyExtractExt, Request, Response, StatusCode, Version,
+        BodyExtractExt, Request, Response, StatusCode, Version,
         proto::h2,
         service::web::{
             extract::Path,
--- a/rama-net/src/fingerprint/akamai/h2.rs
+++ b/rama-net/src/fingerprint/akamai/h2.rs
@@ -8,7 +8,6 @@
 //! - P = PRIORITY frames (stream_id:exclusive:depends_on:weight, comma-separated, or "0" if not present)
 //! - PS = Pseudo-header order (m=method, p=path, a=authority, s=scheme, comma-separated)
 
-use itertools::Itertools as _;
 use rama_core::extensions::Extensions;
 use rama_http_types::proto::h2::{
     PseudoHeader, PseudoHeaderOrder,
@@ -21,7 +20,7 @@ use rama_http_types::proto::h2::{
         StreamId,
     },
 };
-use std::fmt;
+use std::{fmt, io};
 
 #[derive(Clone)]
 pub struct AkamaiH2 {
@@ -56,9 +55,13 @@ impl AkamaiH2 {
         for frame in early_frames.iter() {
             match frame {
                 EarlyFrame::Settings(s) => {
-                    if !s.is_ack()
-                        && let Some(order) = &s.config.setting_order
-                    {
+                    if !s.is_ack() {
+                        let order = s
+                            .config
+                            .setting_order
+                            .as_ref()
+                            .ok_or(AkamaiH2ComputeError::MissingSettingOrder)?;
+
                         for setting_id in order.iter() {
                             let value = match setting_id {
                                 HeaderTableSize => s.config.header_table_size,
@@ -94,6 +97,10 @@ impl AkamaiH2 {
             }
         }
 
+        if settings.is_empty() {
+            return Err(AkamaiH2ComputeError::NoSettings);
+        }
+
         let pseudo_order = pseudo_header_order
             .iter()
             .filter_map(|ph| match ph {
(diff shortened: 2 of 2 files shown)


sim-hash:
@GlenDC Let me know if I missed anything regarding documentation. Also, does the profile from rama-ue require extension with akamai fingerprint?

rama-cli/src/cmd/fp/endpoints.rs
```diff
@@ -156,6 +156,18 @@ pub(super) async fn get_report(req: Request) -> Result<Html, Response> {
         })
     }
 
+    if parts.version == Version::HTTP_2
+        && let Some(akamai_h2) = get_akamai_h2_info(&parts.extensions)
+    {
+        tables.push(Table {
+            title: "🆔 Akamai H2".to_owned(),
+            rows: vec![
+                ("Akamai HTTP2 Client Fingerprint".to_owned(), akamai_h2.hash),
```
GlenDC:
```suggestion
                ("Akamai h2 Client Fingerprint".to_owned(), akamai_h2.hash),
```

rama-cli/src/cmd/fp/endpoints.rs
```diff
@@ -156,6 +156,18 @@ pub(super) async fn get_report(req: Request) -> Result<Html, Response> {
         })
     }
 
+    if parts.version == Version::HTTP_2
+        && let Some(akamai_h2) = get_akamai_h2_info(&parts.extensions)
+    {
+        tables.push(Table {
+            title: "🆔 Akamai H2".to_owned(),
```
GlenDC:
```suggestion
            title: "🆔 Akamai h2".to_owned(),
```

rama-cli/src/cmd/fp/endpoints.rs
```diff
@@ -536,6 +548,18 @@ pub(super) async fn form(req: Request) -> Result<Html, Response> {
...
```
GlenDC:
```suggestion
            title: "🆔 Akamai h2".to_owned(),
```

rama-cli/src/cmd/fp/endpoints.rs
```diff
@@ -536,6 +548,18 @@ pub(super) async fn form(req: Request) -> Result<Html, Response> {
...
+            rows: vec![
+                ("Akamai HTTP2 Client Fingerprint".to_owned(), akamai_h2.hash),
```
GlenDC:
```suggestion
                ("Akamai h2 Client Fingerprint".to_owned(), akamai_h2.hash),
```

src/cli/service/echo.rs
```diff
@@ -711,6 +724,7 @@ impl Service<Request> for EchoService {
                 "path": parts.uri.path().to_owned(),
                 "query": parts.uri.query().map(str::to_owned),
                 "h2": h2,
+                "akamai_h2": akamai_h2,
```
GlenDC: I would move the `akamai_h2` property as as property of the `h2` object, as you already have also the `h2` version check (`if parts.version == Version::HTTP_2 {`) there.

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
+//! Akamai HTTP2 Fingerprinting implementation for Rama.
+//!
+//! The fingerprint format is: `S[;]|WU|P[,]|PS[,]`
```
GlenDC:
```suggestion
//! The fingerprint format is: `S[;]|WU|P[,]#|PS[,]`
```

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        let mut settings = Vec::new();
+        let mut window_update = None;
+        let mut priority_frames = Vec::new();
+
+        for frame in early_frames.iter() {
+            match frame {
+                EarlyFrame::Settings(s) => {
+                    if !s.is_ack()
+                        && let Some(order) = &s.config.setting_order
```
GlenDC: would it not be an error if we do not have a setting order? As it means we can have no settings in case we do not have order

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+                    priority_frames.push(AkamaiPriorityFrame {
+                        stream_id: p.stream_id.into(),
+                        exclusive: p.dependency.is_exclusive,
+                        depends_on: p.dependency.dependency_id.into(),
+                        weight: p.dependency.weight,
+                    });
+                }
+            }
+        }
+
```
GlenDC, line 99: what if there are no settings, probably due to a mistake. Seems there is no placeholder value (like 0) specified for settings. Should we perhaps have a check here `settings.is_empty()` to return an error if so?

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+                .join(",")
+        };
+
+        let pseudo = self.pseudo_header_order.iter().join(",");
+
+        // Format: S[;]|WU|P[,]|PS[,]
+        let raw = format!("{settings}|{window_update}|{priority}|{pseudo}");
+
+        if as_hash {
+            write!(f, "{}", md5_hash(&raw))
```
GlenDC: look at `rama-net/src/fingerprint/ja3.rs`, there is a way to write directly into `fmt::Write` so we avoid double allocating here

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+    }
+
+    #[inline]
+    #[must_use]
+    pub fn to_human_string(&self) -> String {
+        format!("{self:#?}")
+    }
+
+    fn fmt_as(&self, f: &mut fmt::Formatter<'_>, as_hash: bool) -> fmt::Result {
+        let settings = self
```
GlenDC: Check `rama-net/src/fingerprint/ja3.rs`, you can avoid all these allocations.

Right now you allocate a lot of times

- several allocations per "part"
- another allocations to join them
- another allocations to hash
- another allocation to stringify the hash

Check `rama-net/src/fingerprint/ja3.rs` you'll see with a little macro magic you can easily avoid most of these allocations by writing directly into a hashed fmt writer, or fmt writer. Depending on display or debug.

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+        }
+    }
+}
+
+impl std::error::Error for AkamaiH2ComputeError {}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rama_http_types::proto::h2::{
```
GlenDC: you should need to import `rama_http_types::proto::h2::*` again, no? Given it is already imported in `super::*`, no?

rama-net/src/fingerprint/akamai/h2.rs
```diff
@@ -0,0 +1,496 @@
...
+            stream_id: StreamId::zero(),
+            size_increment: 12517377,
+        }));
+
+        ext.insert(create_early_frame_capture(&frames));
+
+        let akamai_h2 = AkamaiH2::compute(&ext).expect("compute akamai h2");
+
+        let debug_str = format!("{akamai_h2:?}");
+        assert_eq!(debug_str, "1:65536;2:0;4:131072;5:16384|12517377|0|m,p,a,s");
```
GlenDC, line 342: where did you find this example?

did you also use some of the examples from papers such as https://www.blackhat.com/docs/eu-17/materials/eu-17-Shuster-Passive-Fingerprinting-Of-HTTP2-Clients-wp.pdf btw? Would probably be a good idea to also include examples from such authoritative sources. That are easily verifiable.

rama-net/src/fingerprint/akamai/h2.rs
sim-hash: Yes and no. Some of them can be removed, others not. It seems like the parent doesn't include those.

rama-net/src/fingerprint/akamai/h2.rs
sim-hash: This e.g. is from my machine. Yes, I agree, will add them shortly.

PR #720, opened 2025-10-20
Issue 462/revisit service api (core) design in 0.3 milestone#4 state for http endpoints

soundofspace:
- Add state for `WebService` and `Router` by adding `IntoEndpointServiceWithState`. And providing this to our extractors
- Refactor FactoryTrait `EndpointServiceFn` with pattern T for any Function to use nested tuples so we have predictable signatures. Will add a paragraph to patterns chapter for this soon 
- Refactor `Arc<BoxService>` to `BoxService` since this is now uses `Arc` internally.
- Update examples to use this state

For separate PRs:
- Docs about pattern used in `EndpointServiceFn` to make predictable generics for function signatures
- State for service_fn. Once this is ready also I will expand the docs around state and explain all the different option for each place and the pros and cons

part of #462

### DIFF
--- a/rama-macros/src/from_ref_macro/mod.rs
+++ b/rama-macros/src/from_ref_macro/mod.rs
@@ -29,45 +29,11 @@ where
     match result {
         Ok(tokens) => {
             let tokens = (quote! { #tokens }).into();
-            if std::env::var_os("AXUM_MACROS_DEBUG").is_some() {
+            if std::env::var_os("RAMA_MACROS_DEBUG").is_some() {
                 eprintln!("{tokens}");
             }
             tokens
         }
         Err(err) => err.into_compile_error().into(),
     }
-}
-
-#[cfg(test)]
-fn run_ui_tests(directory: &str) {
-    #[rustversion::nightly]
-    fn go(directory: &str) {
-        let t = trybuild::TestCases::new();
-
-        if let Ok(mut path) = std::env::var("AXUM_TEST_ONLY") {
-            if let Some(path_without_prefix) = path.strip_prefix("axum-macros/") {
-                path = path_without_prefix.to_owned();
-            }
-
-            if !path.contains(&format!("/{directory}/")) {
-                return;
-            }
-
-            if path.contains("/fail/") {
-                t.compile_fail(path);
-            } else if path.contains("/pass/") {
-                t.pass(path);
-            } else {
-                panic!()
-            }
-        } else {
-            t.compile_fail(format!("tests/{directory}/fail/*.rs"));
-            t.pass(format!("tests/{directory}/pass/*.rs"));
-        }
-    }
-
-    #[rustversion::not(nightly)]
-    fn go(_directory: &str) {}
-
-    go(directory);
 }
--- a/rama-macros/src/from_ref_macro/from_ref.rs
+++ b/rama-macros/src/from_ref_macro/from_ref.rs
@@ -102,11 +102,6 @@ impl Combine for FieldAttrs {
     }
 }
 
-#[test]
-fn ui() {
-    super::run_ui_tests("from_ref");
-}
-
 fn support_root_ts() -> proc_macro2::TokenStream {
     // Prefer the umbrella crate
     if let Ok(found) = crate_name("rama") {


rama-http/src/service/web/endpoint/service.rs
```diff
@@ -3,34 +3,65 @@ use super::extract::{FromRequest, FromRequestContextRefPair};
 use crate::{Request, Response};
 use rama_utils::macros::all_the_tuples_no_last_special_case;
 
+#[derive(Clone, Debug)]
+pub struct State<S>(pub S);
```
soundofspace: We will have something very similar for service_fn, should we expose a different `State` there or should we move this wrapper struct into rama core and reuse it?

examples/http_anti_bot_infinite_resource.rs
```diff
@@ -183,20 +202,14 @@ where
...
             .peer_addr()
             .ip();
-        let block_list = stream
-            .extensions()
-            .get::<State>()
-            .unwrap()
-            .block_list
-            .lock()
-            .await;
+        let block_list = self.block_list.lock().await;
```
soundofspace, line 215: Having state in handlers for endpoints and the likes makes sense, but i believe for services it's mostly a bad pattern, so I refactored this example into what I consider better practice. Do you agree here? (We could also store state into IpService, but dont think we want extensions for this)

GlenDC: Good for me

rama-http/src/service/web/endpoint/service.rs
GlenDC: What does it (`State` wrapper)  do exactly?

soundofspace: To expose this in a function signature you need a wrapper to be able to recognize a parameter is state. Having the raw generic there is not possible so we wrap it with State<> instead. So the only goal is to expose it function signatures, nothing really else

GlenDC: And that's something we will also support for service_fn? If so, fine by me to move it there I guess

GlenDC (approved):
LGTM.

I'm ok with it being a separate PR if you want, but seems small enough that you might as well do it already. Now that you (re)introduce the FromRef capabilities you might as well also add the proc macro with the same macro to `rama-macros` and re-exported in same location as the trait (traits and proc macros live in separate layers and can share the same name, should you not know that yet).

It basically works like any other derive trait where it implements that for any property by implementing FromRef for each property, as you now are doing that anyway manually already in your PR's example updates, which seems a bit silly given how mechanical in nature it most of the times is.

examples/http_key_value_store.rs
```diff
@@ -84,9 +83,31 @@ use tracing_subscriber::layer::SubscriberExt;
...
+
+impl_deref!(Db: Arc<RwLock<HashMap<String, bytes::Bytes>>>);
+
+// AppState is cloneable so implement `FromRef<AppState>` (by deriving it) is enough. But here
+// we decided that our GlobalState will be `Arc<AppState>`. We do this so we can
+// add more fields to AppState but cloning it stays the same size. For this
+// specific use case this is way too overkill, but we do it do demonstrate that it's possible,
+// by manually implementing `FromRef`
+
+impl FromRef<Arc<AppState>> for Db {
```
GlenDC: Somehow It feels like a blanket impl for something like this should be possible

rama-macros/src/from_ref_macro/attr_parsing.rs
```diff
@@ -0,0 +1,32 @@
+use quote::ToTokens;
```
GlenDC, line 1: s/axum_macros/from_ref_macro/ plox

examples/http_key_value_store.rs
soundofspace: Haha exactly what i thought. But has conflicting implementation with the one for clone. So not really possible unless we remove that one...

GlenDC: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=139b8ff430e4e535ee3a43495b674b2b

seems there is some dark magic possible, probably an even cleaner solution is possible, but does seems like we can handle it somehow

soundofspace: We can but I don't we should. For this to work we need an extra generic as seen in your example. 

- First issue: collisions, this should be fine actually as long as people don't implement AsRef twice and both times to Sized types -> So not really an issue
- Second issue: To make this work with closures we don't only need the extra generic in `FromRef` but also in `EndpointServiceFn`, `EndpointServiceFnWrapper`, `IntoEndpointServiceWithState`, the Sealed traits... 

Having this generic leaking all over, just to add only a little bit of convenience is not worth it believe

GlenDC: Agreed, still, I would think the usage of `Arc<State>` is extremely common. Which does make me feel it should somehow be easier to use the `State` as is... from it...  But not sure what the answer is here :P Perhaps the FromRef derive macro should also implement it for `Arc` for each property :P dunno Including for the type itself.. no clue

soundofspace: Yea not sure also. Having the FromRef macro implement it for Arc is also tricky since something like this is not allowed because we neither own the trait or types.
```rust
impl FromRef<Arc<AppState>> for String
```
While on the other hand this is allowed:
```rust
FromRef<Arc<AppState>> for Db
```
So it really depends on where this happens / if new_type pattern is used

GlenDC: hmmm ok... so what do you propose @soundofspace ? as I can keep proposing but seems I am getting noewhere

soundofspace: I'm playing with this a bit, but so far I'm always getting stuck somewhere, or arriving at a solution with more downsides then upsides. Will continue playing for a bit more, but otherwise I would propose removing `Arc<AppState>`, and just always use `AppState` with `Arcs` inside for use cases likes this (it's also what axum is doing).

GlenDC: Yeah I think that's better anyway, similar to extensions you probably want state properties to be cheap to clone (or arc). It's also true that some things you want/need to arc, while others not. So just slapping an arc around entire state is probably anyway a lazy/bad approach, so might not be something we ever want to show in examples as it might give the wrong example

GlenDC: If you agree @soundofspace can you remove the Arc usage, and then let me know that this is up for review. At that point I can probably do the final review.

soundofspace: Yea agree, ready for review again

rama-macros/src/from_ref_macro/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+pub(crate) fn expand<T>(result: syn::Result<T>) -> TokenStream
+where
+    T: ToTokens,
+{
+    match result {
+        Ok(tokens) => {
+            let tokens = (quote! { #tokens }).into();
+            if std::env::var_os("AXUM_MACROS_DEBUG").is_some() {
```
GlenDC:
```suggestion
            if std::env::var_os("RAMA_MACROS_DEBUG").is_some() {
```

rama-macros/src/from_ref_macro/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+    }
+}
+
+#[cfg(test)]
+fn run_ui_tests(directory: &str) {
+    #[rustversion::nightly]
+    fn go(directory: &str) {
+        let t = trybuild::TestCases::new();
+
+        if let Ok(mut path) = std::env::var("AXUM_TEST_ONLY") {
```
GlenDC: all this needs to be updated as well, is it used?

GlenDC (approved):
Approved. Can be merged once you resolve the last remaining feedback.

PR #723, opened 2025-10-29
Issue 462/revisit service api (core) design in 0.3 milestone#6 extensions as breadcrumbs

soundofspace:
What this does:
- New implementation for breadcrumb extensions. Extensions can only be appended, and they are readonly once added. So they leave breadcrumbs of information about the entire flow
- Require Debug for all extensions
- Always transfer connection extensions to request so we have complete extension chain

What this doesn't do:
- Optimize extensions. Current data structure has huge potential for improvement, but left this out of scope so we can focus on developer experience first, and do that as hidden internal followup. Depending on how you see this @GlenDC can be part of 0.3 or 0.4 (if we see that as mainly performance improvements)
- Migrate builders to a new system is that is "breadcrumb aware". Eg dont clone and modify, but something like reduce, or similar

Part of #462

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2885,7 +2885,6 @@ dependencies = [
  "rcgen",
  "rustls",
  "rustls-native-certs",
- "rustls-pemfile",
  "rustls-pki-types",
  "tokio",
  "tokio-rustls",
@@ -3226,15 +3225,6 @@ dependencies = [
  "rustls-pki-types",
  "schannel",
  "security-framework",
-]
-
-[[package]]
-name = "rustls-pemfile"
-version = "2.2.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dce314e5fee3f39953d46bb63bb8a46d40c2f8fb7cc5a3b6cab2bde9721d6e50"
-dependencies = [
- "rustls-pki-types",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -195,7 +195,6 @@ rustls = { version = "0.23", default-features = false, features = [
     "aws_lc_rs",
 ] }
 rustls-native-certs = "0.8"
-rustls-pemfile = "2.2"
 rustls-pki-types = "^1"
 rustversion = "1.0"
 schannel = "0.1"
--- a/docs/thirdparty/fork/README.md
+++ b/docs/thirdparty/fork/README.md
@@ -121,6 +121,12 @@ but still directly in function of Rama.
     - Original: <https://github.com/Michael-F-Bryan/include_dir/blob/d742c6fffce99ee89da91b934e7ce6fb2a82680c/LICENSE>
     - Type: MIT
     - Copy: [./licenses/include_dir](./licenses/include_dir)
+- <https://github.com/rustls/pemfile/tree/1e4ce69dd97b6bd0657a9dee8503cb8657b49235>
+  - archived, no longer maintained
+  - License:
+    - Original: <https://github.com/rustls/pemfile/blob/1e4ce69dd97b6bd0657a9dee8503cb8657b49235/LICENSE-MIT>
+    - Type: MIT
+    - Copy: [./licenses/pemfile](./licenses/pemfile)
 
 ## Small Lib Forks
 
--- a/docs/thirdparty/fork/licenses/pemfile
+++ b/docs/thirdparty/fork/licenses/pemfile
@@ -1,0 +1,25 @@
+Copyright (c) 2016 Joseph Birr-Pixton <jpixton@gmail.com>
+
+Permission is hereby granted, free of charge, to any
+person obtaining a copy of this software and associated
+documentation files (the "Software"), to deal in the
+Software without restriction, including without
+limitation the rights to use, copy, modify, merge,
+publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software
+is furnished to do so, subject to the following
+conditions:
+
+The above copyright notice and this permission notice
+shall be included in all copies or substantial portions
+of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
+SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR
+IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
--- a/examples/tls_rustls_dynamic_certs.rs
+++ b/examples/tls_rustls_dynamic_certs.rs
@@ -71,13 +71,11 @@ use rama::{
         subscriber::{EnvFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt},
     },
     tls::rustls::{
-        dep::{
-            pemfile,
-            rustls::{
-                ALL_VERSIONS, ServerConfig, crypto::aws_lc_rs, server::ResolvesServerCert,
-                sign::CertifiedKey,
-            },
+        dep::rustls::{
+            ALL_VERSIONS, ServerConfig, crypto::aws_lc_rs, server::ResolvesServerCert,
+            sign::CertifiedKey,
         },
+        pemfile,
         server::{TlsAcceptorDataBuilder, TlsAcceptorLayer},
     },
 };
--- a/examples/tls_rustls_dynamic_config.rs
+++ b/examples/tls_rustls_dynamic_config.rs
@@ -64,10 +64,8 @@ use rama::{
         subscriber::{EnvFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt},
     },
     tls::rustls::{
-        dep::{
-            pemfile,
-            pki_types::{CertificateDer, PrivateKeyDer},
-        },
+        dep::pki_types::{CertificateDer, PrivateKeyDer},
+        pemfile,
         server::DynamicConfigProvider,
         server::{TlsAcceptorDataBuilder, TlsAcceptorLayer},
     },
--- a/rama-tls-rustls/Cargo.toml
+++ b/rama-tls-rustls/Cargo.toml
@@ -30,7 +30,6 @@ rama-utils = { workspace = true }
 rcgen = { workspace = true }
 rustls = { workspace = true }
 rustls-native-certs = { workspace = true }
-rustls-pemfile = { workspace = true }
 rustls-pki-types = { workspace = true }
 tokio = { workspace = true, features = ["macros", "io-std"] }
 tokio-rustls = { workspace = true }
--- a/rama-tls-rustls/src/lib.rs
+++ b/rama-tls-rustls/src/lib.rs
@@ -33,6 +33,8 @@ pub mod verify;
 
 pub mod key_log;
 
+pub mod pemfile;
+
 mod type_conversion;
 
 pub mod types {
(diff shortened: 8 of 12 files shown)


rama-http/src/layer/forwarded/get_forwarded.rs
```diff
@@ -262,14 +262,22 @@ where
         }
 
         if !forwarded_elements.is_empty() {
-            if let Some(ref mut f) = req.extensions_mut().get_mut::<Forwarded>() {
-                f.extend(forwarded_elements);
+            // TODO why was this Forwarded header instead of rama_net Forwarded, @glen was this mistake before?
```
soundofspace: @GlenDC

rama-core/src/extensions.rs
```diff
@@ -27,206 +27,88 @@
...
 /// extra data derived from the underlying protocol.
 #[derive(Clone, Default)]
 pub struct Extensions {
-    // If extensions are never used, no need to carry around an empty HashMap.
-    // That's 3 words. Instead, this is only 1 word.
-    map: Option<Box<AnyMap>>,
-    #[cfg(debug_assertions)]
-    type_map: Option<Box<TypeIdMap>>,
-    parent_extensions: Option<Arc<Extensions>>,
+    // TODO optimize this storage
```
GlenDC: What do you want to optimize? Would help to expanda bit more on these kind of comments. E.g. the problem statement, the goal, the ideas, ...

soundofspace: Still needed to create the issue and link to it (fyi https://github.com/plabayo/rama/issues/746)

soundofspace:
> Sometimes I had the instinct to deny the idea but honestly this is a lot simpler for good reasons. E.g. I was thinking why removing parent, you still want to freeze etc, no?? But no if you make it read-only you no longer care, so that's cool. Love it.

Had the exact same instinct in the beginning (but performance, but easy &mut, ...), but after working with this and seeing how it simplifies so much, I don't anymore. Also simplifying this allows us to optimize some other things quite aggressively since we can now make a lot of assumptions about this

rama-http/src/layer/forwarded/get_forwarded.rs
GlenDC: Probably a mistake indeed.... But do double check if not already

examples/socks5_connect_proxy_over_tls.rs
```diff
@@ -100,13 +100,16 @@ async fn main() {
...
-        conn: http_service,
+        mut req,
+        conn: mut http_service,
     } = client
         .connect(request)
         .await
         .expect("establish a proxied connection ready to make http requests");
 
+    req.extensions_mut()
+        .extend(std::mem::take(http_service.extensions_mut()));
```
GlenDC: This still feels like a code smell though... you are essentially erasing information here... I wonder if we can even prevent that...

Can you not just extend it without `std::mem::take`? Because what if that contain info that influences the `serve` stack of http sevice on purpose?? info would be gone now... In the spirit of readonly etc I feel this is very much a move in the opposite direction

rama-core/src/extensions.rs
```diff
@@ -27,224 +27,88 @@
...
-            let map = self.type_map.get_or_insert_with(Box::default);
-            #[allow(clippy::useless_conversion)]
-            map.extend(other_map.into_iter());
-        }
-
-        if let Some(other_map) = other.map {
-            let map = self.map.get_or_insert_with(Box::default);
-            #[allow(clippy::useless_conversion)]
-            map.extend(other_map.into_iter());
+    pub fn maybe_insert<T>(&mut self, val: Option<T>) -> &mut Self
```
GlenDC: is this really that useful? is it that common than that it warrants it own function? Given that you can just as easily do `if let Some(x) { ext.insert(x) }` ? But I do not recall ever needing that? What am I missing

rama-core/src/extensions.rs
```diff
@@ -27,224 +27,88 @@
...
-        self.map
-            .as_mut()
-            .and_then(|map| map.get_mut(&TypeId::of::<T>()))
-            .and_then(|boxed| (**boxed).as_any_mut().downcast_mut())
     }
 }
 
 impl fmt::Debug for Extensions {
-    #[cfg(debug_assertions)]
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
```
GlenDC: Can we not do derive(Debug) now? Seems that you are anyway doing exactly what that derive would do?

rama-core/src/extensions.rs
```diff
@@ -27,224 +27,88 @@
...
-                    .map(|parent| parent.contains::<T>())
-                    .unwrap_or_default()
-            })
+        let type_id = TypeId::of::<T>();
+        self.extensions.iter().rev().any(|item| item.0 == type_id)
     }
 
     /// Get a shared reference to a type previously inserted on this `Extensions` or any of the parents
     #[must_use]
     pub fn get<T: Send + Sync + 'static>(&self) -> Option<&T> {
```
GlenDC, line 84: Why not also already functions `first` and `iter`, seems like a small enough thing to do and easy to test? Feels a bit little to do a in a separate PR, or what is your idea around it.

rama-http-backend/src/client/svc.rs
```diff
@@ -47,10 +47,7 @@ where
     type Response = Response;
     type Error = BoxError;
 
-    async fn serve(&self, mut req: Request<BodyIn>) -> Result<Self::Response, Self::Error> {
-        req.extensions_mut()
-            .set_parent_extensions(Arc::new(self.extensions.clone()));
```
GlenDC, line 48: Seems extensions is still in `HttpClientService` yet it is not used afaik? what is the goal with that?

rama-http-backend/src/server/layer/upgrade/service.rs
```diff
@@ -120,9 +119,7 @@ where
                         async move {
                             match rama_http::io::upgrade::handle_upgrade(&req).await {
                                 Ok(mut upgraded) => {
-                                    upgraded
-                                        .extensions_mut()
-                                        .set_parent_extensions(Arc::new(req.extensions().clone()));
+                                    upgraded.extensions_mut().extend(req.extensions().clone());
```
GlenDC, line 122:
```suggestion
                                    upgraded.extensions_mut().extend(req.extensions());
```

shouldn't this just work without having to clone, given you will iterate internally and extend? Feels a bit pointless to have to clone the vec first as well? Or what am I missing?

rama-http/src/layer/forwarded/get_forwarded.rs
GlenDC: You didn't do anything more with this? Or you consider that out of scope?

rama-http/src/layer/forwarded/get_forwarded.rs
```diff
@@ -262,14 +262,22 @@ where
...
+                .cloned()
+            {
+                forwarded.extend(forwarded_elements);
+                forwarded
             } else {
                 let mut it = forwarded_elements.into_iter();
+                // TODO why was this
                 let mut forwarded = rama_net::forwarded::Forwarded::new(it.next().unwrap());
                 forwarded.extend(it);
-                req.extensions_mut().insert(forwarded);
```
GlenDC: do you mean the insert again in req? or what? If so probably a mistake or something that we needed in past due to taking it out of req but no longer? Please respond and action

rama-tcp/src/client/request.rs
GlenDC, line 22: out of scope for this pR probably but I think we can move this to rama-net and remove http_version. Not sure what the use case is but that would probably be more something for extension anyway. Because (a) doesn't seem specific to tcp and (b) feels weird to have somethhing http specific in there, can you also think about this and perhaps open an issue about it?

rama-ws/src/handshake/server.rs
```diff
@@ -926,8 +926,7 @@ impl Service<upgrade::Upgraded> for WebSocketEchoService {
         };
 
         // TODO improve extensions handling in AsyncWebSocket
```
GlenDC, line 939: Is that your comment from past or mine? I wonder what this is about...

rama-ws/src/handshake/server.rs
```diff
@@ -926,8 +926,7 @@ impl Service<upgrade::Upgraded> for WebSocketEchoService {
...
-        let parent_extensions = io.extensions().clone().into_frozen_extensions();
-        let extensions = Extensions::new().with_parent_extensions(parent_extensions);
+        let extensions = io.extensions().clone();
         let mut socket = AsyncWebSocket::from_raw_socket(io, Role::Server, maybe_ws_config).await;
```
GlenDC, line 941: perhaps make a variant `from_raw_sockets_with_extensions`, current approach feels weird

tests/integration/examples/example_tests/haproxy_client_ip.rs
```diff
@@ -92,13 +95,16 @@ async fn test_server_with_haproxy_v2() {
...
-        conn: http_service,
+        mut req,
+        conn: mut http_service,
     } = client
         .connect(request)
         .await
         .expect("establish a connection to the http server using haproxy v2");
 
+    req.extensions_mut()
+        .extend(std::mem::take(http_service.extensions_mut()));
```
GlenDC: again, unless I miss something I think the `std::mem::take` is an anti-pattern, it's essentailly the same as `clone + clear` and you removed clear for a good reason, no?

GlenDC: Perhaps do not implement `Default` for `new` and put an `allow` to override the lint... This would no longer make `take()` work and you would instead need to use swap/replace from `std::mem`, which we cannot prevent but it would at least not be as easy as `take` :P dunno, just doens't seem like a good idea ever. Either that or it is a good idea and than you might as well allow clear :P

GlenDC (approved):
Looks good, as such I approve. I do have some important remarks that I want to be answered, discussed or actioned before merging. That said, unless you change something big we can probably merge after that :)

rama-core/src/extensions.rs
soundofspace: Nothing I just migrated this, it was already here, but agree that it should be removed

examples/socks5_connect_proxy_over_tls.rs
soundofspace: Agree, was doing this avoid some clones, but can indeed have negative side effects so refactored that, we can always revise this in 0.4 if needed

rama-core/src/extensions.rs
soundofspace: yes forgot to

rama-core/src/extensions.rs
soundofspace: This pr was ready before that discussion and forgot to add them here already, but good remark

rama-http-backend/src/client/svc.rs
soundofspace: HttpClientService can be seen as a connector, and all connector have extensions, so it is used to implement Extensions traits.

It is also used in places where we clone extensions from connectors to requests. But cloning them is done in places that do the final connection setup and not inside the service itself to prevent duplicate applies

rama-ws/src/handshake/server.rs
soundofspace: Mine, it works, but it feels error prone how it is implemented now

soundofspace: > perhaps make a variant from_raw_sockets_with_extensions, current approach feels weird

This is exactly what I also mean with "improve it"

tests/integration/examples/example_tests/haproxy_client_ip.rs
soundofspace: Dont think we have to make it impossible to do, it is enough for us not todo it. Refactored everything to clone instead

rama-http-backend/src/server/layer/upgrade/service.rs
soundofspace: If we do it like this we can use the native extend from vec internally, which uses IntoIterator which is heavily optimized + only one clone of the vec. Otherwise we will have to iterate ourselves and clone each item which is probably slower, but more importantly hides a lot of cloning

rama-http/src/layer/forwarded/get_forwarded.rs
soundofspace: This was indeed a mistake before, so this fixed it

rama-http/src/layer/forwarded/get_forwarded.rs
soundofspace: We still insert it, but instead of inserting in this else branch, we now insert after the if-else branch

rama-ws/src/handshake/server.rs
soundofspace: I'm currently working with websocket a lot, and I am planning to give a detailed look at websocket after this PR solely focused on that, does that work for you?

GlenDC: sure

rama-ws/src/handshake/server.rs
GlenDC: I assume you will look at comments like this in your upcoming WS work

PR #730, opened 2025-11-05
Use connectors instead of jit req inspector

soundofspace:
- Jit request inspectors also need access to connection to access extensions, and at this point, it's just a connector. So remove JIT logic and just use normal connectors instead
- Layers to adapt Request and Response version
- Modify easy web client to use this new logic

part of #462

### DIFF
--- a/rama-tls-boring/src/client/connector.rs
+++ b/rama-tls-boring/src/client/connector.rs
@@ -390,6 +390,7 @@ fn set_target_http_version(
 
 impl<S, K> TlsConnector<S, K> {
     fn connector_data(&self, extensions: &mut Extensions) -> Result<TlsConnectorData, OpaqueError> {
+        #[cfg(feature = "http")]
         let target_version = extensions
             .get::<TargetHttpVersion>()
             .map(|version| ApplicationProtocol::try_from(version.0))
@@ -412,6 +413,7 @@ impl<S, K> TlsConnector<S, K> {
             builder.prepend_base_config(base_builder);
         }
 
+        #[cfg(feature = "http")]
         if let Some(target_version) = target_version {
             builder.try_set_rama_alpn_protos(&[target_version])?;
         }
--- a/rama-tls-rustls/src/client/connector.rs
+++ b/rama-tls-rustls/src/client/connector.rs
@@ -4,7 +4,7 @@ use crate::types::TlsTunnel;
 use rama_core::conversion::{RamaInto, RamaTryFrom};
 use rama_core::error::ErrorContext;
 use rama_core::error::{BoxError, ErrorExt, OpaqueError};
-use rama_core::extensions::{Extensions, ExtensionsMut, ExtensionsRef};
+use rama_core::extensions::{ExtensionsMut, ExtensionsRef};
 use rama_core::stream::Stream;
 use rama_core::telemetry::tracing;
 use rama_core::{Layer, Service};
@@ -16,7 +16,10 @@ use rama_net::transport::TryRefIntoTransportContext;
 use std::fmt;
 
 #[cfg(feature = "http")]
-use rama_http_types::{Version, conn::TargetHttpVersion};
+use ::{
+    rama_core::extensions::Extensions,
+    rama_http_types::{Version, conn::TargetHttpVersion},
+};
 
 /// A [`Layer`] which wraps the given service with a [`TlsConnector`].
 ///


examples/http_mitm_proxy_boring.rs
```diff
@@ -360,6 +361,7 @@ where
 
     let (parts, body) = req.into_parts();
     let parts_copy = parts.clone();
+    println!("incomming all extensions: {:?}", parts_copy.extensions);
```
GlenDC:
```suggestion
    tracing::debug!("incomming all extensions: {:?}", parts_copy.extensions);
```

examples/http_mitm_proxy_boring.rs
```diff
@@ -453,15 +455,16 @@ where
...
+                let socket = AsyncWebSocket::from_raw_socket(
                     upgraded,
                     Role::Server,
                     Some(ingress_socket_cfg),
                 )
                 .await;
-                socket
-                    .extensions_mut()
-                    .set_parent_extensions(Arc::new(request.extensions().clone()));
+                // TODO in a place like this we dont really want to extend but instead prepend
```
GlenDC, line 463: Can you clarify? I suppose it does not happen for now, meaning we we do not have these extensions for now? Or is it already done automatically in "handle_upgrade" into upgraded?

rama-tls-rustls/Cargo.toml
```diff
@@ -23,6 +23,7 @@ default = []
 [dependencies]
 pin-project-lite = { workspace = true }
 rama-core = { workspace = true }
+rama-http-types = { workspace = true }
```
GlenDC: Seems we already were enabling rama-http-types via rama-net/http before, but ... why is this required?

rama-tls* crates should probably also be able to work without http as there is plenty of use case of tls without http. Can we put a feature flag (http) here to enable this optionally? What stops us?

src/http/client/builder.rs
```diff
@@ -415,115 +402,48 @@ impl<T> EasyHttpWebClientBuilder<T, ProxyStage> {
...
-    ) -> EasyHttpWebClientBuilder<HttpConnector<T, I, I2>, HttpStage> {
-        EasyHttpWebClientBuilder {
-            connector: self.connector.with_jit_req_inspector(http_req_inspector),
-            _phantom: PhantomData,
-        }
-    }
+impl<T> EasyHttpWebClientBuilder<T, TlsStage> {
+    /// Add http support to this connector
+    pub fn with_http(self) -> EasyHttpWebClientBuilder<HttpConnector<T>, HttpStage> {
+        let connector = HttpConnector::new(self.connector);
```
GlenDC, line 418: I am a bit confused. Why do we need to explicitly need to add the http connector? Isn't that kinda assumed to be there using the default one if you do not set a custom one??? It's an `EasyHttpWebClientBuilder`, http is in the name? What's the value in forcing this to be always be called? Especially with such a vague name.

If your point is in the spirit of being explicit, and have it be in spirit of the custom one, than perhaps cann it `with_default_http_connector`? As `with_http` is super weird to read in the usage code. I think.

src/http/client/builder.rs
```diff
@@ -415,115 +402,48 @@ impl<T> EasyHttpWebClientBuilder<T, ProxyStage> {
...
+            connector,
             _phantom: PhantomData,
         }
     }
+}
 
+impl<T, I> EasyHttpWebClientBuilder<HttpConnector<T, I>, HttpStage> {
     /// Add a http request inspector that will run just before doing the actual http request
-    pub fn with_svc_req_inspector<I>(
+    pub fn with_svc_req_inspector<X>(
```
GlenDC: Err.... Postbus X? To young I guess.

Anyway... I would call this `U`, in spirit of the alphabet etc.

tests/integration/ua_emulation.rs
```diff
@@ -303,22 +303,19 @@ async fn test_ua_emulation() {
...
-                        .extensions_mut()
-                        .insert_mut(TlsConnectorDataBuilder::default()),
-                };
+                let mut builder = req
+                    .extensions_mut()
+                    .get::<TlsConnectorDataBuilder>()
+                    .cloned()
+                    .unwrap_or_default();
                 builder.set_server_verify_mode(ServerVerifyMode::Disable);
+                req.extensions_mut().insert(builder);
```
GlenDC, line 312: Why do you insert the builder again, if it already comes from the req??? I mean you clone it just to modify it and put it back??? What am I missing here, would you not just either get it or make a new one and inert, but in case exist just modify the existing one?

examples/http_mitm_proxy_boring.rs
soundofspace: This should have been removed, my mistake

src/http/client/builder.rs
soundofspace: Will rename it to `with_default_http_connector`, much better indeed. 

I added this as a manual step since this now gives you the ability to customize each layer, but to also use `with_custom_connector` at each layer. Eg insert connectors between tls-http. Before tls also added http which made that impossible. I thought about adding another way to add tls without http auto, but at that point it seemed more clear and kiss to just continue the pattern of adding each network layer

src/http/client/builder.rs
soundofspace: Good to know, I have habit of take a generic for use cases like this that is "far" from the normal ones so it's really in your phase, but if alphabet is preferred/default I will follow that

tests/integration/ua_emulation.rs
soundofspace: You will just have to kind of ignore this specific detail here. This is actually a glimps of the "breadcrumbs extensions". It's here because all these recent PRs were actually all part of that one. But I split them up to make it easier to review and discuss only one or two things at the same time, makes it much easier for every. If you want I can remove it here, but next PR will add this back

GlenDC: Hmm ok. Suspicious. Leave it in for now than. I'll hammer harder on it in next PR than.

examples/http_mitm_proxy_boring.rs
soundofspace: Something i will handle in next PR, i placed this todo here so i dont forget it. But its indeed something we dont use right now

soundofspace:
> Any concerns that aren't immediately clear from the PR? E.g. things that you have in your mind after having gone through all this code?

Something I'm still thinking about but I'm not entirely sure about `http_req_inspector_svc` I wonder if these should also just be normal services/layers. It feels like http-backend service is doing way too much. Removing JIT was actually a huge pleasure and in my opinion made everything easier

GlenDC:
> > Any concerns that aren't immediately clear from the PR? E.g. things that you have in your mind after having gone through all this code?
> 
> Something I'm still thinking about but I'm not entirely sure about `http_req_inspector_svc` I wonder if these should also just be normal services/layers. It feels like http-backend service is doing way too much. Removing JIT was actually a huge pleasure and in my opinion made everything easier

I am not against simpler. Feel free to make a more in depth proposal offline, discord or github.

GlenDC (approved):

Issue #620, opened 2025-07-06
support X-Clacks-Overhead (response) http header


GlenDC:
Value of header is when encoded picked based on epoch modulo from this list:

```
"Ashlynn"
      , "Terry Davis"
      , "Dennis Ritchie"
      , "Steven Hawking"
      , "John Conway"
      , "Ruth Bader Ginsburg"
      , "Bram Moolenaar"
      , "Grant Imahara"
      , "David Bowie"
      , "Sir Terry Pratchett"
      , "Satoru Iwata"
      , "Kris Nóva"
      , "Joe Armstrong"
      , "Paul Allen"
      , "Kevin Mitnick"
      , "Sir Clive Sinclair"
      , "Matt Trout"
```

- add credits in comment to original author Xe (https://github.com/Xe/site/blob/877872b4d7db92b602683ecb4e9934cbc4340079/dhall/package.dhall#L15-L33)

Support is to be added as

- name constant in https://github.com/plabayo/rama/blob/a117469c08ebe62dbb96502ca87a9f6f76221adc/rama-http-types/src/lib.rs#L129
- typed header in new file in https://github.com/plabayo/rama/tree/main/rama-http-headers/src

More info and also refer to it (incl summary added in code docs): 

https://xclacksoverhead.org/home/about

Easy issue for a lot of brownie points

HN post that inspired this issue: https://news.ycombinator.com/item?id=44440968

Hwatwasthat:
Hi, I would be interested in getting started with this if possible? not contributed to Rama before but been looking for a way to sneak in!

GlenDC:
Hi @Hwatwasthat , glad you finally found something to your liking. We keep issues like these open for the community as opportunities to (a) get you engaged into the framework and our community and (b) a learning opportunity. In return @soundofspace and I are here to mentor and guide you. So in case you are missing details, have questions, remarks, things you wish to discuss. Let us know.

I'll assign it to you in the meanwhile. If you do decide now or later to not want to pursuit it do let us know and we'll un-assign you without an issue. We understand choices can change and life can happen.

GlenDC:
@Hwatwasthat just checking in, no pressure. How is this going? Any guidance or help needed?

Hwatwasthat:
Hi Glen,

Sorry I haven't had as much time as I'd hoped for to work on this. I've started in a bit, but not much. Hopefully have some time this week to generate some questions!

GlenDC:
[...]

PR #734, opened 2025-11-14
[620] - add X-Clacks-Overhead header support

sim-hash:
closes https://github.com/plabayo/rama/issues/620


GlenDC:
Thanks this is good enough to go!

PR #737, opened 2025-11-17
Use closure to create nested routers and webservices to properly pass state

soundofspace:


rama-http/src/service/web/service.rs
```diff
@@ -489,12 +490,16 @@ mod test {
...
-            "/api",
-            WebService::new()
-                .get("/hello", "hello")
-                .post("/world", "world"),
-        );
+        let state = "state".to_owned();
+
+        let svc = WebService::new()
+            .with_state(state)
+            .get("/state", async |State(state): State<String>| state)
```
GlenDC, line 497: Ok so at a type level you already verify here that type-wise it seems to work. But to be usre the state is also actually (value-wise) passed down correctly, shouldn't you also try these endpoints and see if you get the correct value???

rama-http/src/service/web/router.rs
```diff
@@ -581,21 +582,27 @@ mod tests {
...
-                    ),
-                );
+                .sub(&format!("{prefix}users/{{user_id}}"), |router| {
+                    router
+                        .get(prefix, get_user_service())
+                        .get(
+                            &format!("{prefix}orders/{{order_id}}"),
+                            get_user_order_service(),
+                        )
+                        .get("/state", async |State(state): State<String>| state)
```
GlenDC: Ok so at a type level you already verify here that type-wise it seems to work. But to be usre the state is also actually (value-wise) passed down correctly, shouldn't you also try these endpoints and see if you get the correct value???

GlenDC (approved):
LGTM

rama-http/src/service/web/router.rs
soundofspace: Didnt really think it was needed since it's verified at type level, but added them in cases we do refactors, it never hurts to have extra tests

PR #742, opened 2025-11-25
Extractor for extensions

soundofspace:


rama-http/src/service/web/endpoint/extract/extension.rs
```diff
@@ -0,0 +1,117 @@
...
+    State: Send + Sync,
+    T: Send + Sync + Clone + 'static,
+{
+    type Rejection = Infallible;
+
+    async fn from_request_context_ref_pair(
+        parts: &Parts,
+        _state: &State,
+    ) -> Result<Option<Self>, Self::Rejection> {
+        Ok(parts.extensions.get::<T>().map(|ext| Self(ext.clone())))
```
GlenDC, line 50: I would have gone for `Ok(parts.extensions.get::<T>().cloned().map(Self)`, but as long as clippy is happy all good I suppse

PR #747, opened 2025-11-26
Issue 462/revisit service api (core) design in 0.3 milestone#renaming service input and output

soundofspace:
Part of #462


rama-core/src/layer/limit/policy/concurrent.rs
```diff
@@ -127,16 +127,16 @@ where
...
         };
 
-        PolicyResult { request, output }
+        PolicyResult { input, output }
     }
 }
 
 rama_utils::macros::error::static_str_error! {
-    #[doc = "request aborted due to exhausted concurrency limit"]
+    #[doc = "input aborted due to exhausted concurrency limit"]
```
GlenDC: Would be careful with these grep replacements, a sentence like this makes little sense, with request it did make sense.

perhaps label it as `serve aborted....` 🤷

soundofspace: this was not even grep, but yea agree that it sounds weird, just immediatly think of serve instead

rama-core/src/layer/consume_err.rs
```diff
@@ -215,38 +215,38 @@ impl ConsumeErrLayer<Trace> {
     }
 }
 
-impl<F> ConsumeErrLayer<F, DefaulResponse> {
+impl<F> ConsumeErrLayer<F, DefaulOutput> {
```
GlenDC:
```suggestion
impl<F> ConsumeErrLayer<F, DefaultOutput> {
```

rama-core/src/layer/consume_err.rs
```diff
@@ -215,38 +215,38 @@ impl ConsumeErrLayer<Trace> {
...
     /// Set a response to be used in case of errors,
-    /// instead of requiring and using the [`Default::default`] implementation
+    /// instead of inputuiring and using the [`Default::default`] implementation
```
GlenDC:
```suggestion
    /// instead of requiring and using the [`Default::default`] implementation
```

lol

rama-core/src/layer/consume_err.rs
```diff
@@ -261,11 +261,11 @@ mod sealed {
 
     #[derive(Debug, Clone)]
     #[non_exhaustive]
-    /// A sealed type to indicate default response is to be used.
-    pub struct DefaulResponse;
+    /// A sealed type to indicate default output is to be used.
+    pub struct DefaulOutput;
```
GlenDC:
```suggestion
    pub struct DefaultOutput;
```

rama-core/src/matcher/mod.rs
```diff
@@ -1,5 +1,5 @@
-//! matcher utilities for any middleware where need to match
-//! on incoming requests within a given [`Context`]
+//! Matcher utilities for any middleware where need to match
+//! on incoming inputs/Inputs
```
GlenDC:
```suggestion
//! on incoming inputs.
```

rama-net/src/client/pool/mod.rs
```diff
@@ -645,20 +645,20 @@ impl<C, ID: Debug> Debug for LruDropPool<C, ID> {
     }
 }
 
-/// [`ReqToConnID`] is used to convert a `Request` to a connection ID. These IDs
+/// [`ReqToConnID`] is used to convert a `Input` to a connection ID. These IDs
 /// are not unique and multiple connections can have the same ID. IDs are used
-/// to filter which connections can be used for a specific Request in a way that
-/// is independent of what a Request is.
-pub trait ReqToConnID<Request: ExtensionsRef>: Sized + Clone + Send + Sync + 'static {
+/// to filter which connections can be used for a specific Input in a way that
```
GlenDC:
```suggestion
/// to filter which connections can be used for a specific input in a way that
```

rama-net/src/client/pool/mod.rs
```diff
@@ -645,20 +645,20 @@ impl<C, ID: Debug> Debug for LruDropPool<C, ID> {
...
+/// is independent of what a Input is.
```
GlenDC:
```suggestion
/// is independent of what an input is.
```

rama-net/src/client/pool/mod.rs
```diff
@@ -645,20 +645,20 @@ impl<C, ID: Debug> Debug for LruDropPool<C, ID> {
...
 
-    fn id(&self, request: &Request) -> Result<Self::ID, OpaqueError>;
+    fn id(&self, input: &Input) -> Result<Self::ID, OpaqueError>;
 }
 
 /// [`ConnID`] is used to identify a connection in a connection pool. These IDs
 /// are not unique and multiple connections can have the same ID. IDs are used
-/// to filter which connections can be used for a specific Request in a way that
-/// is independent of what a Request is.
+/// to filter which connections can be used for a specific Input in a way that
```
GlenDC:
```suggestion
/// to filter which connections can be used for a specific input in a way that
```

rama-net/src/client/pool/mod.rs
```diff
@@ -645,20 +645,20 @@ impl<C, ID: Debug> Debug for LruDropPool<C, ID> {
...
+/// is independent of what a Input is.
```
GlenDC:
```suggestion
/// is independent of what an input is.
```

rama-net/src/client/pool/mod.rs
```diff
@@ -878,33 +878,33 @@ mod tests {
...
+                input,
                 conn: Conn::new(),
             })
         }
     }
 
     #[derive(Clone)]
-    /// [`StringRequestLengthID`] will map Requests of type ServiceInput<String>, to usize id representing their
-    /// chars length. In practise this will mean that Requests of the same char length will be
+    /// [`StringInputLengthID`] will map Inputs of type ServiceInput<String>, to usize id representing their
```
GlenDC:
```suggestion
    /// [`StringInputLengthID`] will map inputs of type ServiceInput<String>, to usize id representing their
```

rama-net/src/client/pool/mod.rs
```diff
@@ -878,33 +878,33 @@ mod tests {
...
+    /// chars length. In practise this will mean that Inputs of the same char length will be
```
GlenDC:
```suggestion
    /// chars length. In practise this will mean that inputs of the same char length will be
```

rama-net/src/client/conn.rs
```diff
@@ -1,38 +1,30 @@
 use rama_core::{Service, error::BoxError, extensions::ExtensionsMut, service::BoxService};
 use std::fmt;
 
+#[derive(Clone)]
```
GlenDC, line 4: not something that we want to or should enforce, I prefer the manual clone impl here

rama-net/src/client/pool/mod.rs
```diff
@@ -1163,7 +1163,7 @@ mod tests {
         drop(conns);
 
         // We should now have two connections with the same key in the pool,
-        // from most to least recently used. ([conn2, conn1]). Requesting
+        // from most to least recently used. ([conn2, conn1]). Inputing
```
GlenDC:
```suggestion
        // from most to least recently used. ([conn2, conn1]). Requesting
```

rama-tls-rustls/src/client/connector.rs
```diff
@@ -432,7 +444,7 @@ fn set_target_http_version(
 #[non_exhaustive]
 #[derive(Debug, Clone)]
 /// A connector which can be used to establish a connection to a server
-/// in function of the Request, meaning either it will be a seucre
+/// in function of the Input, meaning either it will be a secure
```
GlenDC:
```suggestion
/// in function of the input, meaning either it will be a secure
```

PR #749, opened 2025-11-28
Patch/improve http headers

GlenDC:
Remove use of unwraps and expect for the typed headers here and improve logic around them.

Also introduce new utilities:

- `rama::utils::collections::NonEmptyVec`: vec-like type which guarantees to have at least 1 element
- `rama::str::arcstr::*`: `ArcStr` and friends as better `Arc<str>`  replacement (fromStr, serde, minor improvs, static 0-mem support, ...)

### DIFF
--- a/examples/http_anti_bot_zip_bomb.rs
+++ b/examples/http_anti_bot_zip_bomb.rs
@@ -45,6 +45,7 @@ use rama::{
         level_filters::LevelFilter,
         subscriber::{EnvFilter, fmt, layer::SubscriberExt, util::SubscriberInitExt},
     },
+    utils::str::arcstr::ArcStr,
 };
 
 /// Everything else we need is provided by the standard library, community crates or tokio.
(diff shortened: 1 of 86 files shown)


docs/thirdparty/fork/README.md
```diff
@@ -121,3 +121,19 @@ but still directly in function of Rama.
...
+
+These are forks initially because the libraries are too simple or small
+to really warrant a permanent entry in our dep tree, yet are useful enough
+to give a foundation for similar functionality that we want.
+
+Over time they might diverge from the original as it grows with the rest
+of the rama ecosystem.
+
+- <https://github.com/cloudhead/nonempty/tree/95d5cb131262b12bbe55366cbbd48096f9a05493>
+  - it was more or less unmaintained and missing features due to being behind in rust versions
```
GlenDC:
```suggestion
```

PR #750, opened 2025-11-29
Remove request inspectors

soundofspace:
This fits in the effort of #462 because it removes extra complexity and goes back to the core idea of everything is a service.
The main benefit here is removing extra complexity while still keeping the same functionality (simpler = better most of the time).

What is done in this PR
- Remove request inspector concept/code
- Refactor request inspector service to normal services/layers
- Remove http_req_inspector in http connector. While this is a useful thing, it should not be part of the core http connector
- Refactor EasyHttpClient to be more explicit: build_connector, connector_builder... vs build (what do we build?)
- Support jit_layers in EasyHttpClient, here this concept is still needed, but since this is a higher level abstraction it makes sense here

### DIFF
--- a/examples/http_record_har.rs
+++ b/examples/http_record_har.rs
@@ -305,7 +305,7 @@ async fn http_mitm_proxy(req: Request) -> Result<Response, Infallible> {
         .with_custom_connector(UserAgentEmulateHttpConnectModifierLayer::default())
         .with_default_http_connector()
         .build_client()
-        .with_jit_layers(UserAgentEmulateHttpRequestModifierLayer::default());
+        .with_jit_layer(UserAgentEmulateHttpRequestModifierLayer::default());
 
     let state = req.extensions().get::<State>().unwrap();
 
--- a/rama-cli/src/cmd/send/http/client/curl_writer.rs
+++ b/rama-cli/src/cmd/send/http/client/curl_writer.rs
@@ -1,5 +1,3 @@
-use std::convert::Infallible;
-
 use rama::{
     Service,
     error::{ErrorContext as _, OpaqueError},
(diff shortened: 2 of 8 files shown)


examples/http_record_har.rs
```diff
@@ -297,15 +296,15 @@ async fn http_mitm_proxy(req: Request) -> Result<Response, Infallible> {
...
         .with_default_transport_connector()
         .with_tls_proxy_support_using_boringssl()
         .with_proxy_support()
         .with_tls_support_using_boringssl(Some(Arc::new(base_tls_config)))
         .with_custom_connector(UserAgentEmulateHttpConnectModifierLayer::default())
         .with_default_http_connector()
-        .with_svc_req_inspector(UserAgentEmulateHttpRequestModifier::default())
-        .build();
+        .build_client()
+        .with_jit_layers(UserAgentEmulateHttpRequestModifierLayer::default());
```
GlenDC: I think we should rename this to `with_jit_layer`. If you want me to do it I can do it in a follow-up commit after merge, but either way someone will have to rename it. Reason you can give multiple layers as well is because a tuple of layers is also a layer.

rama-cli/src/cmd/send/http/client/curl_writer.rs
```diff
@@ -20,11 +23,13 @@ impl Service<Request> for CurlWriter {
     type Response = Response;
 
     async fn serve(&self, req: Request) -> Result<Self::Response, Self::Error> {
-        let req = UserAgentEmulateHttpRequestModifier::new()
-            .serve(req)
-            .await
-            .map_err(OpaqueError::from_boxed)
-            .context("rama: (curl-writer) emulate UA")?;
+        let req = UserAgentEmulateHttpRequestModifier::new(service_fn(async |req: Request| {
```
GlenDC: Gonna make a svc for this as bit silly you have to do this effort.

PR #759, opened 2025-12-09
Issue 756/Rename GetExtension to GetInputExtension and create GetOutputExtension

M-Kusumgar:
Closes #756 

I also think I need to create a `GetOutputExtensionLayer` for the `GetOutputExtension`. Is this what you had in mind? I basically added a copy of `serve` but instead it passes the cloned value of the extension from the output to the callback passed in.

PR #761, opened 2025-12-09
Issue 756/Rename AddExtension to AddInputExtension and add AddOutputExtension

M-Kusumgar:
Couple of thoughts I had while doing this one:
*  I am not sure if my example at the top of `add_extension` with the `ResponseCounter` is a good one, happy to change it/remove it if we don't need an example
* The `serve` function for `AddInputExtension` used to return a `Future<>` but for `AddOutputExtension` I needed to await to get the response, so I made both of them consistent and converted them into `async fn`s. This is also consistent with `GetInputExtension` and `GetOutputExtension`. I am not super familiar with async rust (let alone rust) so I am not 100% sure about the implications of this change but I have done a bit of looking into it so `async fn` is the same as returning a `Future<>` from a function that wraps the original function body in an `async move`. The things im not sure about are:
    * `'_` - im not sure we need this, presumably we want the returned value to live as long as the args, is that the default for async functions?
    * `Send` - not sure if we need to enforce it on the return here, `Input` has the trait enforced and its the only one we take a mutable reference to

Relates to #759

### DIFF
--- a/examples/http_mitm_proxy_boring.rs
+++ b/examples/http_mitm_proxy_boring.rs
@@ -86,7 +86,7 @@ use rama::{
             protocol::{Role, WebSocketConfig},
         },
     },
-    layer::{AddExtensionLayer, ConsumeErrLayer},
+    layer::{AddInputExtensionLayer, ConsumeErrLayer},
     matcher::Matcher,
     net::{
         http::RequestContext,
@@ -179,7 +179,7 @@ async fn main() -> Result<(), BoxError> {
             .serve_graceful(
                 guard,
                 (
-                    AddExtensionLayer::new(state),
+                    AddInputExtensionLayer::new(state),
                     // protect the http proxy from too large bodies, both from request and response end
                     BodyLimitLayer::symmetric(2 * 1024 * 1024),
                 )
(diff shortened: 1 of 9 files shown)


GlenDC:
Flawless job, thx!

GlenDC:
> I am not sure if my example at the top of add_extension with the ResponseCounter is a good one, happy to change it/remove it if we don't need an example

That's ok, I'll remove it again myself, as part of some other nitpick minor cleanup, as:

1. it's not really a good example, extensions usually do not work as composed state, so in this case you would better just add the counter yourself
2. the counter is not even functional here as you are not incrementing it
3. doc examples are a pain in the but to maintain, not your fault, just a shortcoming in rust analyzer support for these things

> The serve function for AddInputExtension used to return a Future<> but for AddOutputExtension I needed to await to get the response, so I made both of them consistent and converted them into async fns. This is also consistent with GetInputExtension and GetOutputExtension. I am not super familiar with async rust (let alone rust) so I am not 100% 

Yeah for output you would indeed need that (unless you work with FutureExt trait and map it, but that's something I prefer to avoid in rama to make it more ergonomic... So yeah good call

No need to make it consistent, but neither I really care here, it's ok. Can stay like you did it. There's a minor difference but nothing that I mind here. The `async fn` notation is just sugar and will expand to a future that is returned but with some forced implications. 

`'_` is a special notation for a lifetime which basically implies the liftime of the value itself. I refer you to the official docs for more info on it.

PR #765, opened 2025-12-15
add windows (11) arm support for CI + Release

GlenDC:


GlenDC:
Problem is in rama-boring, build config of sys seems to wrongly assume arm64 ties to macos for some branches... Need to fix that upstream first.

PR #766, opened 2025-12-15
Issue 374/Benchmark for e2e h1 client server interaction

M-Kusumgar:
Here is a first go at setting up a skeleton for benchmarking this type of interaction. Is this the right approach? This is just the demo, I will do a benchmark for a bigger payload and I will make the smaller payload a bit bigger than that if required haha.

I have a couple of questions that ive written as comments in the code and a couple of more general questions:
1. you said in the ticket it would be nice to add a realistic amount of layers, are there any in particular you had in mind?
2. should i be benchmarking just the client request? or starting the server too?

PS im just beginning to understand this project, its cool, its like a pick n mix but for networks

### DIFF
--- a/benches/e2e_h1_client_server.rs
+++ b/benches/e2e_h1_client_server.rs
@@ -3,19 +3,18 @@ use std::{thread, time::Duration};
 use rama::{
     Layer, Service,
     http::{
-        Body, Request,
+        Body, HeaderName, HeaderValue, Request,
         client::EasyHttpWebClient,
         layer::{
             compression::CompressionLayer, decompression::DecompressionLayer, trace::TraceLayer,
         },
-        server::HttpServer,
-        service::web::WebService,
-        HeaderName, HeaderValue,
         layer::{
             cors::CorsLayer,
             required_header::{AddRequiredRequestHeadersLayer, AddRequiredResponseHeadersLayer},
             set_header::{SetRequestHeaderLayer, SetResponseHeaderLayer},
         },
+        server::HttpServer,
+        service::web::WebService,
     },
     net::address::SocketAddress,
     rt::Executor,
@@ -124,7 +123,6 @@ async fn request_payload(client: impl Service<Request>, payload: &Payload, body_
     let _ = client.serve(req).await;
 }
 
-// make sure SAMPLE_COUNT is divisible by SEEDS.len()
 const SAMPLE_COUNT: u32 = 1000;
 
 #[divan::bench(sample_count = SAMPLE_COUNT, args = [
@@ -146,8 +144,7 @@ fn h1_client_server(bencher: divan::Bencher, payload: Payload) {
             let server_random_bytes = random_bytes_by_size(&mut rng, payload.server);
             let client_random_bytes = random_bytes_by_size(&mut rng, payload.client);
 
-            server_thread =
-                tokio::spawn(run_server(payload.server, server_random_bytes));
+            server_thread = tokio::spawn(run_server(payload.server, server_random_bytes));
             block_on(tokio::time::sleep(Duration::from_micros(5)));
 
             let client = (


benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,96 @@
...
+
+const ADDRESS: SocketAddress = SocketAddress::local_ipv4(62004);
+
+#[global_allocator]
+static ALLOC: divan::AllocProfiler = divan::AllocProfiler::system();
+
+fn main() {
+    // Make Tokio available for the duration of the process:
+    let executor = tokio::runtime::Runtime::new().unwrap();
+    let _guard = executor.enter();
```
M-Kusumgar: see https://github.com/nvzqz/divan/issues/39#issuecomment-2614321742

not sure about the validity of this considering in the same thread they note that `block_on` adds a lot of overhead

benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,96 @@
...
+    HttpServer::http1()
+        .listen(ADDRESS, http_service)
+        .await
+        .unwrap();
+}
+
+async fn get_small_payload(client: impl Service<Request>) {
+    let req = Request::builder()
+        .uri(format!("http://{ADDRESS}/small-payload"))
+        .method("GET")
```
M-Kusumgar: should we be benchmarking client large post and client small post as well?

benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    b.bench_local(|| block_on(get_small_payload(client.clone())));
+}
+
+fn setup_tracing() {
+    tracing::subscriber::registry()
+        .with(fmt::layer())
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::INFO.into())
```
M-Kusumgar: the debug output was getting printed in the middle of the benchmark numbers probably because its on a different thread, should i change it back to debug? will it affect the benchmarks?

GlenDC:
Thank you for this swift attempt of yours @M-Kusumgar , please give me some time to come back to you in some days. It's suddenly extremely busy. So I want to give myself some time if you're ok with that to gather my toughts on this benchmarking topic so that I can be fair to the work you put into this and guide you towards the appropriate milestones. Looking forward to do this. Thank you for getting this started.

@soundofspace please also provide feedback here, also think about how we might want to use these benchmarks (e.g. do we display results somewhere? Can we keep a tap on these numbers? Do we somehow integrate this in our PRs? How, What? Etc...

M-Kusumgar:
no problem! no rush on it

benches/e2e_h1_client_server.rs
GlenDC: It adds overhead yes, but it's consistent overhead. E.g. instead of 100ms it might take 10ms. We can take that mentally in account and more importantly that overhead is consistent over time, so we would still be able to spot improvements / regressions of the actual benchmarked code regardless. So what you do is fine for me.

benches/e2e_h1_client_server.rs
GlenDC: I would trace it to a temporary file path and print the file path once at startup. this way you can allow debug traces. Which is good to be part of bench code.

benches/e2e_h1_client_server.rs
GlenDC: Yes

GlenDC (changes requested):
Looks good to me. Do add some more cases already as mentioned in ticket and also as you suggest things like client payload, and both server / client small and large. 

You can also make it a bit more of a complex stack with compression added, also things like adding required headers etc...

What you have is cool though, so good start. Sorry for getting back so late.

What is the output of the test you have so far?

Once you add more tests btw, do try to make sure to keep DRY and not have too much (boring setup) code repeated through your various tests. But great start, I like it.

M-Kusumgar:
sorry missed this notification! ill be back on this probably tomorrow or day after at the latest, excited to do some more work on this project!

M-Kusumgar:
Sorry for the delay! I added some layers, am now piping the trace to a file and have a structure for testing out different payload sizes (i may have gone overboard with different sizes, im happy to change it to just small and large and remove medium if you think this is overkill) without too much code duplication (i think), i haven't added different cases yet, however i had a concern about this that i can't seem to figure out:

Firstly these are the results i get on my computer:
image

These seem reasonable on the surface however if i isolate just the client payload size changing with the same server payload size, say medium then i get:
image

it seems that the client large payload does not cause as much of a delay as the server large payload, is this difference just due to the compression layer? or am i doing something wrong with the setup?

~~Edit: I cannot seem to recreate this CI error, `cargo check --workspace --all-targets --all-features` and `just qa` both pass on my machine, i have seen `HttpServer::http1()` being used without the executor arg but maybe its because of the second error, not sure~~

benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,165 @@
...
+
+#[derive(Debug)]
+struct Payload {
+    server: Size,
+    client: Size,
+}
+
+fn get_body_for_size(size: &Size) -> &'static [u8] {
+    match size {
+        Size::Empty => &[b'x'; 0],
```
soundofspace: I'll leave the review up to @GlenDC but you will want to generate "randomly" consistent data here. Static data like this can give very weird results in any type of benchmark since it's easily compressed (too easy) and doesn't mimic real world data

benches/e2e_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+        .create(true)
+        .open(&log_file_path)
+        .unwrap();
+    tracing::subscriber::registry()
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::DEBUG.into())
+                .from_env_lossy(),
+        )
+        .with(fmt::layer().with_writer(file))
```
soundofspace: In normal setups when logging/tracing to a file you want some sort of buffer/appender (eg tracing-appender) otherwise this will cause a lot of blocking and might have quite a big impact on the benchmark

benches/e2e_utils.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+    let file = OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(&log_file_path)
+        .unwrap();
+    tracing::subscriber::registry()
+        .with(
+            EnvFilter::builder()
+                .with_default_directive(LevelFilter::DEBUG.into())
```
soundofspace: @GlenDC do we want to support env filter here. I guess it's useful for testing, but it seems dangerous that if you have RUST_LOG set it might "affect" benchmark results

GlenDC: True I would not support RUST_LOG here. Just leave it at the default debug+ level.

benches/e2e_utils.rs
GlenDC: Would perhaps logging to stderr be better?  Or what you suggest here @soundofspace

benches/e2e_h1_client_server.rs
GlenDC: Agreed, You'll want to probably just generate random data using either `Rand` or the `InfiniteReader` from rama body and turn that into bytes that you can reuse. Do make sure to do that in the phase prior to it benchmarking.

GlenDC:
I'll review begin next week, but do already work with @soundofspace to get his feedback resolved. Can you also please make a habbit of resolving threads which you have resolved, and other threads that have to remain open those should then have a comment in the thread on why it is open e.g. perhaps you have a question or w/e

benches/e2e_h1_client_server.rs
M-Kusumgar: have opted for using rand, thanks

benches/e2e_utils.rs
M-Kusumgar: i have changed it to include trace-appender with its non blocking writer, keeping this thread open because of your question @GlenDC

benches/e2e_utils.rs
M-Kusumgar: unsure about this thread, could i have some clarification @GlenDC, just leaving it at the default debug level means i dont have to do anything here? am i understanding that correctly?

M-Kusumgar:
~~@GlenDC I will deal with the merge conflicts tomorrow, the CI was failing because I hadn't pulled in `main`, but the general structure of the code shouldn't change, just how I start up the http server~~

I believe I've also addressed @soundofspace's feedback so should be good to review while i work on pulling `main`

benches/e2e_h1_client_server.rs
```diff
@@ -149,6 +153,27 @@ async fn request_payload(client: impl Service<Request>, payload: &Payload) {
...
+    };
+    let mut client_random_bytes = server_random_bytes.clone();
+
+    let mut rng = rand::rng();
+    rng.fill_bytes(&mut server_random_bytes.small);
+    rng.fill_bytes(&mut server_random_bytes.medium);
+    rng.fill_bytes(&mut server_random_bytes.large);
+    rng.fill_bytes(&mut client_random_bytes.small);
+    rng.fill_bytes(&mut client_random_bytes.medium);
+    rng.fill_bytes(&mut client_random_bytes.large);
```
soundofspace: For this to be consistent between different runs you also need to seed the rng, that way each run will be consist. And it's probably also best if you seed it every time before calling fill_bytes, so if this function is ever refactored, that tests payloads stay the same

benches/e2e_utils.rs
soundofspace: Here it means not using from_env_lossy but just a static logLevel. Otherwise this function will always give priority to the environment variable RUST_LOG, which could be different in CI, locally, my machine vs your machine. By making it static this benchmark because again more predictable

benches/e2e_h1_client_server.rs
M-Kusumgar: i did think about doing a seedable rng, but then wouldnt that open up the possibility of hitting edge cases where certain paths in the code are taken because of the data? for example, take the extreme case that the random bytes by chance are something easy to compress, then the benchmarks would not be representative of real world randomness, we will only ever be testing one payload for large server response for example

perhaps the seed can be different for each iteration? but still derived in a stable way, like `seed_from_u64` or something like that? that way we explore the benchmark over more of the randomness but the benchmark still remains very consistent?

happy to try and implement that

benches/e2e_utils.rs
M-Kusumgar: thanks for the clarification! have removed any reference to the env and instead just did a static level filter

benches/e2e_h1_client_server.rs
soundofspace: How many iterations does a benchmark like this run? If it's high enough then I agree with you

benches/e2e_utils.rs
soundofspace: @GlenDC this is what I meant. With appender layer it should be fine

benches/e2e_h1_client_server.rs
M-Kusumgar: i have written some code for this, ive (randomly :wink:) selected 5 seeds, each of them run for 2000 iterations, the median values are pretty stable, i ran the full benchmark a couple of times with cpu affinity set to reduce noise and the medians were within 1-2% bands roughly, do you think that iteration number is high enough?

i also didnt reset the rng between getting `server_random_bytes` and `client_random_bytes` because the server and client bytes probably shouldnt be correlated, but they will always be consistent given the seed and `Payload` so should be more reproducible

benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,189 @@
...
+            compression::CompressionLayer, decompression::DecompressionLayer, trace::TraceLayer,
+        },
+        server::HttpServer,
+        service::web::WebService,
+    },
+    net::address::SocketAddress,
+    rt::Executor,
+};
+
+use rama_http::{
```
GlenDC: only import from `rama` root crate please, not `rama_http` but `rama::http`

benches/e2e_h1_client_server.rs
```diff
@@ -0,0 +1,189 @@
...
+        .method("POST")
+        .body(Body::from(body_content))
+        .unwrap();
+
+    let _ = client.serve(req).await;
+}
+
+// make sure SAMPLE_COUNT is divisible by SEEDS.len()
+const SAMPLE_COUNT: u32 = 10_000;
+const SEEDS: [u64; 5] = [42, 10191, 451, 73, 8128];
```
GlenDC: I would prefer not to hardcode seeds. Should just be a fresh one per iteration.

Bench numbers are never gonna be perfect, fine to have some margin on it due to things like seed giving something that looks like random.

Cargo.toml
```diff
@@ -459,12 +463,15 @@ itoa = { workspace = true }
 libc = "0.2"
 parking_lot = { workspace = true }
 pin-project-lite = { workspace = true }
+rand = { workspace = true }
+rand_chacha = { workspace = true }
```
GlenDC: Don't think we need to go for another dependency just for this. This is not a secure use case, regular rand is fine.

benches/e2e_h1_client_server.rs
M-Kusumgar: done

benches/e2e_h1_client_server.rs
M-Kusumgar: sure, i have removed the seeding logic and have just used `rand::rng` fresh at the start of each iteration

Cargo.toml
M-Kusumgar: have removed this dep

GlenDC:
@M-Kusumgar I'm going to merge https://github.com/plabayo/rama/pull/816 in a bit. Once it's merged can you look at the final squashed commit that's in main to learn from). If you are still up for more, let me know. The next step would be to add a tls enum to the `TestParameters`:

```
enum Tls {
    None,
    Rustls,
    Boring,
}
```

Would also mean you need to enable these tls features in the cargo.toml.

That would allow us to also see what tls on top of that adds to the results :)

And with that in place I think the http benchmark is complete. Let me know if you are up for it and in that case you can start working towards your next PR!

M-Kusumgar:
hey @GlenDC im definitely up for more, ill take a look at that PR, and extend that with TLS, unfortunately i wont have time tonight to do it, but i will have time tomorrow and the day after if that delay is fine?

GlenDC:
> hey @GlenDC im definitely up for more, ill take a look at that PR, and extend that with TLS, unfortunately i wont have time tonight to do it, but i will have time tomorrow and the day after if that delay is fine?

Sure, tomorrow and the day after is plenty on time. No worries. Thank you for your engagement.

PR #770, opened 2025-12-21
Adapt HTTP/2/3 requests to HTTP/1.x by merging multiple Cookie headers (RFC 6265 5.4)

### LINKED ISSUE #768: HTTP/2 → HTTP/1.1 upstream conversion sends multiple `Cookie` headers (instead of a single merged header), causing request failures
### Description
When using **rama** as a proxy that **accepts HTTP/2 from the client** and forwards requests **upstream as HTTP/1.1**, some requests fail against certain servers

From packet captures, the failing behavior appears to be that rama forwards **multiple `cookie:` header fields** in the outgoing **HTTP/1.1** request, rather than merging them into **a single `Cookie:` header**. Some servers reject or mis-handle this, and it violates RFC behavior for HTTP/1 cookie handling.

This does **not** happen when the proxy only allows HTTP/1.x from the client (working case).

### Why this seems related to HTTP/2 → HTTP/1.1 translation
In HTTP/2, it is common/allowed for intermediaries to represent cookies as multiple header entries (and header names are lowercase). When emitting an HTTP/1.1 request, intermediaries should serialize cookies into a single `Cookie` header field.

Relevant RFCs:
- RFC 6265 §5.4: “When the user agent generates an HTTP request, the user agent MUST NOT attach more than one Cookie header field.”  
  https://datatracker.ietf.org/doc/html/rfc6265#section-5.4  

### Expected behavior
When converting and forwarding as **HTTP/1.1**, rama should ensure:
- At most **one** `Cookie` header is present in the outgoing HTTP/1.1 request.
- If multiple cookie header fields exist internally (e.g., from HTTP/2), they should be combined into one header value joined by `"; "` (preserving order).

Example expected outgoing header:
```
Cookie: a=1; b=2; c=3
```

### Actual behavior
When the proxy accepts HTTP/2, the upstream HTTP/1.1 request contains multiple cookie header fields, e.g.:
```
Cookie: uaid=...
cookie: MSCC=...
cookie: MUID=...
cookie: fptctx2=...
cookie: MSPRequ=...
cookie: MSPOK=...
...
```

In my tests, converting these multiple cookie headers into a single `Cookie:` header in Burp makes the request succeed immediately, strongly suggesting the multiple-cookie-header serialization is the cause of the failure.

### Possible fix
In the HTTP/2→HTTP/1.1 request conversion:
- Collect all `cookie` header fields.
- Join values using `"; "` into a single `Cookie` header.
- Remove extra cookie header fields before serializing the HTTP/1.1 request.

ntrippar:
Adapt HTTP/2/3 requests to HTTP/1.x by merging multiple Cookie headers (RFC 6265 5.4)

* Implements RFC 6265 compliance when downgrading requests from HTTP/2 or HTTP/3 to HTTP/1.x by merging multiple Cookie headers into a single header joined with "; ".
* Adds merge_cookie_headers_for_http1() and invokes it from adapt_request_version() only for 2/3 → 1.0/1.1 conversions; preserves header order, logs with tracing, and surfaces parse errors via OpaqueError.
* No-op when versions are equal; no merge when upgrading to HTTP/2/3; single or no Cookie headers remain unchanged.
* Includes tests covering HTTP/2→HTTP/1.1, upgrade no-merge behavior, and same-version no-op.

Closes #768

### DIFF
--- a/rama-http/src/layer/version_adapter/request.rs
+++ b/rama-http/src/layer/version_adapter/request.rs
@@ -1,15 +1,20 @@
 use rama_core::Layer;
 use rama_core::Service;
+use rama_core::bytes::BytesMut;
 use rama_core::extensions::ChainableExtensions;
 use rama_core::telemetry::tracing;
 use rama_error::BoxError;
+use rama_error::ErrorContext;
 use rama_error::OpaqueError;
 use rama_http_headers::HeaderMapExt;
 use rama_http_headers::Upgrade;
+use rama_http_types::HeaderValue;
 use rama_http_types::Method;
 use rama_http_types::Request;
 use rama_http_types::Version;
 use rama_http_types::conn::TargetHttpVersion;
+use rama_http_types::header::COOKIE;
+use rama_http_types::header::Entry;
 use rama_net::client::{ConnectorService, EstablishedClientConnection};
 use rama_utils::macros::generate_set_and_with;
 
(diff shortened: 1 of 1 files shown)


ntrippar:
@GlenDC I did not know if this was the right place to add this. so if you have any changes or recommendation please let me know and I will change it.

rama-http/src/layer/version_adapter/request.rs
```diff
@@ -155,6 +155,206 @@ pub fn adapt_request_version<Body>(
...
+        merge_cookie_headers_for_http1(request)?;
+    }
+
     *request.version_mut() = target_version;
     Ok(())
 }
+
+/// Merge multiple cookie headers into a single Cookie header for HTTP/1.x compliance
+/// per RFC 6265 §5.4: "the user agent MUST NOT attach more than one Cookie header field"
+fn merge_cookie_headers_for_http1<Body>(request: &mut Request<Body>) -> Result<(), OpaqueError> {
```
GlenDC, line 176: Can you try this implementation instead, I haven't tested it though just did it quickly in my editor, but the difference is that:

1. remove does not remove all headers just first entry
2. you force it to be a utf-8 string while that is not strictly required (so keeping bytes is fine)
3. we can allocate all required bytes at once given we know about the headers

```rust
fn merge_cookie_headers_for_http1<Body>(request: &mut Request<Body>) -> Result<(), OpaqueError> {
    if let Entry::Occupied(cookie_headers) = request.headers_mut().entry(COOKIE) {
        let Some((bytes_count, header_count)) = cookie_headers
            .iter()
            .map(|v| (v.as_bytes().len(), 1usize))
            .reduce(|a, b| (a.0 + b.0, a.1 + b.1))
        else {
            return Ok(());
        };
        if header_count <= 1 {
            return Ok(());
        }

        let (header_name, mut header_values) = cookie_headers.remove_entry_mult();

        let mut buffer = BytesMut::with_capacity(bytes_count + ((header_count - 1) * 2));
        if let Some(header_value) = header_values.next() {
            buffer.extend_from_slice(header_value.as_bytes());
        }
        for header_value in header_values {
            buffer.extend_from_slice(b"; ");
            buffer.extend_from_slice(header_value.as_bytes());
        }

        let new_header_value = HeaderValue::from_maybe_shared(buffer)
            .context("create new cookie header value from combined multiple values")?;

        request.headers_mut().insert(header_name, new_header_value);
    }

    Ok(())
}
```

rama-http/src/layer/version_adapter/request.rs
```diff
@@ -155,6 +155,206 @@ pub fn adapt_request_version<Body>(
...
+            .header(COOKIE, "first=1")
+            .header(COOKIE, "second=2")
+            .header(COOKIE, "third=3")
+            .header(COOKIE, "fourth=4")
+            .body(())
+            .unwrap();
+
+        adapt_request_version(&mut req, Version::HTTP_11).unwrap();
+
+        let cookie_value = req.headers().get(COOKIE).unwrap().to_str().unwrap();
```
GlenDC: you do not test here that there aren't multiple header values, which you should test. Your implementation should (AFAIK) fail on that test as you are not removing all headers

for that use `get_all` or w/e and ensure you only have 1

rama-http/src/layer/version_adapter/request.rs
```diff
@@ -155,6 +155,206 @@ pub fn adapt_request_version<Body>(
         *request.method_mut() = Method::CONNECT;
     }
 
+    // RFC 6265 §5.4: When converting to HTTP/1.x, merge multiple cookie headers into one
+    // In HTTP/2 and HTTP/3, multiple cookie headers are allowed, but HTTP/1.x must have
+    // at most one Cookie header field with values joined by "; "
+    if (target_version == Version::HTTP_10 || target_version == Version::HTTP_11)
```
GlenDC: can you do this instead (also replace that other  if function regarding upgrade in same manner:

`if target_version <= Version::HTTP_11 && request_version >= Version::HTTP_2 {`

that should afaik do the same but easier and shorter to write

ntrippar:
Perfect, Yeah I will change this but also reading the RFC I think there are other cases that might be affected. I will write in the parent issue about it

rama-http/src/layer/version_adapter/request.rs
ntrippar: I changed and tested and seems to works correctly

rama-http/src/layer/version_adapter/request.rs
ntrippar: done, chenged to match your recomendation.

rama-http/src/layer/version_adapter/request.rs
ntrippar: I did some changes in the tests, also added a test that from http1 to http1 shoudn't convert as that is intended behaviour that the user is sending in propose two same headers (could be performing a security test)

this way I check doing the get_all and also compare directly with bytes instead of converting to str.

https://github.com/plabayo/rama/pull/770/changes#diff-dc25649e8dee2146c54325e24857f767fe0e7f2b63470c39282fe54ed4fc41bfR354-R359

ntrippar:
Before merging this I want to perform some more testing and will add a few more test with "broken" cookies, like that they don't have a key but maybe not a value, will test that now

rama-http/src/layer/version_adapter/request.rs
```diff
@@ -155,6 +160,246 @@ pub fn adapt_request_version<Body>(
...
+        adapt_request_version(&mut req, Version::HTTP_11).unwrap();
+
+        // also should have only one cookie header with complex values preserved
+        let cookie_values: Vec<_> = req.headers().get_all(COOKIE).iter().collect();
+        assert_eq!(cookie_values.len(), 1);
+        assert_eq!(
+            cookie_values[0].as_bytes(),
+            b"uaid=abc123def456; MSCC=NR; MUID=1234567890ABCDEF; VAL1=ASD=DSA&HASH=41&LV=41&V=4&LU=41; empty=",
+        );
+    }
```
ntrippar, line 362: I added the cases `VAL1=ASD=DSA&HASH=41&LV=41&V=4&LU=41` and `empty=` to be sure that we do not have any corner case there.

ntrippar:
Now I think that is ok with your proposed changes and I added also some better testing. also I did check running `just qa` also FYI, I did the force pushes to my branch mostly so I don't populate the git history with every change and in the PR there is the minimal commits.

PR #775, opened 2025-12-26
correctly apply target version when using connection pool

soundofspace:


src/http/client/builder.rs
```diff
@@ -535,13 +542,42 @@ impl<T> EasyHttpConnectorBuilder<T, HttpStage> {
     ///
     /// Use `wait_for_pool_timeout` to limit how long we wait for the pool to give us a connection
     ///
+    /// This also applies a [`RequestVersionAdapter`] layer to make sure that request versions
+    /// are adapted when pooled connections are used, which you almost always need, but in case
+    /// that is unwanted, you can use [`Self::with_custom_connection_pool_without_version_adapter`] instead.
```
GlenDC: Would  it not make sense to drop the `without_version_adapter` variants and just assume that people can do all that using the existing `with_custom_connection_pool`?

src/http/client/mod.rs
```diff
@@ -86,7 +89,10 @@ where
             .with_default_transport_connector()
             .with_tls_proxy_support_using_boringssl()
             .with_proxy_support()
-            .with_tls_support_using_boringssl(Some(tls_config))
+            .with_tls_support_using_boringssl_and_default_http_version(
```
GlenDC: I wonder if that really _should_ be the default? For proxies I understand. But for pure clients I would think there can be no implicit error here? That's why I didn't adapt the default version, WDYT?

src/http/client/builder.rs
soundofspace: Not convinced of that, maybe with good docs? Forgetting this layer is a really big footgun and one I often forget, so I can imagine it will happen to other people also, but agree that this approach also feels weird. What is your opinion, will probably just follow that as I dont really have strong opinion here?

src/http/client/mod.rs
soundofspace: Many people use this client all over the place, including in proxy setups where the default client just works, except for this tiny detail. So I do think this is a better "default" config

GlenDC: yeah fair enough I guess... But it does mean that if someone wants `h2` without ALPN negotation they will need to add `TargetHttpVersion` in the extensions or it will fail... I think we can perhaps solve that if we make `Version` in `Request` optional so that if someone explicitly choses `Version` it can be seen as a choice.

If we do that we would basically have as order:

```
TargetHttpVersion > request.version > default
```

GlenDC: So if we do that than I am ok you resolve this is as-is

src/http/client/builder.rs
GlenDC: I think it's a good idea that you add `RequestVersionAdapter` to the setters where we use the default pool. I think however that we do not need the `without_request_version_adapter` and neither do we need it for the custom pool. I think you are at a point where you have the knowledge to implement and use your own pool that you also can be responsible for knowing if you need the adapter or not. Especially if we add a big fat warning to the docs of the custom pool setter.

src/http/client/mod.rs
soundofspace: "h2 without ALPN negotation" this is in my opinion a use case that is not meant for the default client?

GlenDC: why not? it's called prior knowledge, very regular use case

soundofspace: Yea but with prior knowledge it's almost always without TLS? Or in that case TargetHttpVersion can be set

soundofspace: Either way will remove this default for now and we can think about it, I like the idea of version be optional also, but it needs something extra still...

PR #776, opened 2025-12-26
feat(tcp): add into_std, into_inner, and Unix FD traits to TcpListener

seuros:
Adds methods to extract the underlying file descriptor from `TcpListener`, enabling zero-downtime deployments via `SCM_RIGHTS` socket passing.                                                                                   
                                                                                                                                                                                                                                   
  ## Motivation                                                                                                                                                                                                                    
                                                                                                                                                                                                                                   
  Process supervisors can achieve zero-downtime restarts by passing listener file descriptors between old and new instances using Unix `SCM_RIGHTS`. This avoids the brief window where no process is accepting connections during deployment.
                                                                                                                                                                                                                                   
  Currently, `TcpListener` can be created from `std::net::TcpListener` but cannot be converted back. This makes it impossible to extract the FD for transfer.

### DIFF
--- a/rama-tcp/src/server/listener.rs
+++ b/rama-tcp/src/server/listener.rs
@@ -233,18 +233,20 @@ impl TcpListener {
         self.inner.ttl()
     }
 
-    /// Converts this `TcpListener` into a `std::net::TcpListener`.
+    /// Converts this [`TcpListener`] into a [`std::net::TcpListener`].
     ///
     /// The returned listener will be in blocking mode. To convert it back
     /// to non-blocking for use with Rama, use [`TryFrom<std::net::TcpListener>`].
     ///
     /// This is useful for zero-downtime restarts where listener file descriptors
-    /// need to be passed between processes via SCM_RIGHTS.
+    /// need to be passed between processes via `SCM_RIGHTS`.
     pub fn into_std(self) -> io::Result<std::net::TcpListener> {
-        self.inner.into_std()
+        let std_listener = self.inner.into_std()?;
+        std_listener.set_nonblocking(false)?;
+        Ok(std_listener)
     }
 
-    /// Consumes this `TcpListener` and returns the inner `tokio::net::TcpListener`.
+    /// Consumes this [`TcpListener`] and returns the inner [`tokio::net::TcpListener`].
     pub fn into_inner(self) -> TokioTcpListener {
         self.inner
     }
@@ -417,7 +419,7 @@ async fn handle_accept_err(err: io::Error) {
 #[cfg(unix)]
 mod unix_fd {
     use super::TcpListener;
-    use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, IntoRawFd, RawFd};
+    use std::os::unix::io::{AsFd, AsRawFd, BorrowedFd, RawFd};
 
     impl AsRawFd for TcpListener {
         fn as_raw_fd(&self) -> RawFd {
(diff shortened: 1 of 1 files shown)


rama-tcp/src/server/listener.rs
```diff
@@ -232,6 +232,22 @@ impl TcpListener {
     pub fn ttl(&self) -> io::Result<u32> {
         self.inner.ttl()
     }
+
+    /// Converts this `TcpListener` into a `std::net::TcpListener`.
```
GlenDC:
```suggestion
    /// Converts this [`TcpListener`] into a [`std::net::TcpListener`].
```

rama-tcp/src/server/listener.rs
```diff
@@ -232,6 +232,22 @@ impl TcpListener {
...
+    ///
+    /// The returned listener will be in blocking mode. To convert it back
+    /// to non-blocking for use with Rama, use [`TryFrom<std::net::TcpListener>`].
+    ///
+    /// This is useful for zero-downtime restarts where listener file descriptors
+    /// need to be passed between processes via SCM_RIGHTS.
```
GlenDC:
```suggestion
    /// need to be passed between processes via `SCM_RIGHTS`.
```

rama-tcp/src/server/listener.rs
```diff
@@ -232,6 +232,22 @@ impl TcpListener {
...
+    pub fn into_std(self) -> io::Result<std::net::TcpListener> {
+        self.inner.into_std()
+    }
+
+    /// Consumes this `TcpListener` and returns the inner `tokio::net::TcpListener`.
```
GlenDC:
```suggestion
    /// Consumes this [`TcpListener`] and returns the inner [`tokio::net::TcpListener`].
```

rama-tcp/src/server/listener.rs
```diff
@@ -232,6 +232,22 @@ impl TcpListener {
...
+
+    /// Converts this `TcpListener` into a `std::net::TcpListener`.
+    ///
+    /// The returned listener will be in blocking mode. To convert it back
+    /// to non-blocking for use with Rama, use [`TryFrom<std::net::TcpListener>`].
+    ///
+    /// This is useful for zero-downtime restarts where listener file descriptors
+    /// need to be passed between processes via SCM_RIGHTS.
+    pub fn into_std(self) -> io::Result<std::net::TcpListener> {
+        self.inner.into_std()
```
GlenDC:
```suggestion
        let std_listener = self.inner.into_std()?;
        std_listener.set_nonblocking(false)?;
        Ok(std_listener)
```

rama-tcp/src/server/listener.rs
```diff
@@ -397,3 +413,31 @@ async fn handle_accept_err(err: io::Error) {
...
+        }
+    }
+
+    impl AsFd for TcpListener {
+        fn as_fd(&self) -> BorrowedFd<'_> {
+            self.inner.as_fd()
+        }
+    }
+
+    impl IntoRawFd for TcpListener {
```
GlenDC: I would not implement this. As we are forcing here a panic.
The user can easily do 

```
self.inner
                .into_std()
                .expect("into_std for IntoRawFd")
                .into_raw_fd()
```

themselves and at least they would have the choice to handle the error, while here we force the hand of the user in silence. Into stuff should never be fallible.

Please remove `impl IntroRawFd` here. The rest can stay.

GlenDC (changes requested):
Awesome. Looks good. Just a couple of remarks. Once those are resolved I am happy to merge this in.

GlenDC:
Also @seuros , run `just qq` please. Seems you have clippy errors.

GlenDC (approved):
Perfect, great contribution!

PR #778, opened 2025-12-26
chore(tcp): add examples for into_std, into_inner, and Unix FD traits

seuros:
Add two examples demonstrating the new TcpListener functionality:

- tcp_listener_unix_fd_reuse.rs: Shows Unix FD reuse via AsRawFd/AsFd traits for zero-downtime restarts using SCM_RIGHTS
- tcp_listener_zero_downtime.rs: Demonstrates into_std() and TryFrom<std::net::TcpListener> for FD preservation across restarts

Both examples include proper documentation in Cargo.toml, examples README, and the rama book.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -430,6 +430,7 @@ h2-support = { path = "rama-http-core/tests/h2-support" }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 itoa = { workspace = true }
+libc = "0.2"
 parking_lot = { workspace = true }
 pin-project-lite = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
@@ -648,6 +649,10 @@ required-features = ["http-full", "boring"]
 [[example]]
 name = "tls_sni_router"
 required-features = ["http-full", "boring"]
+
+[[example]]
+name = "tcp_listener_fd_passing"
+required-features = ["http-full"]
 
 [[example]]
 name = "tcp_listener_hello"
--- a/examples/tcp_listener_fd_passing.rs
+++ b/examples/tcp_listener_fd_passing.rs
@@ -1,0 +1,287 @@
+//! TCP listener FD passing ("teleport") - real zero-downtime restart via SCM_RIGHTS
+//!
+//! Demonstrates actual file descriptor transfer between parent and child processes
+//! using Unix domain sockets and SCM_RIGHTS. This technique, sometimes called "teleporting"
+//! a socket, allows the parent process to create a listener, spawn a child, transfer the FD,
+//! and have the child take over serving without dropping connections.
+//!
+//! # Run the example
+//!
+//! ```sh
+//! cargo run --example tcp_listener_fd_passing --features=http-full
+//! ```
+//!
+//! # Expected output
+//!
+//! - Parent binds to 127.0.0.1:62046
+//! - Parent spawns child process
+//! - Parent sends listener FD via Unix socket
+//! - Child receives FD and starts serving
+//! - Parent gracefully exits
+//! - Test with: `curl http://127.0.0.1:62046`
+//!
+//! This only works on Unix systems (Linux, macOS, BSDs).
+
+#[cfg(target_family = "unix")]
+mod unix_example {
+    use std::{
+        convert::Infallible,
+        env, io,
+        os::unix::io::{AsRawFd, FromRawFd, RawFd},
+        process::{Command, Stdio},
+        time::Duration,
+    };
+
+    use rama::{
+        error::BoxError, graceful::Shutdown, http::Request, http::server::HttpServer,
+        http::service::web::response::Json, rt::Executor, service::service_fn,
+        tcp::server::TcpListener,
+    };
+    use serde_json::json;
+
+    const SOCKET_PATH: &str = "/tmp/rama_fd_passing.sock";
+    const ROLE_ENV: &str = "RAMA_FD_ROLE";
+
+    pub(crate) fn run() {
+        // Check if we're parent or child
+        match env::var(ROLE_ENV).as_deref() {
+            Ok("child") => {
+                // Child process: receive FD and serve
+                tokio::runtime::Runtime::new()
+                    .unwrap()
+                    .block_on(child_process())
+                    .unwrap();
+            }
+            _ => {
+                // Parent process: create listener, spawn child, transfer FD
+                tokio::runtime::Runtime::new()
+                    .unwrap()
+                    .block_on(parent_process())
+                    .unwrap();
+            }
+        }
+    }
+
+    async fn parent_process() -> Result<(), BoxError> {
+        tracing_subscriber::fmt::init();
+
+        println!("=== Parent Process ===");
+        println!("Creating TCP listener...");
+
+        // Create listener
+        let listener = TcpListener::bind("127.0.0.1:62046").await?;
+        let addr = listener.local_addr()?;
+        println!("✓ Listening on {addr}");
+
+        // Clean up old socket file
+        let _ = std::fs::remove_file(SOCKET_PATH);
+
+        // Create Unix socket for FD passing (std blocking socket required for libc sendmsg)
+        let unix_listener = std::os::unix::net::UnixListener::bind(SOCKET_PATH)?;
+        println!("✓ Created control socket at {SOCKET_PATH}");
+
+        // Convert to std listener for FD passing
+        let std_listener = listener.into_std()?;
+        println!("✓ Converted to std::net::TcpListener");
+
+        // Spawn child process
+        println!("\nSpawning child process...");
+        let exe = env::current_exe()?;
+        let mut child = Command::new(exe)
+            .env(ROLE_ENV, "child")
+            .stdin(Stdio::null())
+            .stdout(Stdio::inherit())
+            .stderr(Stdio::inherit())
+            .spawn()?;
+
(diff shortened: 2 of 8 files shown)


examples/tcp_listener_zero_downtime.rs
```diff
@@ -0,0 +1,87 @@
...
+//! cargo run --example tcp_listener_zero_downtime --features=tcp,http-full
+//! ```
+//!
+//! # Expected output
+//!
+//! The server will simulate a zero-downtime restart by converting the listener to std,
+//! then back to rama. You can test with:
+//!
+//! ```sh
+//! curl http://127.0.0.1:62001
```
GlenDC: port `62001` is already taken by another example, you want to go up and take the first one available. E.g. for your first example you can choose port `62046` and the other one port `62047`, afaik these are both still available.

This is important as it avoids port conflicts in the current test setup

examples/tcp_listener_zero_downtime.rs
```diff
@@ -0,0 +1,87 @@
+//! Zero-downtime restart example using TCP listener FD passing.
+//!
+//! This example demonstrates how to use `into_std()` and `TryFrom<std::net::TcpListener>`
+//! to pass file descriptors between processes for zero-downtime restarts.
+//!
+//! # Run the example
+//!
+//! ```sh
+//! cargo run --example tcp_listener_zero_downtime --features=tcp,http-full
```
GlenDC: I guess you copied this from other places, but know that `http-full` already enables `tcp`. So you can just depend on `http-full` in all your feature declarations of your PR and it will be fine

examples/tcp_listener_zero_downtime.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    // Simulate zero-downtime restart flow
+    println!("\n=== Simulating zero-downtime restart ===\n");
+
+    // 1. Convert rama listener to std::net::TcpListener (blocking mode)
+    // This is what you'd do before serializing the FD for passing to new process
+    let std_listener = listener.into_std()?;
+    println!("✓ Converted to std::net::TcpListener (blocking mode)");
+
+    // 2. In real scenario, you would:
```
GlenDC: I would either drop this example or do the real thing. I suggest something like the following:

Have the example run either in orchestrator mode or runner mode

- orchestrator spawns a child: if that childed was started with an FD it will "reuse it" otherwise it will create a new one
- orchestrator monitors the process, if it is killed it starts a new process
- for the server (Ran by the child) have perhaps a long running `SSE` endpoint which counts up, 1 per second, and also 1 method `kill` which kills itself

If this really works with 0 downtime your SSE endpoint should continue to function despite having swiched process...

WDYT?

examples/tcp_listener_unix_fd_reuse.rs
```diff
@@ -0,0 +1,111 @@
...
+        let listener1 = TcpListener::bind("127.0.0.1:62002").await?;
+        let addr = listener1.local_addr()?;
+        println!("First listener bound to {addr}");
+
+        // Extract raw FD using AsRawFd trait
+        let raw_fd = listener1.as_raw_fd();
+        println!("Raw FD: {raw_fd}");
+
+        // Extract borrowed FD using AsFd trait (modern approach)
+        let borrowed_fd = listener1.as_fd();
```
GlenDC: I think you can drop this example and just do one. Don't think there's a point in having 2 examples for basially the same concept. And also I would only add an example if you make it real. I do not have a single example where I do not demonstrate the concept for real, and neither do I want to start with that.

GlenDC (changes requested):
Let me know if you want to still continue adding the example given my feedback. If so make sure to also add an e2e test for it in https://github.com/plabayo/rama/tree/main/tests/integration/examples/example_tests.

examples/tcp_listener_zero_downtime.rs
seuros: The orchestrator is pretty big and complex , i will opensource the full implementation on alpha.4 is out.

We can have it as example, if oyu want.

Cargo.toml
```diff
@@ -430,6 +430,7 @@ h2-support = { path = "rama-http-core/tests/h2-support" }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 itoa = { workspace = true }
+libc = "0.2"
```
seuros, line 433: I added this dev dependency. 1.0.0 alpha exist.

examples/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,289 @@
...
+        Ok(())
+    }
+
+    /// Send a file descriptor via Unix domain socket using SCM_RIGHTS
+    fn send_fd(stream: &UnixStream, fd: RawFd) -> io::Result<()> {
+        let sock_fd = stream.as_raw_fd();
+
+        // Prepare iovec with dummy byte
+        let dummy = [b'F'];
+        let mut iov = libc::iovec {
```
GlenDC, line 191: Why do you need to use libc for this? We already have support for unix sockets under `rama::unix`. Do you think you miss something there that forces you to go to C? Ideally I would prefer you stick to a safe rust w/ rama approach here.

examples/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,289 @@
...
+        // Create listener
+        let listener = TcpListener::bind("127.0.0.1:62046").await?;
+        let addr = listener.local_addr()?;
+        println!("✓ Listening on {addr}");
+
+        // Clean up old socket file
+        let _ = std::fs::remove_file(SOCKET_PATH);
+
+        // Create Unix socket to communicate with child
+        let unix_listener = std::os::unix::net::UnixListener::bind(SOCKET_PATH)?;
```
GlenDC: use `rama::unix` please. This one also ensures to do socket cleanup.

tests/integration/examples/example_tests/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,50 @@
...
+#[ignore]
+async fn test_tcp_listener_fd_passing() {
+    utils::init_tracing();
+
+    let runner = utils::ExampleRunner::interactive("tcp_listener_fd_passing", None);
+
+    // Retry until child process is serving (FD handoff takes variable time)
+    let response = async {
+        for i in 0..8 {
+            match runner.get("http://127.0.0.1:62046").send().await {
```
GlenDC: The runner already has a retry layer applied to to it (https://github.com/plabayo/rama/blob/fa0f41e68d7e06fe116565da808ab4810aaa5516/tests/integration/examples/example_tests/utils/mod.rs#L188-L198), did it not work for you without this manual retry logic?

tests/integration/examples/example_tests/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,50 @@
...
+                    tokio::time::sleep(std::time::Duration::from_millis(200 + 150 * i)).await;
+                }
+            }
+        }
+        None
+    }
+    .await
+    .expect("child process to become ready");
+
+    let first = response.try_into_json::<Value>().await.unwrap();
```
GlenDC: Work with `serde` and deserialize in a struct please, better practice than to work with a `Value`

tests/integration/examples/example_tests/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,50 @@
...
+    assert_eq!(first["message"], "Hello from child process!");
+    assert_eq!(first["zero_downtime"], true);
+    let pid = first["pid"].as_u64().expect("pid should be a number");
+    assert!(pid > 0);
+
+    // Second request to verify same child process is still serving
```
GlenDC: what's the value of the 2nd request? What does it proof more? Is there something we do that might make the 2nd request not work, given the first 1 was also already served by the child process, no?

Cargo.toml
GlenDC: don't think we have an actual need for it for this PR, see other comment.

examples/tcp_listener_fd_passing.rs
seuros: rama::unix provides UnixListener/UnixStream for socket creation and cleanup, but doesnt have SCM_RIGHTS support for FD passing.   

std::os::unix::net::SocketAncillary is still nightly-only.

 libc is the lowest library that provide that . Nix is a option, but it bring it own ecosystem.

GlenDC: Fair enough. I guess we'll have to wait on either `std` or `socket2` (https://github.com/rust-lang/socket2/issues/614) to support it. I guess latter will be first. Can you add a comment describing that and track an issue for it in rama that you also link here next to the comment. So I can keep track of it.

tests/integration/examples/example_tests/tcp_listener_fd_passing.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+#[tokio::test]
+#[ignore]
+async fn test_tcp_listener_fd_passing() {
+    utils::init_tracing();
+
+    let runner = utils::ExampleRunner::interactive("tcp_listener_fd_passing", None);
+
+    // Runner has built-in retry with exponential backoff
```
GlenDC:
```suggestion
```

No need to comment that, that's just a detail

GlenDC (commented):
Looks pretty good now some nitpicks, and if that is resolved + CI passes we can merge!

PR #779, opened 2025-12-26
prepare rama 0.3.0-alpha.4 release notes

GlenDC:


CHANGELOG.md
```diff
@@ -5,6 +5,119 @@ All notable changes to this project will be documented in this file.
...
+
+* **Context Type Removal**: The `Context` object is no longer passed through services. Data, state, and executors are now managed via **Extensions** attached directly to the `Request`, `Response`, or `Connection` (#711, #714).
+* **Forked HTTP Modules**: Integrated and forked parts of the `http` crate's request/response modules into Rama to provide native support for our specific `Extension` requirements and performance optimizations (#696).
+* **Extensions-First Design**: Extensions are now supported throughout `rama-http-core` for both HTTP/1.1 and HTTP/2 (#706, #727).
+* **State Migration**: State management has moved from `Context` to `Extensions`. New utilities like `AddInputExtensionLayer`, `AddOutputExtensionLayer` and `IntoEndpointServiceWithState`
+have been introduced to handle state injection (#685, #720).
+
+During this process we playeda t some point with the concept of **Parent Extensions** (#715).
```
soundofspace: Typo?

CHANGELOG.md
```diff
@@ -5,6 +5,119 @@ All notable changes to this project will be documented in this file.
...
+#### Input/Output
+
+the `Service` trait now uses the type parameters:
+
+- `Input`, instead of `Request`;
+- `Output`, instead of `Response`;
+
+This aligns much better with `Rama`'s future but even for its present.
+For example service and layers also operate on the transport and tls layers,
+where the parameter types `Request` and `Response` make noe sense.
```
soundofspace:
```suggestion
where the parameter types `Request` and `Response` make no sense.
```

CHANGELOG.md
```diff
@@ -5,6 +5,119 @@ All notable changes to this project will be documented in this file.
...
+    * **Redirects**: Added redirect/rewrite/forward HTTP services and layers (#717).
+* **TLS & ACME**:
+    * **DNS Challenges**: Support for **DNS-01 challenges** in `rama-acme` and TXT record lookups in `rama-dns`.
+    * **Wildcard Support**: Added support for wildcard domains in `rama-net` and better wildcard handling in the `DynamicIssuer`.
+    * **ACME Certs**: Support for downloading ACME certificates as raw PEM bytes.
+
+### Changed
+
+* **API Naming Conventions**:
+    * Applied consistent `try_` prefix for fallible setters and constructors (e.g., `try_port`, `try_host`).
```
soundofspace: Do we not always use try_set_port or try_with_port, maybe stupid detail for here, but ...

GlenDC: the set/with is another convention, you can have both conventions combined ofc. Perhaps I should also add a point about the setter macros and convention that goes with it?

CHANGELOG.md
GlenDC: yup "planned at" it should be

CHANGELOG.md
soundofspace: Not really sure that belongs here, can be short point indeed. But would also not mind skipping it here, just mentioned it since try_port seemed like a setter to me

CHANGELOG.md
soundofspace: seems it should be played at? planed at also sounds wrong in this sentence

PR #780, opened 2025-12-27
Health check logic for pooled connections

soundofspace:


soundofspace:
@GlenDC this is a rough first implementation that can handle the reported issues, but it can also be extended in the future to support async health checks such as doing a db query. 

There are some todo's scattered all over which are mainly questions, when you have the time can you check if you agree with this approach, and if you have opinions on those questions. Also I went async first for this to support everything, but for performance reasons, we could also only support sync here, potentially with async after different trait?

Currently this approach assumes that all connectors/connections that will be used with a connection pool implement this healtcheck trait, which is pretty easy for us to do for all connectors and we can also provide a "connector wrapper" which adds this to any "connection" and assumes the connection is always ok

rama-net/src/client/pool/mod.rs
```diff
@@ -716,58 +746,77 @@ where
...
-                    permit
-                }
-            }
-        };
+            let EstablishedClientConnection { input, conn } =
+                self.inner.connect(input).await.map_err(Into::into)?;
 
-        let EstablishedClientConnection { input, conn } =
-            self.inner.connect(input).await.map_err(Into::into)?;
+            // TODO here we create a new connection, but do we also want to health check this for consistency,
```
GlenDC: I think it's fair to assume that the inner connection only returns a solid connector back? Because otherwise even if it passes _now_, what about 1 sec later 🤷

GlenDC (commented):
I'm fine with this. Also let's just keep it purely async. There is not really a cost for async if you do not need to actually wait afaik. Compared to IO especially no issue imho. Keeps it simpler.

Not sure if we should force OpaqueError or attach an error type to the trait. But then again, perhaps keeps it simpler to just have it opaqueError. So no strong opinion about that one.

Anything you question? As otherwise all good for me to continue with this appraoch.
What's the solution to make it work with layered connectors that are not connectors themselves?

soundofspace:
@GlenDC do you agree with this new approach todo a health check(s)? Only focus on this health check part, still need to finish the pool logic (but rough code is also ready in case you are interested).

Some questions also for you:
- OpaqueError, or BoxError, or do we need custom one for this (dont think so)
- How to limit this in pooled connector and do we even limit this. I'm thinking max attemps to get a connection / timeout on individual or total health check duration, or only global timeout on pool...
- Do we run all health checks in parallel or sequential. Speed vs doing too much work?
- Having support for boxed service to do healthy is really nice since now this can also be used for things such as database connection where you do an active health check, or some advanced logic is needed .But it also introduces all the problems i mentioned above that now we need to probably put limits on this. So while its really nice that we support use cases like that, do we want to, I'm thinking yes, but curious about your opinion

rama-core/src/extensions.rs
```diff
@@ -31,6 +31,8 @@ use std::any::{Any, TypeId};
 use std::pin::Pin;
 use std::sync::Arc;
 
+use futures::stream;
```
GlenDC:
```suggestion
use crate::futures::stream;
```

rama-core/src/extensions.rs
```diff
@@ -89,6 +97,35 @@ impl Extensions {
             .and_then(|ext| (*ext.1).as_any().downcast_ref())
     }
 
+    /// Get a shared reference to the most recently insert item of type T, or insert in case no item was found
+    ///
+    /// Note: [`Self::get`] will return the last added item T, in most cases this is exactly what you want, but
+    /// if you need the oldest item T use [`Self::first`]
+    pub fn get_or_insert<T, F>(&mut self, create_fn: F) -> &T
+    where
+        T: Clone + Send + Sync + std::fmt::Debug + 'static,
```
GlenDC, line 106: The `Extension` (empty for now) trait is for another PR I suppose where you do the actual extension refactor?

rama-core/src/extensions.rs
```diff
@@ -118,6 +155,13 @@ impl Extensions {
             .filter(move |item| item.0 == type_id)
             .map(|ext| (*ext.1).as_any().downcast_ref().unwrap())
     }
+
+    /// Stream iterator over all the inserted items of type T
+    ///
+    /// Note: items are ordered from oldest to newest
+    pub fn stream_iter<T: Send + Sync + 'static>(&self) -> stream::Iter<impl Iterator<Item = &T>> {
```
GlenDC, line 162: Why is this called stream iter? We are not actually streaming, right??? I mean it's a regular iter atm?

rama-net/src/conn.rs
```diff
@@ -17,3 +23,45 @@ pub fn is_connection_error(e: &io::Error) -> bool {
...
+        f.debug_struct("ConnectionHealth")
+            .field("status", &self.status())
+            .finish()
+    }
+}
+
+#[repr(u8)]
+#[derive(Debug, PartialEq, Clone, Copy, Eq)]
+pub enum ConnectionHealthStatus {
+    Unknown = 0,
```
GlenDC, line 64: What does the unknown matter? Wouldn't that just be anyway safely assumed as broken anyway?

GlenDC: Would have thought this can just be a binary boolean?

tests/integration/client/mod.rs
```diff
@@ -1,20 +1,24 @@
...
-            server::SelfSignedData,
-            server::{ServerAuth, ServerConfig},
+            server::{SelfSignedData, ServerAuth, ServerConfig},
         },
     },
     rt::Executor,
     service::service_fn,
     tls::boring::{client::TlsConnectorDataBuilder, server::TlsAcceptorLayer},
 };
+use rama_http::{HeaderValue, Version, header};
```
GlenDC: try to keep these grouped, e.g. move the http ones above

rama-core/src/extensions.rs
soundofspace: Yes this pr was only focussed on the health check logic, with this addition to make that a bit more straightforward

rama-core/src/extensions.rs
soundofspace: It a "stream" created from an iterator. So i'ts not really async but its still a stream, which means all the stream utils/traits can be used on it

rama-net/src/conn.rs
soundofspace: By also having unknown we can treat things can also customize logic on how to handle timeout. In that case we dont really have failed, so we can customize logic. With the new extension logic this will look much nicer, but still I would prefer to keep it like this in case that one is delayed for whatever reason, and I need to unknown state

rama-net/src/client/pool/mod.rs
```diff
@@ -349,7 +339,27 @@ where
...
+            if let Some(health) = pooled_conn.extensions().get::<ConnectionHealth>()
+                && health.status() == ConnectionHealthStatus::Broken
+            {
+                continue;
+            }
+
+            return Some((idx, pooled_conn));
+        };
+
+        // TODO since we are dropping connections does it still make sense to report idx in metrics?
```
soundofspace: @GlenDC

rama-net/src/conn.rs
GlenDC: ok fair enough

rama-net/src/client/pool/mod.rs
GlenDC: where do we report index in metrics?

PR #784, opened 2025-12-31
feat: add datastar v1.0.0-RC.7 sourcemap with convenience helper

seuros:
the other PR need merging first. , then i will rebase.


rama-http/src/service/web/endpoint/response/datastar.rs
```diff
@@ -32,3 +33,66 @@ impl IntoResponse for DatastarScript {
...
+/// Embedded version of the datastar source map file
+/// for improved debugging experience.
+///
+/// Learn more about datastar at <https://data-star.dev/>
+/// or in the book: <https://ramaproxy.org/book/web_servers.html#datastar>.
+pub struct DatastarSourceMap(Script<&'static str>);
+
+impl Default for DatastarSourceMap {
+    fn default() -> Self {
+        Self(Script(include_str!("./datastar.js.map")))
```
GlenDC: can you add a comment above this for the link where you got this map from? I also forgot to do that myself I think for the actual js file. Can you do that for that line as well? Easier to maintain and validate that way.

rama-http/src/service/web/endpoint/response/datastar.rs
```diff
@@ -32,3 +33,66 @@ impl IntoResponse for DatastarScript {
...
+
+#[derive(Debug, Clone, Copy)]
+/// Datastar source map ready to be served.
+///
+/// Embedded version of the datastar source map file
+/// for improved debugging experience.
+///
+/// Learn more about datastar at <https://data-star.dev/>
+/// or in the book: <https://ramaproxy.org/book/web_servers.html#datastar>.
+pub struct DatastarSourceMap(Script<&'static str>);
```
GlenDC: I would store this just as a `&'static str` and within your IntoResponse impl you'll want to do this:

```
(Headers::single(ContentType::json()), self.0).into_response();
```

Something like that, don't know the exact syntax. But the point is that a sourcemap is json, not javascript.

rama-http/src/service/web/endpoint/response/datastar.rs
```diff
@@ -32,3 +33,66 @@ impl IntoResponse for DatastarScript {
...
+/// ```rust,ignore
+/// use rama::http::service::web::{Router, response::datastar_assets};
+///
+/// let router = datastar_assets(Router::new(), "/assets");
+/// ```
+///
+/// This will serve:
+/// - `/assets/datastar.js` - The datastar script
+/// - `/assets/datastar.js.map` - The source map for debugging
+pub fn datastar_assets<State>(router: Router<State>, prefix: &str) -> Router<State>
```
GlenDC: please remove this and also the use of it, it goes a bit against the principles we adhere to in rama abstractions. This adds to little value (if anything) to warrant a new level of abstraction. And revert the use from before.

examples/http_sse_datastar_hello.rs
```diff
@@ -109,11 +109,13 @@ async fn main() {
...
-        let app = Router::new_with_state(controller.clone())
-            .with_get("/", handlers::index)
-            .with_post("/start", handlers::start)
-            .with_get("/hello-world", handlers::hello_world)
-            .with_get("/assets/datastar.js", DatastarScript::default());
+        let app = datastar_assets(
+            Router::new_with_state(controller.clone())
+                .with_get("/", handlers::index)
+                .with_post("/start", handlers::start)
+                .with_get("/hello-world", handlers::hello_world),
```
GlenDC: as requested, you'll just have the assets maually defined here, the js file as well as the sourcemap. For the sourcemap I would recommend to add a `#[cfg(debug_assertions)]` on top so that the source map is only included in debug builds (not in release/production builds). Good to do such things in examples.

rama-http/src/service/web/endpoint/response/datastar.rs
seuros: The map is literally in the same CDN. 

It reference in the last line https://github.com/plabayo/rama/commit/ad5b48b291e0e7ed9964c53062123a884624e3c7#diff-722f04175826144c3de61c0e9b93ab99b55df3e531e29785f63e1c5bbedba630R2-R9

We could clean it , but you will have to remember to do it everytime you fetch it from the CDN.

seuros: The source map is not used only in development...
When your javascript crash you dont see  a  🖕 and wonder what the hell it that, you see the exact source code.

PS: I'm not trolling,  🖕 is part of D* 🚀  source code . i love it.

GlenDC: No I wouldn't clean it, just keep it as-is.

GlenDC: Rama itself does not prevent you from also serving the source map in production. I am just saying please in the rama example do gate that behind that flag. Doesn't stop anyone from doing it differently. As I personally do not ever serve a source map myself.

Issue #488, opened 2025-04-01
add GRPC support as rama-grpc


GlenDC:
Can probably be a fork of tonic, but to ensure we do not need to rely on an entire different stack, but can instead build upon rama's building blocks, while still benefiting from a wider ecosystem.

Blocked until later in time, when a clear plan of what needs to happen and how is defined. Free for anyone to already provide feedback and input however.

GlenDC:
The work for gRPC is now a priority ticket with the developer hours funded by a commercial partner.
Agreed deadline is ~ New Year.

GlenDC:
:loudspeaker: gRPC is now supported in Rama via the new `rama-grpc` and `rama-grpc-build` crates. It is also available through feature flags in the mono crate as `rama::grpc`.

Highlights:
* Available for both clients and servers
* Middleware support
* Built in code generation
* grpc-web support, allowing gRPC over HTTP/1.1
* Uses Rama transports under the hood. On the client side this can be as simple as an `EasyHttpWebClient`, while any HTTP or HTTPS service works as well. This also means you can use `rama::unix` to run gRPC over Unix sockets, or any custom network stack you build in Rama

You can find examples to try out here:
https://github.com/plabayo/rama/tree/main/examples/grpc

Feedback is very welcome, especially from anyone using this in production.
This work was funded, and we are very grateful to our commercial partners who help unlock valuable new features like this :pray: :tada:

andreystepanov:
@GlenDC I suggest you reserve `rama-grpc` and `rama-grpc-build` (maybe some others you forgot to publish) crates, because now they can be claimed by anyone:

- https://crates.io/crates/rama-grpc
- https://crates.io/crates/rama-grpc-build

The are also linked from the book, which is how I discovered it.

Hope that helps 😉

GlenDC:
@andreystepanov thank you, have done so now.
turns out there is a tool for that: cargo-squatty

Issue #489, opened 2025-04-01
provide rama trait implementations for rama


GlenDC:
Can live as `rama-opentelemetry` crate which we can expose also as `rama::opentelemetry` and which depends on crates like `rama-grpc` and/or `rama-http-backend` dependent on what features are enabled.

Would implement traits such as:

- https://docs.rs/opentelemetry-http/0.29.0/opentelemetry_http/trait.HttpClient.html
- https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/metrics/exporter/trait.PushMetricExporter.html

etc...

Blocked until a more detailed plan is provided and this is deemed ready to go. Feedback and input is already welcome.

Goal of this story is to ensure usage of openetelemetry doesn't mean you need to bring in additional http stacks such as hyper and beyond, as that is a bit silly and might even give issues in certain combinations.

This is also the last remaining (dev) dependency of hyper and friends in the rama (dev) tree. Already much longer this is the case for non-dev dependencies, but for dep this is the last remaining usage.

GlenDC:
Gonna start already on the first part: provide the Http client, and use it in `rama-cli`, as this is right now the only reason why we pull in `reqwest`:

```
reqwest v0.12.20
├── opentelemetry-http v0.30.0
│   └── opentelemetry-otlp v0.30.0
│       [dev-dependencies]
│       └── rama v0.3.0-alpha.1 (/Users/glendc/code/github.com/plabayo/rama)
└── opentelemetry-otlp v0.30.0 (*)
```

as well as `hyper` and friends:

```
cargo tree -i hyper
hyper v1.6.0
├── hyper-timeout v0.5.2
│   └── tonic v0.13.1
│       ├── opentelemetry-otlp v0.30.0
│       │   [dev-dependencies]
│       │   └── rama v0.3.0-alpha.1 (/Users/glendc/code/github.com/plabayo/rama)
│       └── opentelemetry-proto v0.30.0
│           └── opentelemetry-otlp v0.30.0 (*)
├── hyper-util v0.1.14
│   ├── hyper-timeout v0.5.2 (*)
│   ├── reqwest v0.12.20
│   │   ├── opentelemetry-http v0.30.0
│   │   │   └── opentelemetry-otlp v0.30.0 (*)
│   │   └── opentelemetry-otlp v0.30.0 (*)
│   └── tonic v0.13.1 (*)
├── reqwest v0.12.20 (*)
└── tonic v0.13.1 (*)
```

[...]

PR #790, opened 2026-01-04
add gRPC support to rama

GlenDC:
Closes https://github.com/plabayo/rama/issues/488

PR Checklist:

- [x] add grpc impl w/ build/codegen capabilities (client + server)
- [x] web-grpc (mostly server focussed)
- [x] add examples for the grpc code + tests
- [x] test grpc generated code (can be without actual transport layer)
- [x] add generated health service
- [x] support grpc interceptors
- [x] improve origin code
- [x] check what to do with utilities like the timeout
- [x] use and test the grpc tracing support
- [x] add tests under rama-grpc/tests
- [x] add grpc router

Outside of scope of this PR, but that will be tackled still later this month:

* proxy'ng grpc traffic
* decent client support for web-grpc + related RPC-adjacent protocols
* providing layers and utilities to easily:
   * combine multiple services (at server side)
* provide ability to easily use things like web-grpc in case http/1 is used ,vs regular grpc over h2
* OTLP grpc service (https://github.com/plabayo/rama/issues/489)

### DIFF
--- a/rama-grpc-build/src/manual.rs
+++ b/rama-grpc-build/src/manual.rs
@@ -2,7 +2,7 @@
 //! purely in Rust without the need of `proto` files. It also enables you to set a custom `Codec`
 //! if you want to use a custom serialization format other than `protobuf`.
 
-use crate::code_gen::CodeGenBuilder;
+use crate::CodeGenBuilder;
 
 use proc_macro2::TokenStream;
 use quote::ToTokens;


docs/thirdparty/fork/README.md
```diff
@@ -29,6 +29,11 @@ as a distant relative.
     - Original: <https://github.com/hyperium/headers/blob/de0b1a1e97d20f3667a346c4d5b5973d92ab58f9/LICENSE>
     - Type: MIT
     - Copy: [./licenses/headers](./licenses/headers)
+- h2: <https://github.com/hyperium/tonic/tree/88a448a2fdedf06340deac645a061120a2612537>
```
soundofspace:
```suggestion
- tonic: <https://github.com/hyperium/tonic/tree/88a448a2fdedf06340deac645a061120a2612537>
```

GlenDC:
Ready for you to review @soundofspace.

examples/grpc/src/shared/tests/examples/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+    net::client::EstablishedClientConnection,
+    service::{BoxService, service_fn},
+};
+
+mod health;
+mod helloworld;
+
+pub(super) type WebClient = BoxService<http::Request, http::Response, OpaqueError>;
+
+// TODO: might make sense in future to turn this into a general utility,
```
soundofspace, line 14: I also have other use case where this would indeed by helpful. But I do think for this use case here MockConnector is probably a better choice (will discuss this on discord).

Fyi this is what I'm using now to help me there, basically the same but with a struct
```rust
struct IoToConnector<S> {
    inner: Arc<Mutex<Option<S>>>,
}

impl<S> IoToConnector<S> {
    fn new(io: S) -> Self {
        Self {
            inner: Arc::new(Mutex::new(Some(io))),
        }
    }
}

impl<S, Request> Service<Request> for IoToConnector<S>
where
    S: Stream + ExtensionsMut + Unpin,
    Request: Send + 'static,
{
    type Error = Infallible;
    type Output = EstablishedClientConnection<S, Request>;

    async fn serve(&self, input: Request) -> Result<Self::Output, Self::Error> {
        let stream = self.inner.lock().take().unwrap();
        Ok(EstablishedClientConnection {
            input,
            conn: stream,
        })
    }
}
```

examples/grpc/src/shared/tests/examples/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+// for testing or generic one-time stuff. Be it perhaps with an actual error
+// returned incase of duplicate consumptions.
+
+fn mock_io_client(client: tokio::io::DuplexStream) -> WebClient {
+    let client_opt = Arc::new(Mutex::new(Some(client)));
+    rama::Service::boxed(
```
soundofspace: why `rama::Service::boxed` vs just chaining `.boxed()`?

examples/grpc/src/shared/tests/integration/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+mod client_layer;
+mod http2_keep_alive;
+mod http2_max_header_list_size;
+mod max_message_size;
+mod streams;
+mod timeout;
+
+pub(super) type WebClient = BoxService<http::Request, http::Response, OpaqueError>;
+
+fn mock_io_client(client: tokio::io::DuplexStream) -> WebClient {
```
soundofspace: same remark about MockConnector

examples/grpc/src/shared/tests/integration/streams.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+
+    let svc = test_stream_server::TestStreamServer::new(Svc);
+
+    let _ = HttpServer::h2(Executor::new()).service(svc);
+}
+
+#[allow(dead_code)]
+struct SyncStream(*mut ());
```
soundofspace: What is the goal of this test with pointers and unimplemented?

rama-grpc-build/src/protobuf/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+    compile_well_known_types: bool,
+    codec_path: String,
+    root_crate: TokenStream,
+    disable_comments: HashSet<String>,
+}
+
+impl ServiceGenerator {
+    /// Create a new ServiceGenerator
+    #[allow(clippy::too_many_arguments)]
+    fn new(
```
soundofspace: Why not just make things pub on struct if you just pass all of them in?

rama-grpc-build/src/protobuf/mod.rs
```diff
@@ -0,0 +1,925 @@
...
+        if let Some(path) = &self.file_descriptor_set_path {
+            config.file_descriptor_set_path(path);
+        }
+
+        if self.skip_protoc_run {
+            config.skip_protoc_run();
+        }
+
+        if self.build_client || self.build_server {
+            let service_generator = ServiceGenerator::new(
```
soundofspace: directly making struct instead of new is more readable I believe

rama-grpc-build/src/code_gen.rs
```diff
@@ -0,0 +1,99 @@
...
+
+impl CodeGenBuilder {
+    /// Create a new code gen builder with default options.
+    #[must_use]
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    /// Enable code generation to emit the package name.
+    pub fn emit_package(&mut self, enable: bool) -> &mut Self {
```
soundofspace: dont we also want generate_set_with macro here? If this is only used internally maybe not

rama-grpc/src/client/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+//! provides a codec agnostic gRPC client dispatcher and a decorated `rama`
+//! service trait.
+//!
+//! This client is generally used by some code generation tool to provide stubs
+//! for the gRPC service. Thusly, they are a bit cumbersome to use by hand.
+//!
+//! ## Concurrent usage
+//!
+//! Upon using the your generated client, you will discover all the functions
+//! corresponding to your rpc methods take `&mut self`, making concurrent
```
soundofspace: This is probably technical limitation from hyper/tower/tonic stack. Cloning works now, but is this something that we could change later down the line? Or are there any other reasons &mut self is needed

rama-grpc/src/codec/buffer.rs
```diff
@@ -0,0 +1,146 @@
...
+        if ret.len() > self.len {
+            &ret[..self.len]
+        } else {
+            ret
+        }
+    }
+
+    #[inline]
+    fn advance(&mut self, cnt: usize) {
+        assert!(cnt <= self.len);
```
soundofspace, line 42: debug assert/error or this could panic, seems similar to the work you did in http-core to avoid things like this crashing an entire thread?

rama-grpc/src/codec/buffer.rs
```diff
@@ -0,0 +1,146 @@
...
+        self.buf.advance(cnt);
+        self.len -= cnt;
+    }
+
+    #[inline]
+    fn copy_to_bytes(&mut self, len: usize) -> Bytes {
+        assert!(len <= self.len);
```
soundofspace, line 49: same

rama-grpc/src/codec/compression.rs
```diff
@@ -0,0 +1,448 @@
...
+        for encoding in self.inner.into_iter().flatten() {
+            value.put_slice(encoding.as_str().as_bytes());
+            value.put_u8(b',');
+        }
+
+        if value.is_empty() {
+            return Ok(None);
+        }
+
+        value.put_slice(b"identity");
```
soundofspace, line 67: Is this needed?

rama-grpc/src/metadata/key.rs
```diff
@@ -0,0 +1,248 @@
...
+    ///
+    /// This function panics when the static string is a invalid metadata key.
+    ///
+    /// This function requires the static string to only contain lowercase
+    /// characters, numerals and symbols, as per the HTTP/2.0 specification
+    /// and header names internal representation within this library.
+    #[must_use]
+    pub fn from_static(src: &'static str) -> Self {
+        let name = HeaderName::from_static(src);
+        if !VE::is_valid_key(name.as_str()) {
```
soundofspace, line 70: panic fine here?

rama-grpc/src/service/health/server.rs
```diff
@@ -0,0 +1,382 @@
...
+use super::pb::{HealthCheckRequest, HealthCheckResponse};
+
+/// Creates a `HealthReporter` and a linked `HealthServer` pair. Together,
+/// these types can be used to serve the gRPC Health Checking service.
+///
+/// A `HealthReporter` is used to update the state of gRPC services.
+///
+/// A `HealthServer` is a Tonic gRPC server for the `grpc.health.v1.Health`,
+/// which can be added to a Tonic runtime using `add_service` on the runtime
+/// builder.
```
soundofspace:
```suggestion
/// A `HealthServer` is a Rama gRPC server for the `grpc.health.v1.Health`,
/// which can be added to a Rama runtime using `add_service` on the runtime
/// builder.
```

rama-grpc/src/web/client.rs
```diff
@@ -0,0 +1,74 @@
...
+
+    async fn serve(&self, mut req: Request<B1>) -> Result<Self::Output, Self::Error> {
+        if req.version() != Version::HTTP_11 {
+            tracing::debug!(
+                "coercing request' and target version to HTTP1.1 (was: {:?})",
+                req.version()
+            );
+            req.extensions_mut()
+                .insert(TargetHttpVersion(Version::HTTP_11));
+            *req.version_mut() = Version::HTTP_11;
```
soundofspace: should probably use adapt_request_version()

rama-grpc/src/codegen.rs
```diff
@@ -0,0 +1,9 @@
+//! Codegen exports used by `rama-grpc-build`.
+//!
+//! Not meant to be used directly by the user
```
soundofspace, line 3: doc(hidden) maybe?

rama-grpc/src/request.rs
```diff
@@ -0,0 +1,386 @@
...
+
+fn duration_to_grpc_timeout(duration: Duration) -> Option<SmolStr> {
+    fn try_format<T: Into<u128>>(
+        duration: Duration,
+        unit: char,
+        convert: impl FnOnce(Duration) -> T,
+    ) -> Option<SmolStr> {
+        // The gRPC spec specifies that the timeout most be at most 8 digits. So this is the largest a
+        // value can be before we need to use a bigger unit.
+        let max_size: u128 = 99_999_999; // exactly 8 digits
```
soundofspace: const for this?

rama-grpc/src/status.rs
```diff
@@ -0,0 +1,1111 @@
...
+/// associated functions.
+/// ```rust
+/// # use rama_grpc::{Status, Code};
+/// let status1 = Status::new(Code::InvalidArgument, "name is invalid");
+/// let status2 = Status::invalid_argument("name is invalid");
+///
+/// assert_eq!(status1.code(), Code::InvalidArgument);
+/// assert_eq!(status1.code(), status2.code());
+/// ```
+#[derive(Clone)]
```
soundofspace, line 47: Debug

rama-http-core/src/service.rs
```diff
@@ -7,14 +7,6 @@ use rama_http::service::web::response::IntoResponse;
 use rama_http_types::{Request, Response};
 use std::{convert::Infallible, fmt};
 
-pub trait HttpService<ReqBody>: sealed::Sealed<ReqBody> {
```
soundofspace, line 10: Nice to see this removed

examples/grpc/src/shared/tests/examples/mod.rs
GlenDC: No good reason. It worked better for my IDE intellisense at the time, but ofc both would work. Given we have a convention so far to use `.boxed()` I changed it to that.

examples/grpc/src/shared/tests/integration/streams.rs
GlenDC: Purely to see if it can compile. But honeslty in its current ported form it makes little sense. The original code used it I think to test if unsynced streams also work. But as we do require sync it matters little. Will remove.

rama-grpc-build/src/protobuf/mod.rs
GlenDC: Agreed, the short answer is that this is OG code. But yeah I agree, so gonna clean up a bit.

rama-grpc-build/src/protobuf/mod.rs
GlenDC: Agreed, as confirmed in your definition comment. Will resolve.

rama-grpc-build/src/code_gen.rs
GlenDC: For consistency sake why not. It was also `pub` but I demoted it to `pub(crate)` as for now I do not see a good reason to expose this even if the OG was. For any usecase I am aware of you do not need this directly so we might as well hide it for now.

rama-grpc/src/client/mod.rs
GlenDC: Deleted the comment. This doesn't apply to this ported stack.
This is different for the server where we need it atm due to how it uses the codec internally, but for the client there is already no requirement on it being `&mut`. So comment is removed.

rama-grpc/src/codec/compression.rs
GlenDC: Seems not, checked the specs and also cleaned it up.

rama-grpc/src/web/client.rs
GlenDC: Agreed. Fixed it!

rama-grpc/src/codec/compression.rs
GlenDC: Nvm. I realize again why the OG code does this. It ensures that identity is always seen as a valid choice. It doesn't hurt either way.

rama-grpc/src/status.rs
GlenDC: It has that, but manually implemented.

rama-grpc/src/codegen.rs
GlenDC: Already the case

```
#[doc(hidden)]
pub mod codegen;
```

rama-grpc/src/codec/buffer.rs
GlenDC: It's the trait contract:

```
/// # Panics
    ///
    /// This function panics if `len > self.remaining()`.
```

rama-grpc/src/metadata/key.rs
GlenDC: enabled it for debug builds only
It's kinda ok as it's for static input, but yeah not ideal. For now however is ok

rama-grpc/src/codec/buffer.rs
GlenDC: Same same, part of contract.

examples/grpc/src/shared/tests/examples/mod.rs
GlenDC: MockConnector works flawlessly here indeed! thx

examples/grpc/src/shared/tests/integration/mod.rs
GlenDC: Thank you, works great!

GlenDC:
Ready for review once again @soundofspace . All feedbck is applied afaik.

PR #794, opened 2026-01-10
do not put executor in extensions

### LINKED ISSUE #789: Http should use it's own configured executor instead of the one provided in extensions
Right now `http_core_serve_connection`  does this to get the executor which is used to create a shutdownguard
```rust
            let guard = io
                .extensions()
                .get::<Executor>()
                .and_then(|exec| exec.guard())
                .cloned();
```

This works if we use something like a TcpListener which injects this, but if fails when we use HttpServer directly since it's not there. This is especially confusing for H2 where you need to provide an Executor already, but it's not used for that part.

Instead I propose that we only look at the configured executor and also require this for H1 (we already have it for auto and H2), and don't look at extensions. (I also think this is a good excuse to cleanup that Sealed trait since now we have integrated http-core and we can do this cleaner)

I also believe we should remove this from extensions in TcpListener/Unixlistener... and require the user to use something like AddExtensionLayer for this as otherwise this is really weird that it might or might not be there, and in case it is there which one is there? Tcp one, http one... By doing this manually this is all really explicit and hopefully clearer?

GlenDC:
instead have extensions be part of the structs that optionally want to make use of it

users can still inject it themselves in extensions should they ever want (or better still, static type state)

Closes #789

### DIFF
--- a/src/http/client/mod.rs
+++ b/src/http/client/mod.rs
@@ -78,8 +78,19 @@ impl<Body> Default
 where
     Body: StreamingBody<Data: Send + 'static, Error: Into<BoxError>> + Unpin + Send + 'static,
 {
-    #[cfg(feature = "boring")]
+    #[inline(always)]
     fn default() -> Self {
+        Self::default_with_executor(Executor::default())
+    }
+}
+
+impl<Body>
+    EasyHttpWebClient<Body, EstablishedClientConnection<HttpClientService<Body>, Request<Body>>, ()>
+where
+    Body: StreamingBody<Data: Send + 'static, Error: Into<BoxError>> + Unpin + Send + 'static,
+{
+    #[cfg(feature = "boring")]
+    pub fn default_with_executor(exec: Executor) -> Self {
         let tls_config =
             rama_tls_boring::client::TlsConnectorDataBuilder::new_http_auto().into_shared_builder();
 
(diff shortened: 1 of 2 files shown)


rama-http-backend/src/client/conn.rs
```diff
@@ -40,13 +41,24 @@ pub struct HttpConnector<S, Body> {
...
+            exec: None,
             _phantom: PhantomData,
         }
     }
 
+    rama_utils::macros::generate_set_and_with! {
+        /// Set the [`Executor`] to be used for spawning child tasks.
+        ///
+        /// By default [`tokio::spawn`] is used if no executor is set.
+        pub fn executor(mut self, exec: Option<Executor>) -> Self {
```
soundofspace: Just a question but instead of a setter, should we always pass this in with constructor, seems like it would otherwise be easy to forget especially on things such as upgrade layer where you might not expect this.

rama-http-backend/src/client/conn.rs
```diff
@@ -91,18 +103,13 @@ where
 
         let io = Box::pin(conn);
 
-        let executor = req
-            .extensions()
-            .get::<Executor>()
-            .cloned()
-            .unwrap_or_default();
+        let exex = self.exec.clone().unwrap_or_default();
```
soundofspace:
```suggestion
        let exec = self.exec.clone().unwrap_or_default();
```

rama-ws/src/handshake/server.rs
```diff
@@ -750,13 +794,8 @@ where
...
 
-                let exec = req
-                    .extensions()
-                    .get::<Executor>()
-                    .cloned()
-                    .unwrap_or_default();
-
-                exec.spawn_task(
+                let exec = self.exec.clone().unwrap_or_default();
+                exec.into_spawn_task(
```
soundofspace, line 800: I also wonder instead of storing `Option<Executor>` should we not just store `Executor`. If we pass option during constructor and always force passing this in then we can do `unwrap_or_default` in constructor and then we dont always have to clone() or default() in places we use it

GlenDC:
I agree with your feedback @soundofspace ,
will fix that and than I'll merge.

rama-http-backend/src/client/conn.rs
GlenDC: Agreed.

rama-ws/src/handshake/server.rs
GlenDC: Agreed.

PR #795, opened 2026-01-12
Fix link to mTLS tunnel example in documentation

ianthetechie:
What it says in the title :) Noticed this while browsing on docs.rs.


GlenDC (approved):
Thank you very much! Greatly appreciated.

PR #797, opened 2026-01-12
switch to cargo nextest for all tests except docs

GlenDC:
docs are currently not tested by cargo nextest,
but everything else we can already switch...

should make it a lot faster to run 'just test' etc... CI times hopefully a bit better as well


GlenDC:
ok should pass now, ready for your feedback @soundofspace

soundofspace (approved):
Nice to see this switch!

PR #803, opened 2026-01-20
drop implicit Arc'ing of http/tcp listener

GlenDC:


examples/http_record_har.rs
```diff
@@ -258,19 +258,21 @@ async fn http_connect_proxy(upgraded: Upgraded) -> Result<(), Infallible> {
...
-        MapResponseBodyLayer::new(Body::new),
-        TraceLayer::new_for_http(),
-        ConsumeErrLayer::default(),
-        UserAgentEmulateLayer::new(state.ua_db.clone())
-            .with_try_auto_detect_user_agent(true)
-            .with_is_optional(true),
-        CompressionLayer::new(),
-        AddRequiredRequestHeadersLayer::new(),
-        EmulateTlsProfileLayer::new(),
+) -> impl Service<Request, Output = Response, Error = Infallible> + Clone {
```
soundofspace: is this arc needed if you already arc at line 148

GlenDC: If it was only used there, no, but the function is also used to create a service directly at line 246 and there it is needed. THat said we can also just arc it there, so I'll do that :) Good catch

PR #808, opened 2026-01-24
Rebuild only if proto files have changed

soundofspace:


examples/grpc/build.rs
```diff
@@ -3,6 +3,8 @@ use std::{env, path::PathBuf};
 fn main() {
     build_examples();
     build_tests();
+
+    println!("cargo:rerun-if-changed=proto");
```
GlenDC: I know I have it already elsewhere, but let's start using the newer 1.77+ syntax


```suggestion
    println!("cargo::rerun-if-changed=proto");
```

examples/grpc/build.rs
GlenDC:
```suggestion
        .for_each(|file| println!("cargo::rerun-if-changed={file}"));
```

rama-grpc/build.rs
```diff
@@ -1,4 +1,7 @@
 fn main() {
     #[cfg(feature = "protobuf")]
-    rama_grpc_build::protobuf::compile_protos("proto/health.proto").unwrap();
+    {
+        rama_grpc_build::protobuf::compile_protos("proto/health.proto").unwrap();
+        println!("cargo:rerun-if-changed=proto");
```
GlenDC:
```suggestion
        println!("cargo::rerun-if-changed=proto");
```

PR #809, opened 2026-01-25
Append only Vec

soundofspace:


rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+            vec.push(NoSize);
+        });
+    }
+
+    #[test]
+    fn drop_of_partial_consumed_into_iter() {
+        create_builder().check(|| {
+            let vec = AppendOnlyVec::<String, 2, 1>::new();
+            vec.push("a".to_owned());
+            // vec.push(4);
```
GlenDC: Why is this commented out?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+        let is_zst = std::mem::size_of::<T>() == 0;
+
+        for (i, atomic_ptr) in self.data.iter_mut().enumerate() {
+            #[cfg(not(all(loom, test)))]
+            let bucket_ptr = *atomic_ptr.get_mut();
+
+            #[cfg(all(test, loom))]
+            let bucket_ptr = atomic_ptr.with_mut(|ptr| *ptr);
+
+            // Before `reserve()` was added we also stopped if remaining == 0`. However
```
GlenDC, line 300: but reserve was removed again it seems?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+///
+/// AMOUNT_OF_BINS is total amount of item bins (=arrays). Each bin has double
+/// the capacity then the one before, so even with a low number here,
+/// we should be able to store a huge amount of items.
+///
+/// BIN_OFFSET calculates the offset of the first bin. Effectively this mean our
+/// first bin will have 2^BIN_OFFSET size.
+pub struct AppendOnlyVec<T, const AMOUNT_OF_BINS: usize = 32, const BIN_OFFSET: u32 = 3> {
+    count: AtomicUsize,
+    reserved: AtomicUsize,
```
GlenDC, line 34: can you add docs to count and reserved, as they are both ambigious and not clear. Would love some doc strings here

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+            reserved: AtomicUsize::new(0),
+            data: std::array::from_fn(|_| AtomicPtr::new(std::ptr::null_mut())),
+        }
+    }
+
+    pub fn push(&self, element: T) -> usize {
+        let idx = self.reserved.fetch_add(1, Ordering::Relaxed);
+        let (bin_idx, offset) = Self::indices(idx);
+
+        // Only allocate a bin if offset = 0. This means there will only ever be one thread
```
GlenDC, line 82: Your comment is very confusing. You say it's non-cooperative, but also that it is? What is it? Perhaps I just fail to read, but it confused me regardless.

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+        Self::INITIAL_BIN_SIZE << idx
+    }
+
+    /// Get item with idx from this vec
+    ///
+    /// # Safety
+    /// This function is safe if idx < self.len()
+    pub unsafe fn get_unchecked(&self, idx: usize) -> &T {
+        let (bin_idx, offset) = Self::indices(idx);
+        let bucket = self.data[bin_idx].load(Ordering::Acquire);
```
GlenDC, line 256: push can reserve an idx that maps to `bin_idx >= AMOUNT_OF_BINS`, which will panic or worse (out of bounds access)

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+    /// ```
+    ///
+    /// ```compile_fail
+    /// use rama_utils::collections::AppendOnlyVec;
+    /// // This should fail because the total size exceeds isize::MAX
+    /// let _ = AppendOnlyVec::<u64, 60, 10>::new();
+    /// ```
+    pub fn new() -> Self {
+        // This has as a side effect that it will check if array layout T is not too big
+        const { Self::assert_layout() }
```
GlenDC, line 65: ok so that does actually work... 

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=974386ab6a0e96b29d4c482059ff3754

call me surprised, did not know this counts as a const context

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+/// vec without having a mutable reference to it.
+///
+///
+/// AMOUNT_OF_BINS is total amount of item bins (=arrays). Each bin has double
+/// the capacity then the one before, so even with a low number here,
+/// we should be able to store a huge amount of items.
+///
+/// BIN_OFFSET calculates the offset of the first bin. Effectively this mean our
+/// first bin will have 2^BIN_OFFSET size.
+pub struct AppendOnlyVec<T, const AMOUNT_OF_BINS: usize = 32, const BIN_OFFSET: u32 = 3> {
```
GlenDC, line 30: maybe it's already verified implicitly but do you test somewhere `AMOUNT_OF_BINS == 0`?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+                #[allow(
+                    clippy::expect_used,
+                    reason = "constructor has checked this on creation"
+                )]
+                let layout = Layout::array::<T>(Self::bin_size(bin_idx))
+                    .expect("layout of array T with size");
+
+                (Some(layout), unsafe { alloc(layout) as *mut T })
+            };
+
```
GlenDC, line 203: maybe I'm missing it, but dont think you are checking if new_ptr is not null...

it can return null in ase of failure according to the docs:

> Returns a pointer to newly-allocated memory, or null to indicate allocation failure.

(https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#tymethod.alloc)

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+            let bucket_ptr = *atomic_ptr.get_mut();
+
+            #[cfg(all(test, loom))]
+            let bucket_ptr = atomic_ptr.with_mut(|ptr| *ptr);
+
+            // Before `reserve()` was added we also stopped if remaining == 0`. However
+            // with reserve it's possible that we already created bins that have no items
+            // in them, so make sure to also clean those up.
+            if bucket_ptr.is_null() {
+                break;
```
GlenDC, line 304: this one seems dangerous and easy to miss if we do ever change other logic that might no longer guarantee the order. no? Might be safer to use `continue`? Dunno maybe I'm overthinking it

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+        }
+        ptr
+    }
+
+    /// Calculate the position in our data structure
+    ///
+    /// Returns (bin_index, offset_in_this_bin)
+    const fn indices(i: usize) -> (usize, usize) {
+        // offset this so we are alligned for ilog2
+        let i = i + Self::INITIAL_BIN_SIZE;
```
GlenDC, line 232: Are you sure that in extreme cases this can not overflow (silent in release)? Not better to either panic, make it error-able or saturate?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+        unsafe { self.get_unchecked(idx) }
+    }
+}
+
+impl<T, const AMOUNT_OF_BINS: usize, const BIN_OFFSET: u32>
+    AppendOnlyVec<T, AMOUNT_OF_BINS, BIN_OFFSET>
+{
+}
+
+/// A double-ended iterator for ExtensionVec
```
GlenDC: Please use links to types in docs, also this is probably still from original code as now it is `AppendOnlyVec`?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    #[cfg(all(test, loom))]
+    {
+        let _ = failures;
+        loom::thread::yield_now();
+    }
+}
+
+unsafe impl<T: Send, const AMOUNT_OF_BINS: usize, const BIN_OFFSET: u32> Send
```
GlenDC, line 365: please add SAFETY comments for each unsafe block and impl

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+
+    fn index(&self, idx: usize) -> &Self::Output {
+        // Bounds check + Acquire ordering to ensure data visibility
+        assert!(idx < self.len(), "Index out of bounds");
+
+        unsafe { self.get_unchecked(idx) }
+    }
+}
+
+impl<T, const AMOUNT_OF_BINS: usize, const BIN_OFFSET: u32>
```
GlenDC: empty impl block, why?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+    }
+
+    pub fn len(&self) -> usize {
+        self.count.load(Ordering::Acquire)
+    }
+
+    /// Returns the maximum number of elements this configuration can hold.
+    /// Total capacity = initial_bin_size * (2^AMOUNT_OF_BINS - 1)
+    pub const fn capacity(&self) -> usize {
+        Self::INITIAL_BIN_SIZE * ((1 << AMOUNT_OF_BINS) - 1)
```
GlenDC, line 162: you might also want to do this in your layout chck or similar const check as for now it is not really verified. I think it will be fine, but it makes me paranoid that this is not checked explicitly

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+
+        Self {
+            count: AtomicUsize::new(0),
+            reserved: AtomicUsize::new(0),
+            data: std::array::from_fn(|_| AtomicPtr::new(std::ptr::null_mut())),
+        }
+    }
+
+    pub fn push(&self, element: T) -> usize {
+        let idx = self.reserved.fetch_add(1, Ordering::Relaxed);
```
GlenDC, line 75: Something tells me you cannot just use relaxed here, if you do think it's solid please explain and once we agree this probably has to be very well documented in-line.

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+        };
+
+        unsafe {
+            bucket_ptr.add(offset).write(element);
+        }
+
+        let mut failures = 0;
+        while self
+            .count
+            .compare_exchange(idx, idx + 1, Ordering::Release, Ordering::Relaxed)
```
GlenDC, line 119: you sure this can be relaxed?

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+
+            // Before `reserve()` was added we also stopped if remaining == 0`. However
+            // with reserve it's possible that we already created bins that have no items
+            // in them, so make sure to also clean those up.
+            if bucket_ptr.is_null() {
+                break;
+            }
+
+            let bin_cap = Self::bin_size(i);
+            let to_drop = std::cmp::min(remaining, bin_cap);
```
GlenDC, line 308: add a big fat warning comment here as you will deallocate unitialised methods if you ever refactor the code to support holes (e.g. in case of your reserve logic).

rama-utils/src/collections/append_only_vec.rs
```diff
@@ -0,0 +1,662 @@
...
+
+        // The pros and cons of either approach are:
+        // Not cooperative: single allocation, but other threads need to use spin_wait until this allocation is finished
+        // Cooperative: potential of many allocations for the same bin (dropped after, so only short spike), who-ever is fastest wins
+
+        // One first sight cooperative seems nicer, but also note that our idx and length logic is sequential, so even
+        // if we make the allocation cooperative and not blocking, we will mostly be moving the spin_wait to that step.
+        // For our use case we also don't expect many (if any) concurrent/parallel pushes to this vec.
+
+        let bucket_ptr = if offset == 0 {
```
GlenDC, line 97: I think to be safe you might always want to create bin if needed, dunno how much more expensive that might make it but I have a feeling that if there's a thread that aborts for w/e reason while holding that offset = 0 lock and before allocating you will be waiting forever in your else branch due to the nil ptrs that will never fill???

not sure how easy will be to test this, but pretty certain that is the case here

rama-utils/src/collections/append_only_vec.rs
soundofspace: yes but I'm keeping this logic here in case we do want to add it in the future

rama-utils/src/collections/append_only_vec.rs
soundofspace: mistake

rama-utils/src/collections/append_only_vec.rs
soundofspace: This will be caught be loom if we do ever add it, it's also why i left the reserve test around to easily enable it again

rama-utils/src/collections/append_only_vec.rs
soundofspace: True, but the other atomic count logic will also hang it that case. The comment already tries to mention that, but I guess it's a lot when reading it the first time

rama-utils/src/collections/append_only_vec.rs
soundofspace: There are other places also that would need the warning. It's why I didnt add it (and maybe never will). But dont think it is needed to explain it everywhere (will probably miss places), but that if we do implement it that it will be a tricky one

rama-utils/src/collections/append_only_vec.rs
soundofspace: This part can easily be switched to co-op but other parts not so much without decent rework, this is mostly here for me to remember that (or you)

rama-utils/src/collections/append_only_vec.rs
soundofspace: Added extra check to push

rama-utils/src/collections/append_only_vec.rs
soundofspace: leftover from refactor

rama-utils/src/collections/append_only_vec.rs
soundofspace: What do you mean with it's not verified?

rama-utils/src/collections/append_only_vec.rs
soundofspace: Why, we don't need to synchronize with other variables (where you need ordering), the sync logic is part of the atomicPtrs and count

rama-utils/src/collections/append_only_vec.rs
soundofspace: We dont care about the failure case and nothing depends on it, so can be relaxed

rama-utils/src/collections/append_only_vec.rs
soundofspace: Yea surprised me also. It's also why I added doctests to actually confirm this (found doctests the easiest option to test something small for this)

rama-utils/src/collections/append_only_vec.rs
soundofspace: Wow missed this alloc doesnt mention this but global alloc method does

rama-utils/src/collections/append_only_vec.rs
GlenDC: Nvm.

PR #814, opened 2026-01-30
feat: add `ends_with_ignore_ascii_case`

### LINKED ISSUE #801: add ends_with_ignore_ascii_case str (search) utility
It can probably be something like this:

```rust
fn ends_with_ignore_ascii_case(path: &str, suffix: &str) -> bool {
    if path.len() < suffix.len() {
        return false;
    }

    let start = path.len() - suffix.len();
    path.get(start..)
        .is_some_and(|tail| tail.eq_ignore_ascii_case(suffix))
}
```

We already have `starts_with_ignore_ascii_case`, from which that was ported.
Put it in the same module as that function

Also add sufficient test cases similar to the `starts_with` version.

azzamsa:
fixes #801 

I intentionally made the test fail during the debugging process.
The function snippet passes all test cases.

### DIFF
--- a/rama-utils/src/str/search.rs
+++ b/rama-utils/src/str/search.rs
@@ -83,13 +83,19 @@ mod tests {
     #[test]
     fn test_starts_with_ignore_ascii_case() {
         assert!(starts_with_ignore_ascii_case("user-agent", "user"));
+        assert!(starts_with_ignore_ascii_case("User-Agent", "user"));
+        assert!(starts_with_ignore_ascii_case("USER-AGENT", "user"));
         assert!(!starts_with_ignore_ascii_case("user-agent", "agent"));
+        assert!(!starts_with_ignore_ascii_case("User-Agent", "agent"));
     }
 
     #[test]
     fn test_ends_with_ignore_ascii_case() {
         assert!(ends_with_ignore_ascii_case("user-agent", "agent"));
+        assert!(ends_with_ignore_ascii_case("User-Agent", "agent"));
+        assert!(ends_with_ignore_ascii_case("USER-AGENT", "agent"));
         assert!(!ends_with_ignore_ascii_case("user-agent", "user"));
+        assert!(!ends_with_ignore_ascii_case("User-Agent", "user"));
     }
 
     #[test]


GlenDC:
Slightly out of scope @azzamsa but can you please to both the tests of the `start` and `end` variants also add a testcase that ensures we also match regardless of the casing of the word? As right now (also partly my fault) we only test purely lowercase, whereas something like `User-Agent` and `USER_AGENT` should both match that it ends with `agent` (same goes for the _start_ variant). Once you add that we can probably merge, thx!

rama-utils/src/str/search.rs
```diff
@@ -75,6 +86,12 @@ mod tests {
         assert!(!starts_with_ignore_ascii_case("user-agent", "agent"));
     }
 
+    #[test]
+    fn test_ends_with_ignore_ascii_case() {
+        assert!(ends_with_ignore_ascii_case("user-agent", "agent"));
+        assert!(!ends_with_ignore_ascii_case("user-agent", "user"));
+    }
```
GlenDC, line 99:
```suggestion
        assert!(!ends_with_ignore_ascii_case("User-Agent", "user"));
    }
```

rama-utils/src/str/search.rs
```diff
@@ -75,6 +86,12 @@ mod tests {
         assert!(!starts_with_ignore_ascii_case("user-agent", "agent"));
     }
 
+    #[test]
+    fn test_ends_with_ignore_ascii_case() {
+        assert!(ends_with_ignore_ascii_case("user-agent", "agent"));
```
GlenDC, line 94:
```suggestion
        assert!(ends_with_ignore_ascii_case("user-agent", "agent"));
        assert!(ends_with_ignore_ascii_case("User-Agent", "agent"));
        assert!(ends_with_ignore_ascii_case("USER-AGENT", "agent"));
```

rama-utils/src/str/search.rs
```diff
@@ -75,6 +86,12 @@ mod tests {
         assert!(!starts_with_ignore_ascii_case("user-agent", "agent"));
```
GlenDC, line 88:
```suggestion
        assert!(!starts_with_ignore_ascii_case("user-agent", "agent"));
        assert!(!starts_with_ignore_ascii_case("User-Agent", "agent"));
```

rama-utils/src/str/search.rs
GlenDC, line 85:
```suggestion
        assert!(starts_with_ignore_ascii_case("user-agent", "user"));
        assert!(starts_with_ignore_ascii_case("User-Agent", "user"));
        assert!(starts_with_ignore_ascii_case("USER-AGENT", "user"));
```

azzamsa:
> Slightly out of scope @azzamsa but can you please to both the tests of the `start` and `end` variants also add a testcase that ensures we also match regardless of the casing of the word? As right now (also partly my fault) we only test purely lowercase, whereas something like `User-Agent` and `USER_AGENT` should both match that it ends with `agent` (same goes for the _start_ variant). Once you add that we can probably merge, thx!

You did already, Thanks!

PR #815, opened 2026-02-03
feat: add OTLP gRPC exporter

seuros:
Added Opentelemetry to the gRPC exporter , this mirror the HTTP example.

### DIFF
--- a/rama-grpc/Cargo.toml
+++ b/rama-grpc/Cargo.toml
@@ -33,6 +33,7 @@ pin-project-lite = { workspace = true }
 prost = { workspace = true, optional = true }
 prost-types = { workspace = true, optional = true }
 radix_trie = { workspace = true }
+parking_lot = { workspace = true }
 rama-core = { workspace = true }
 rama-grpc-build = { workspace = true }
 rama-http = { workspace = true }
--- a/rama-grpc/src/service/opentelemetry/transform.rs
+++ b/rama-grpc/src/service/opentelemetry/transform.rs
@@ -10,9 +10,8 @@ use super::proto::{
     ProtoMetric, ProtoResource, ProtoResourceMetrics, ProtoScopeMetrics, ProtoSum, ResourceSpans,
     ScopeSpans, Span, Status,
 };
-use rama_core::telemetry::opentelemetry;
 use rama_core::telemetry::opentelemetry::{
-    Array, Value,
+    self, Array, Value,
     sdk::{
         Resource,
         metrics::{
@@ -29,6 +28,8 @@ use rama_core::telemetry::opentelemetry::{
     },
     trace as otrace,
 };
+use rama_core::telemetry::tracing;
+
 use std::{
     collections::HashMap,
     time::{Duration, SystemTime, UNIX_EPOCH},
@@ -50,32 +51,40 @@ fn to_nanos(time: SystemTime) -> u64 {
 // Common: attributes + values
 // ──────────────────────────────────────────────
 
-fn key_value_from(kv: opentelemetry::KeyValue) -> KeyValue {
-    KeyValue {
-        key: kv.key.as_str().to_owned(),
-        value: Some(value_into_any(kv.value)),
+fn key_value_from(kv: opentelemetry::KeyValue) -> Option<KeyValue> {
+    let key = kv.key.as_str().to_owned();
+    let value = value_into_any(kv.value).or_else(|| {
+        tracing::debug!("dropping unsupported otel attribute key={key}");
+        None
+    })?;
+
+    Some(KeyValue {
+        key,
+        value: Some(value),
         ..Default::default()
-    }
+    })
 }
 
-fn key_value_from_ref(key: &opentelemetry::Key, value: &Value) -> KeyValue {
-    KeyValue {
-        key: key.as_str().to_owned(),
-        value: Some(value_into_any(value.clone())),
+fn key_value_from_ref(key: &opentelemetry::Key, value: &Value) -> Option<KeyValue> {
+    let key = key.as_str().to_owned();
+    let value = value_into_any(value.clone()).or_else(|| {
+        tracing::debug!("dropping unsupported otel attribute key={key}");
+        None
+    })?;
+
+    Some(KeyValue {
+        key,
+        value: Some(value),
         ..Default::default()
-    }
+    })
 }
 
-fn key_value_from_kv_ref(kv: &opentelemetry::KeyValue) -> KeyValue {
-    KeyValue {
-        key: kv.key.as_str().to_owned(),
-        value: Some(value_into_any(kv.value.clone())),
-        ..Default::default()
-    }
+fn key_value_from_kv_ref(kv: &opentelemetry::KeyValue) -> Option<KeyValue> {
+    key_value_from_ref(&kv.key, &kv.value)
 }
 
-fn value_into_any(value: Value) -> AnyValue {
-    AnyValue {
+fn value_into_any(value: Value) -> Option<AnyValue> {
+    Some(AnyValue {
         value: match value {
             Value::Bool(val) => Some(proto::any_value::Value::BoolValue(val)),
             Value::I64(val) => Some(proto::any_value::Value::IntValue(val)),
@@ -86,11 +95,17 @@ fn value_into_any(value: Value) -> AnyValue {
                 Array::I64(vals) => array_into_proto(vals),
                 Array::F64(vals) => array_into_proto(vals),
                 Array::String(vals) => array_into_proto(vals),
-                _ => unreachable!("nonexistent array type"),
+                other => {
+                    tracing::debug!("unkown array value ({other:?}): please report bug in rama");
+                    return None;
+                }
             })),
-            _ => unreachable!("nonexistent value type"),
+            other => {
+                tracing::debug!("unkown value value ({other:?}): please report bug in rama");
+                return None;
+            }
         },
-    }
+    })
 }
 
 fn array_into_proto<T>(vals: Vec<T>) -> ArrayValue
@@ -99,19 +114,19 @@ where
 {
     let values = vals
         .into_iter()
-        .map(|val| value_into_any(Value::from(val)))
+        .filter_map(|val| value_into_any(Value::from(val)))
         .collect();
     ArrayValue { values }
 }
 
 fn attributes_from_iter<I: IntoIterator<Item = opentelemetry::KeyValue>>(kvs: I) -> Vec<KeyValue> {
-    kvs.into_iter().map(key_value_from).collect()
+    kvs.into_iter().filter_map(key_value_from).collect()
 }
 
 fn resource_attributes(resource: &Resource) -> Vec<KeyValue> {
     resource
         .iter()
-        .map(|(k, v)| key_value_from(opentelemetry::KeyValue::new(k.clone(), v.clone())))
+        .filter_map(|(k, v)| key_value_from(opentelemetry::KeyValue::new(k.clone(), v.clone())))
         .collect()
 }
 
@@ -308,7 +323,7 @@ fn sdk_resource_into(resource: &Resource) -> ProtoResource {
     ProtoResource {
         attributes: resource
             .iter()
-            .map(|(k, v)| key_value_from_ref(k, v))
+            .filter_map(|(k, v)| key_value_from_ref(k, v))
             .collect(),
         dropped_attributes_count: 0,
         ..Default::default()
@@ -352,10 +367,10 @@ impl Numeric for u64 {
         self as f64
     }
     fn into_exemplar_value(self) -> proto::exemplar::Value {
-        proto::exemplar::Value::AsInt(i64::try_from(self).unwrap_or_default())
+        proto::exemplar::Value::AsInt(saturating_u64_to_i64(self))
     }
     fn into_data_point_value(self) -> proto::number_data_point::Value {
-        proto::number_data_point::Value::AsInt(i64::try_from(self).unwrap_or_default())
+        proto::number_data_point::Value::AsInt(saturating_u64_to_i64(self))
     }
 }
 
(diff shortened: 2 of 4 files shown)


examples/README.md
```diff
@@ -35,6 +35,7 @@ You can find these integration tests at [../tests/integration/examples](../tests
 
 ### gRPC
 
+- [`grpc_example.rs`](./grpc_example.rs) - Telemetry and monitoring via OTLP gRPC exporter
```
GlenDC: can you move the mention, + doc + example to  `examples/grpc`? E.g. the example file can probably live in `examples/grpc/opentelemetry/exporter.rs` or so.

Cargo.toml
```diff
@@ -617,6 +619,10 @@ required-features = ["http-full"]
 name = "http_telemetry"
 required-features = ["http-full", "opentelemetry"]
 
+[[example]]
+name = "grpc_example"
```
GlenDC: no need to have that here, grpc examples are in a separate project

Cargo.toml
```diff
@@ -149,6 +149,7 @@ opentelemetry-otlp = { version = "0.31", default-features = false, features = [
     "metrics",
     "trace",
 ] }
+opentelemetry-proto = { version = "0.31", default-features = false }
```
GlenDC: Don't think we want to actually import  or use `opentelemetry-proto` crate (https://github.com/open-telemetry/opentelemetry-rust/tree/main/opentelemetry-proto). There is no real value for us AFAIK. It is mostly about tonic. You can for sure fork code from it if it helps you with the implementation, but we do not want to use the crate.

Instead we should:

- include the proto files directly in our repo (in the `rama-grpc` crate and build them (behind the feature flag) using our regular code
- make use of that to support opentelemetry in `rama-grpc`.

rama-grpc/Cargo.toml
```diff
@@ -23,10 +23,24 @@ full = ["protobuf", "transport", "compression"]
...
+    "dep:opentelemetry",
+    "dep:opentelemetry-proto",
+    "dep:opentelemetry_sdk",
+]
+internal-logs = ["opentelemetry", "opentelemetry/internal-logs"]
 
 [dependencies]
 base64 = { workspace = true }
 flate2 = { workspace = true, optional = true }
+opentelemetry = { workspace = true, optional = true }
```
GlenDC: there should be no need for you to import opentelemetry or opentelemetry-sdk directly. Both are afaik already part of rama-core and can be used via there. 

opentelemetry-proto should be dropped (see other comment).

rama-grpc/build.rs
```diff
@@ -4,4 +4,11 @@ fn main() {
         rama_grpc_build::protobuf::compile_protos("proto/health.proto").unwrap();
         println!("cargo::rerun-if-changed=proto");
     }
+
+    #[cfg(feature = "opentelemetry")]
```
GlenDC, line 9: this can be replaced with compiling grpc protos

GlenDC: There should be no need to watch these env variables btw we already watch proto folder, should be enough for this, this is just about building the code from proto so no need to check env variables

examples/grpc_example.rs
```diff
@@ -0,0 +1,170 @@
...
+//! It also sets up [`tracing`] in a basic manner.
+//!
+//! Learn more about telemetry at <https://ramaproxy.org/book/intro/telemetry.html>.
+//! In this book chapter you'll also find more information on how you can
+//! consume the metrics of this example in tools such as Prometheus and Grafana.
+//!
+//! [`opentelemetry`]: https://opentelemetry.io/
+//! [`tracing`]: https://tracing.rs/
+//!
+//! This example will create a server that listens on `127.0.0.1:62012.
```
GlenDC: Make sure to use a unique port for each example. `62012` is already used by the http otlp example

examples/grpc/src/otel_exporter.rs
```diff
@@ -0,0 +1,170 @@
...
+                ])
+                .build(),
+        );
+
+        let counter = meter.i64_up_down_counter("visitor_counter").build();
+        Self { counter }
+    }
+}
+
+#[tokio::main]
```
GlenDC, line 99: example does look good, nice

rama-grpc/src/service/opentelemetry.rs
```diff
@@ -0,0 +1,652 @@
+use crate::{
+    Request,
+    client::{Grpc, GrpcService},
+    codec::CompressionEncoding,
+    metadata::MetadataMap,
+    protobuf::ProstCodec,
+};
+use opentelemetry::{otel_debug, otel_warn};
+use opentelemetry_proto::{
```
GlenDC: this will have to be dropped, completely, and instead you'll need to do it fully with rama-grpc, not with tonic

rama-grpc/src/service/opentelemetry.rs
```diff
@@ -0,0 +1,652 @@
...
+    }
+
+    /// Override the metrics-specific compression.
+    pub fn with_metrics_compression(mut self, compression: CompressionEncoding) -> Self {
+        self.metrics.compression = Some(compression);
+        self
+    }
+
+    /// Override the metrics-specific metadata.
+    pub fn with_metrics_metadata(mut self, metadata: MetadataMap) -> Self {
```
GlenDC: use `rama::utils::macros::generate_set_and_with` for all of these setters please, it will ensure to generate the `with_` and `set_` variants automatically for you. e.g.

```rust
rama::utils::macros::generate_set_and_with! {
    /// Override the metrics-specific metadata.
   pub fn metrics_metadata(mut self, metadata: MetadataMap) -> Self {
    self.metrics.metadata = metadata;
        self
    }
}
```

GlenDC:
Let me know @seuros if you need any guidance or help with this one.
Will be good to have it as it will mean we can remove tonic and tonic-prost from the dep tree:

```
% cargo tree --all-features -i tonic
tonic v0.14.4
├── opentelemetry-proto v0.31.0
│   └── opentelemetry-otlp v0.31.0
│       └── rama v0.3.0-rc1 (/Users/glendc/code/github.com/plabayo/rama)
└── tonic-prost v0.14.4
    └── opentelemetry-proto v0.31.0 (*)
```

docs/book/src/intro/telemetry.md
```diff
@@ -30,12 +30,14 @@ Rama does not provide specific exporters or consumers of these metrics.
 Instead — similar to tracing — you'll have to add your own dependency for it.
 
 We recommend you to use <https://crates.io/crates/opentelemetry-otlp> as it
-plays nicely with the openetelemetry dependencies used by Rama as well as
+plays nicely with the OpenTelemetry dependencies used by Rama as well as
 popular ingestion tools such as [Prometheus](https://prometheus.io/).
 
 ### Rama Telemetry Example
 
 > Source Code: [/examples/http_telemetry.rs](https://github.com/plabayo/rama/tree/main/examples/http_telemetry.rs)
```
GlenDC: I would reformat this a bit by writing something like...

Rama supports OTEL exporters out of the box using the rama network stack, the source code for which an be found below:

- HTTP exporter: ...
- gRPC exporter ...

Something like that... but a bit improved ofc..

rama-grpc/Cargo.toml
```diff
@@ -43,6 +45,8 @@ tokio = { workspace = true, features = ["macros", "net"] }
 zstd = { workspace = true, optional = true }
 
 [build-dependencies]
+prost-build = { workspace = true }
```
GlenDC: Why are you depending on these here??? This is already part of rama-grpc (via rama-grpc-build). What am I missing here?

rama-grpc/build.rs
```diff
@@ -4,4 +4,35 @@ fn main() {
...
+    {
+        // Proto files vendored from https://github.com/open-telemetry/opentelemetry-proto
+        // tag: v1.5.0, commit: 2bd940b2b77c1ab57c27166af21384906da7bb2b
+        let mut config = prost_build::Config::new();
+        config.disable_comments(["."]);
```
GlenDC: why is this needed? And even if it is, please doc why as it is not clear to me. Right now I am not convinced we need this, but you tell me

rama-grpc/build.rs
```diff
@@ -4,4 +4,35 @@ fn main() {
...
+
+        // Use vendored protoc to avoid requiring system protoc installation
+        if let Ok(path) = protoc_bin_vendored::protoc_bin_path() {
```
GlenDC: AFAIK you can also do this with rama via rama-grpc > rama-grpc-build... as we already have support for protoc-bin-vendored there, so I doubt we need that as a direct dep here

rama-grpc/build.rs
```diff
@@ -4,4 +4,35 @@ fn main() {
...
+            config.protoc_executable(path);
+        }
+
+        println!("cargo::rerun-if-changed=build.rs");
```
GlenDC: It's a good one, but doesn't belong in this branch, can be moved to the top of the main fn here

rama-grpc/build.rs
```diff
@@ -4,4 +4,35 @@ fn main() {
...
+        println!(
```
GlenDC: are these 2 explicit rerun lines really need? if this code runs it means we have the opentelemetry feature enabled, meaning we also have implicitly the protobuf feature enabled... which reruns if the `proto/` root dir changes... Does that not also include all children, regardless how deep??

rama-grpc/src/service/opentelemetry/proto.rs
```diff
@@ -0,0 +1,90 @@
...
+#[allow(
+    dead_code,
+    unreachable_pub,
+    missing_docs,
+    clippy::default_trait_access,
+    clippy::derive_partial_eq_without_eq,
+    clippy::enum_variant_names,
+    clippy::trivially_copy_pass_by_ref
+)]
+pub mod opentelemetry {
```
GlenDC: why do you have thi snested in opentelemetry and proto? We are already in the opentelemetry::proto module... seems a bit to omuch? Is there agood reason not to have modules such as `common` directly in the proto.rs file instead of nested under opentelemetry { proto { ???

rama-grpc/proto/opentelemetry/proto/collector/metrics/v1/metrics_service.proto
```diff
@@ -0,0 +1,79 @@
...
+option java_multiple_files = true;
+option java_package = "io.opentelemetry.proto.collector.metrics.v1";
+option java_outer_classname = "MetricsServiceProto";
+option go_package = "go.opentelemetry.io/proto/otlp/collector/metrics/v1";
+
+// Service that can be used to push metrics between one Application
+// instrumented with OpenTelemetry and a collector, or between a collector and a
+// central collector.
+service MetricsService {
+  // For performance reasons, it is recommended to keep this RPC
```
GlenDC: Did you add this comment? Original file does not have it: https://github.com/open-telemetry/opentelemetry-proto/blob/3ca54b660a5b307a618c59840b88532b18673869/opentelemetry/proto/collector/metrics/v1/metrics_service.proto#L6

rama-grpc/proto/opentelemetry/proto/collector/trace/v1/trace_service.proto
```diff
@@ -0,0 +1,79 @@
...
+option java_multiple_files = true;
+option java_package = "io.opentelemetry.proto.collector.trace.v1";
+option java_outer_classname = "TraceServiceProto";
+option go_package = "go.opentelemetry.io/proto/otlp/collector/trace/v1";
+
+// Service that can be used to push spans between one Application instrumented with
+// OpenTelemetry and a collector, or between a collector and a central collector (in this
+// case spans are sent/received to/from multiple Applications).
+service TraceService {
+  // For performance reasons, it is recommended to keep this RPC
```
GlenDC: Same here... I do not find that comment in the original file?

rama-grpc/proto/opentelemetry/proto/resource/v1/resource.proto
```diff
@@ -0,0 +1,37 @@
...
+option java_outer_classname = "ResourceProto";
+option go_package = "go.opentelemetry.io/proto/otlp/resource/v1";
+
+// Resource information.
+message Resource {
+  // Set of attributes that describe the resource.
+  // Attribute keys MUST be unique (it is not allowed to have more than one
+  // attribute with the same key).
+  repeated opentelemetry.proto.common.v1.KeyValue attributes = 1;
+
```
GlenDC, line 34: also differences with https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/resource/v1/resource.proto it seems?

rama-grpc/proto/opentelemetry/proto/trace/v1/trace.proto
```diff
@@ -0,0 +1,357 @@
+// Copyright 2019, OpenTelemetry Authors
```
GlenDC, line 1: a lot of differences (in comments) compared to https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto, is there a good reason for it? Otherwise better we have their version, no?

rama-grpc/src/service/opentelemetry/transform.rs
```diff
@@ -0,0 +1,496 @@
...
+};
+
+// ──────────────────────────────────────────────
+// Time utilities
+// ──────────────────────────────────────────────
+
+fn to_nanos(time: SystemTime) -> u64 {
+    time.duration_since(UNIX_EPOCH)
+        .unwrap_or_else(|_| Duration::from_secs(0))
+        .as_nanos() as u64
```
GlenDC: use the safe version instead of `as` please and choose something like a saturation approach or w/e

examples/grpc/src/otel_exporter.rs
```diff
@@ -0,0 +1,170 @@
...
+//! ```sh
+//! cargo run -p rama-grpc-examples --bin otel-exporter
+//! ```
+//!
+//! # Expected output
+//!
+//! The server will start and listen on `:62013`. You can use `curl`:
+//!
+//! ```sh
+//! curl -v http://127.0.0.1:62013
```
GlenDC: `62013` is already chosen as a hardcoded port for another example, can you pick one that is not yet used by any example please... (sorry)

PR #818, opened 2026-02-06
Issue 374/Add TLS enum to benchmarks

M-Kusumgar:
I had a couple of thoughts/questions about this one. Also I basically had this PR done earlier this evening however got stuck on trying to debug something that I think is an issue, so some help would be much appreciated!

Thoughts/bugs/questions:
1. I had to box the http service so that I don't duplicate the code for different tls values. Is that okay? or is there a better way of doing this
2. The divan args list is growing and it'll triple when we add the proxy enum, is there anything we can do about this? i cant find an easy way to pass in things into divan benches other than the args list, if we could do it programatically then we can create arrays of possible values of enums and do every combination. I also don't know if this is much of a problem or worth the extra complexity and time
3. I believe I've found a potential bug (but best to verify on your machine just in case), the http2 route with tls none takes ~40ms on my computer, this seem very abnormal compared to http1 tls none which is about 0.3-1.5ms - this is the case in main as well, not just on my branch. Is there a problem with the http2? or have we done the setup wrong? I noticed that the numbers became more reasonable on main after replacing the client post url from http to https however, even more confusingly changing the client post url no longer has any effect on the benchmark times in my branch here, its consistently at ~40ms
4. Also rustls tls is faster than no tls? Is this normal too?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,6 +15,10 @@ on:
       - main
   pull_request: {}
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   precheck-rust:
     strategy:


GlenDC (approved):

GlenDC:
@M-Kusumgar 

1. boxing is fine
2. agreed, found a way to fix it
3. and 4. my results are different, check again with latest code

GlenDC:
Next up will be adding proxy support (None/HTTP/SOCKS5). This will require adding another enum, and also another server (the proxy one).

M-Kusumgar:
never knew you could have a const fn, thats neat! Ill start on the proxy support but might take me a little bit of time! should have it done this weekend, quick question: why did you decide to take out the rng stuff, surely the compression will be trivial with that?

edit: just ran it up and the numbers look more reasonable after pulling your code in

GlenDC:
> quick question: why did you decide to take out the rng stuff, surely the compression will be trivial with that?

That was a mistake @M-Kusumgar . Had removed it tmp to not distract me, and had forgotten to put it back. Thank you for catching it.

PR #822, opened 2026-02-11
fix HAR for spec compliance

fatpusheen:
minor fixes as per [the spec](https://github.com/plabayo/rama/blob/main/rama-http/src/layer/har/spec.md)

### DIFF
--- a/rama-http/src/layer/har/spec.rs
+++ b/rama-http/src/layer/har/spec.rs
@@ -22,7 +22,6 @@ use rama_http_types::mime::Mime;
 use rama_http_types::proto::h1::Http1HeaderName;
 use rama_http_types::proto::h1::ext::ReasonPhrase;
 use rama_http_types::{HeaderMap, Version as RamaHttpVersion, proto::h1::Http1HeaderMap};
-use rama_net::address::SocketAddress;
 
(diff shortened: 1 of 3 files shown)


rama-http/src/layer/har/spec.rs
```diff
@@ -302,7 +302,7 @@ pub struct Entry {
     /// IP address of the server that was connected
     ///
     /// (result of DNS resolution).
-    #[serde(rename = "serverAddress")]
+    #[serde(rename = "serverIPAddress")]
     pub server_address: Option<SocketAddress>, // TODO: be able to provide for client middleware
```
GlenDC:
```suggestion
    pub server_ip_address: Option<IpAddr>, // TODO: be able to provide for client middleware
```

please rename also the property to reflect the correct name and change the type to IpAddr, as this is not a SocketAddress. The replay example test fails now because of this.

rama-http/src/layer/har/spec.rs
GlenDC, line 5:
```suggestion
use std::net::IpAddr;
use std::str::FromStr;
```

GlenDC (changes requested):

PR #834, opened 2026-02-21
add ALPS extension with support for new_codepoint

ntrippar:
- change `ClientHelloExtension::ApplicationSettings` to include `protocols` and `new_codepoint`
- parse ALPS using both extension IDs and preserve which codepoint was received
- propagate ALPS settings into `TlsConnectorDataBuilder` and configure boring TLS accordingly
- update CLI/echo extension rendering to match the new `ApplicationSettings` shape

This requires the changes in https://github.com/plabayo/rama-boring/pull/4

with this we can have exactly the same ja4 and ja3


ntrippar:
We should add a test cases for `parse_tls_extension_application_settings_content`

ntrippar:
There is an error that I need to fix when deserialize embedded profiles as now its in another format, we should update the file with the new format

ntrippar:
I updated the chrome profiles to set the new_codepoint and the other ones I left with the old. with this script

```python
import json

def parse_profile(profile_path):
    with open(profile_path, 'r') as f:
        profile_data = json.load(f)
        for profile in profile_data:
            uastr = profile.get('uastr', '')
            
            if profile["tls_client_hello"]:
                if profile["tls_client_hello"]["extensions"]:
                    for i in range(0, len(profile["tls_client_hello"]["extensions"])):
                        if profile['tls_client_hello']['extensions'][i].get("ApplicationSettings", None):
                            protocols = profile['tls_client_hello']['extensions'][i]["ApplicationSettings"]
                            if "Version/" in uastr and "Safari/" in uastr:
                                app_settings = {"protocols": protocols, "new_codepoint": False}
                            elif "Firefox/" in uastr:
                                app_settings = {"protocols": protocols, "new_codepoint": False}
                            else:
                                app_settings = {"protocols": protocols, "new_codepoint": True}
                            profile['tls_client_hello']['extensions'][i]["ApplicationSettings"] = app_settings

            
    return profile_data

print(json.dumps(parse_profile("embed_profiles_copy.json")))
```

GlenDC:
> We should add a test cases for `parse_tls_extension_application_settings_content`

Added a first basic test

GlenDC:
Will squash+merge this PR if CI passes now. AFAIK it's complete:

- thank you for the initial contribution @ntrippar 
- have updated and while I was at it improved rama-boring
- ensured it passes on all platforms
- added the missing parsing test in rama-net

PR #838, opened 2026-02-28
Run events in span scope

soundofspace:


rama-http/src/layer/trace/make_span.rs
```diff
@@ -105,6 +114,8 @@ impl<B> MakeSpan<B> for DefaultMakeSpan {
         let found_domain_cow_str = found_domain.as_ref().map(|d| d.to_str());
         let found_domain_str = found_domain_cow_str.as_deref();
 
+        let otel_name = self.otel_name.as_deref().unwrap_or("request");
```
GlenDC: can do unwrap_or_else into a static arcstr

rama-http/src/layer/trace/make_span.rs
```diff
@@ -39,6 +39,7 @@ where
 pub struct DefaultMakeSpan {
     level: Level,
     include_headers: bool,
+    otel_name: Option<String>,
```
GlenDC:
```suggestion
    otel_name: Option<ArcStr>,
```

whenever you have a read-only string better to use ArcStr.. That one also supports static for static cases. 

Or for stack-appropriate ones <21 bytes you can use SmolStr. For this one however there is not much advantage in keeping it stack os ArcStr I would use here

PR #839, opened 2026-02-28
New extensions behind a cfg flag

soundofspace:
Use new extensions when extensions_new cfg is enabled. This still uses the old API but is meant so we (and maybe other people) can try new extension storage logic.

After some time has passed and no issues popped up the goal is to migrate completely to the new stack and also start using the new API and features (&self, multiple copies...)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -267,7 +267,7 @@ zstd = "0.13"
 [workspace.lints.rust]
 unreachable_pub = "deny"
 unused_imports = "warn"
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(fuzzing)', 'cfg(loom)'] }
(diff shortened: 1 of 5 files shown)


soundofspace:
@GlenDC by having it behind a cfg flag we can start using this internally already to confirm all the internals work as expected. Do you agree with this approach? Also should we enable this for some tests in CI, if so how without duplicating all of them? 

Once we figured out details I will finish this PR, and it can be review. No point in reviewing it before that, except for to get high level idea of what I'm doing here

GlenDC:
Closing this PR as we do not want to go for cfg flag for this. We do want the new extensions but we should just bite the bullet and go for it at once. I think we are aligned on this, but in case I miss something, do sync with me offline.

PR #842, opened 2026-03-10
Expose BytesRWTracker

soundofspace:


GlenDC:
Should probably move to `rama-core/io` in future, not really bound to networking at all, just bound to IO (asyncRead/AsyncWrite). Anyway good enough for now

soundofspace:
> Should probably move to `rama-core/io` in future, not really bound to networking at all, just bound to IO (asyncRead/AsyncWrite). Anyway good enough for now

Agree, ping me once your io PR is merged, and am happy to move this.

PR #844, opened 2026-03-13
fix(router): return 405 with Allow header on method mismatch

### LINKED ISSUE #843: http web router - returns 404 instead of 405 on method mismatch
In the http web router, when a request path matches a registered route but the HTTP method does not, the router falls through to a 404 response instead of returning 405 Method Not Allowed.

Per RFC 7231 §6.5.5 (https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.5), a 405 response must include an Allow header listing the methods supported for that resource.

Example: A router with GET /users/{id} and DELETE /users/{id} registered — a PUT /users/123 request incorrectly returns 404 instead of 405 Method Not Allowed with Allow: DELETE, GET.

ikorason:
fix #843 

Per RFC 7231, when a path matches but no registered method matches, return 405 Method Not Allowed with an Allow header listing the valid methods instead of falling through to 404. Sub-services still take priority over the 405 response.

- add `MethodMatcher::iter_str()` to enumerate method name strings
- add `HttpMatcher::allowed_methods()` (pub crate) to extract the method constraint from a matcher
- defer 405 decision until after sub_services check in `serve()`


rama-http/src/matcher/method.rs
```diff
@@ -47,6 +47,26 @@ impl MethodMatcher {
     pub const fn or(self, other: Self) -> Self {
         Self(self.0 | other.0)
     }
+
+    /// Returns an iterator over the HTTP method name strings represented by this matcher.
+    ///
+    /// The iterator yields names in alphabetical order (CONNECT, DELETE, GET, HEAD, OPTIONS,
+    /// PATCH, POST, PUT, TRACE) matching only the bits that are set.
+    pub fn iter_str(self) -> impl Iterator<Item = &'static str> {
```
ikorason: @GlenDC I wonder if this is the correct approach here

rama-http/src/matcher/mod.rs
```diff
@@ -874,6 +874,30 @@ impl<Body> HttpMatcher<Body> {
...
+}
+
+impl<Body> HttpMatcherKind<Body> {
+    fn allowed_methods(&self) -> Option<MethodMatcher> {
+        match self {
+            Self::Method(m) => Some(*m),
+            // All other variants (All, Any, Path, Domain, Version, Uri, Header, Socket,
+            // SubdomainTrie, Custom) either have no method constraint or cannot be
+            // enumerated — callers treat None as "no finite Allow list available".
+            _ => None,
```
GlenDC, line 915: While a bit exotic pretty certain you do want to deal better regarding All / Any. Those do have potential use here, no?

rama-http/src/matcher/method.rs
GlenDC: I would recommend returning the actual `Method` variants instead :) Rust is a typed language with a nice type system, let's not fall back to strings unless really warranted

rama-http/src/matcher/mod.rs
```diff
@@ -874,6 +874,30 @@ impl<Body> HttpMatcher<Body> {
...
     }
+
+    /// Returns the set of HTTP methods this matcher explicitly allows, if determinable.
+    ///
+    /// Returns `None` when the matcher is negated, has no method constraint (meaning it would
+    /// accept any method), or contains a custom matcher whose allowed methods cannot be
+    /// inspected. A `None` result means the caller cannot enumerate a finite Allow list.
+    pub(crate) fn allowed_methods(&self) -> Option<MethodMatcher> {
+        if self.negate {
+            return None;
```
GlenDC: Negate does not work like that though... 
How it works is that if you have a `MethodMatcher::GET` and you negate it, it means that all other methods are allowed.

rama-http/src/service/web/router.rs
```diff
@@ -605,6 +611,14 @@ where
                     return service.serve(req).await;
                 }
             }
+
+            // Path matched but no method matched — collect for a potential 405.
+            // Do not return yet: a sub_service may still handle this request.
+            for (matcher, _) in matched.value.iter() {
```
GlenDC, line 621: This you can collect into the _typed_ header `Allow` (`rama-http-headers/src/common/allow.rs`) which uses a SmallVec with space for 7 on stack.. so most likely you can fit this nice on the stack without heap allocations that way :)

rama-http/src/service/web/router.rs
```diff
@@ -700,6 +714,18 @@ where
             }
         }
 
+        // A route matched the path but no registered method matched, and no sub_service
+        // handled the request — return 405 with the Allow header per RFC 7231.
+        if !allowed_methods.is_empty() {
+            let allow_val = allowed_methods.into_iter().collect::<Vec<_>>().join(", ");
```
GlenDC: Once you use a typed method you can do something like

```
return Ok((
     Headers::single(allow_header),
     StatusCode::METHOD_NOT_ALLOWED,
).into_response())
```

ikorason:
@GlenDC thank you for the feedbacks, please check again and let me know, thanks!

rama-http/src/matcher/mod.rs
```diff
@@ -874,6 +874,52 @@ impl<Body> HttpMatcher<Body> {
...
+    /// "every known method except GET", returned via the bitwise complement.
+    pub(crate) fn allowed_methods(&self) -> Option<MethodMatcher> {
+        match (&self.kind, self.negate) {
+            // Non-negated method: the matcher's own bits.
+            (HttpMatcherKind::Method(m), false) => Some(*m),
+            // Negated method: every known method except those in m.
+            (HttpMatcherKind::Method(m), true) => Some(m.complement()),
+            // Non-negated: delegate to kind.
+            (_, false) => self.kind.allowed_methods(),
+            // Negated All/Any: too complex to enumerate — treat as unconstrained.
```
GlenDC: Why is that difficult to do? Maybe I am missing something here that you figured out, but isn't this what yo uneed in psuedo code:

```
all => ALL_KNOWN & M1 & M2 & .... & Mn
any => M1 | M2 | .... | Mn
```

?? Or am I missing something

Negating is simply that result but reversed?  In fact currently you are not calculating correct... What you need is:

```
let m = match &self.kind { ... }
if self.negate { m.complement() } else { m }
```

These 2 combined covers all that you have I think, no?

Chained matchers ( all / any) which contain no Method matcher should just be ignored in computation. So those can return None as to not give weird results.

rama-http/src/service/web/router.rs
```diff
@@ -700,6 +718,19 @@ where
             }
         }
 
+        // A route matched the path but no registered method matched, and no sub_service
+        // handled the request — return 405 with the Allow header per RFC 7231.
+        if let Some(matcher) = allowed_methods {
+            let methods: SmallVec<[Method; 7]> = matcher.iter().collect();
```
GlenDC: i would replace this logic with

```
if let Some(matcher) = allowed_methods 
     && let Some(methods) = NonEmptySmallVec::collect(matcher.iter()) { 
     return Ok((
                    Headers::single(Allow(methods)),
                    StatusCode::METHOD_NOT_ALLOWED,
                )
                    .into_response())
 }
```

Easier and same result.

rama-http/src/matcher/mod.rs
```diff
@@ -874,6 +874,52 @@ impl<Body> HttpMatcher<Body> {
...
+
+    /// Returns the set of HTTP methods this matcher explicitly allows, as a [`MethodMatcher`].
+    ///
+    /// Returns `None` when the allowed method set cannot be enumerated — e.g. a
+    /// negated `All`/`Any` matcher, or a non-method constraint (Path, Domain, Custom, …)
+    /// that imposes no method restriction.
+    ///
+    /// Negated `Method` matchers are handled correctly: a negated `GET` matcher means
+    /// "every known method except GET", returned via the bitwise complement.
+    pub(crate) fn allowed_methods(&self) -> Option<MethodMatcher> {
```
GlenDC:
```suggestion
    pub fn allowed_methods(&self) -> Option<MethodMatcher> {
```

This one can be exposed :) as it is used outside this module anyway. This allows others implementing their own router or service to also make use of it.

rama-http/src/matcher/method.rs
```diff
@@ -47,6 +59,42 @@ impl MethodMatcher {
     pub const fn or(self, other: Self) -> Self {
         Self(self.0 | other.0)
     }
+
+    /// Performs the AND operation between the [`MethodMatcher`] in `self` with `other`.
+    ///
+    /// Useful for intersecting two method sets (e.g. inside an `All` compound matcher).
+    #[must_use]
+    pub(crate) const fn and(self, other: Self) -> Self {
```
GlenDC:
```suggestion
    pub const fn and(self, other: Self) -> Self {
```

no reason to hide this one

GlenDC: That said I would rename this and the or one with `or_method` and `and_method` as to not confuse it with the `Matcher` and/or. Not your fault but it's part of improving the codebase when you encounter stuff

rama-http/src/matcher/method.rs
```diff
@@ -47,6 +59,42 @@ impl MethodMatcher {
...
+        Self(self.0 & other.0)
+    }
+
+    /// Returns the complement: all known methods NOT in `self`.
+    ///
+    /// Used when a method matcher is negated — e.g. `NOT GET` → every known method except GET.
+    #[must_use]
+    pub(crate) const fn complement(self) -> Self {
```
GlenDC:
```suggestion
    pub const fn complement(self) -> Self {
```

also fine to expose

rama-http/src/matcher/method.rs
```diff
@@ -38,6 +38,18 @@ impl MethodMatcher {
...
+        .or(Self::DELETE)
+        .or(Self::GET)
+        .or(Self::HEAD)
+        .or(Self::OPTIONS)
+        .or(Self::PATCH)
+        .or(Self::POST)
+        .or(Self::PUT)
+        .or(Self::TRACE);
+
     pub(crate) const fn contains(self, other: Self) -> bool {
```
GlenDC:
```suggestion
    pub const fn contains(self, other: Self) -> bool {
```

also fine to expose, not your fault, but see no reason to keep these hidden.

rama-http/src/matcher/mod.rs
GlenDC: In that context i Thiink you can merge the MatcherKind's allow_methods with this one. I do not see a reason to have them separate.

GlenDC: If you do keep them separate than this one would basically just become (which also is fine for me, up to you):

```
let m = self.kind.allow_methods();
if self.negate { m.complement() } else { m }
```

GlenDC (changes requested):
Great, have some more feedback, hopefully last round, unless you come with more surprises :D Keep it up!

GlenDC:
Looks good. Will merge it early next week. Need to first get another PR merged, for now CI is broken. Not your fault.

ikorason:
No worries, thanks for the feedbacks learned a lot ;)

PR #845, opened 2026-03-19
fix(tls-rustls): constrain ALPN based on TargetHttpVersion before handdshake

seuros:
The rustls connector only validated TargetHttpVersion after TLS negotiation and errored out when the negotiated version mismatched.

This caused WebSocket connections to fail when servers support h2: the websocket builder correctly sets TargetHttpVersion(HTTP/1.1) since WebSocket requires HTTP/1.1 Upgrade, but TlsConnectorData defaults to advertising both h2+h1.1 in ALPN. The server picks h2, then the post-handshake check errors with "TargetHttpVersion incompatible with tls ALPN negotiated version".

Now both ConnectorKindAuto and ConnectorKindSecure resolve connector data early and call constrain_alpn_for_target_version() to override ALPN protocols before the TLS handshake.

### DIFF
--- a/rama-tls-rustls/src/client/connector.rs
+++ b/rama-tls-rustls/src/client/connector.rs
@@ -213,7 +213,7 @@ where
             transport_ctx.app_protocol,
         );
 
-        let connector_data = input.extensions().get::<TlsConnectorData>().cloned();
+        let connector_data = self.connector_data(&input)?;
 
         let (stream, negotiated_params) = self
             .handshake(connector_data, Some(server_host), conn)
@@ -266,7 +266,7 @@ where
 
         let server_host = &transport_ctx.authority.host;
 
-        let connector_data = input.extensions().get::<TlsConnectorData>().cloned();
+        let connector_data = self.connector_data(&input)?;
 
         let (conn, negotiated_params) = self
             .handshake(connector_data, Some(server_host), conn)
@@ -312,7 +312,7 @@ where
             });
         };
 
-        let connector_data = input.extensions().get::<TlsConnectorData>().cloned();
+        let connector_data = self.connector_data(&input)?;
 
         let (conn, negotiated_params) = self
             .handshake(connector_data, maybe_server_host, conn)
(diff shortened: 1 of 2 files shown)


GlenDC (changes requested):
I like your approach... you are right that rustls connector indeed not yet supporting setting ALPN based on the version. However your approach is not desired. Right now you are _always_ setting http ALPN (auto by default). That is not desired. Also no need to do all this manual work of setting to 1 and then the other. What I would do instead is to do the following: 

1. get connector data... leave None if undefined
2. in http only: check if targt version: if no data use the creator methods try_new_http_1, try_new_http_2. Or if we did have connector data, use setters to set it

For(2) in case you need to set the ALPN for existing connector data... you will need to create setters for it for http1 and http2, currently it only has an http audo setter

```
 rama_utils::macros::generate_set_and_with! {
        /// Set alpn protocols to most commonly used http protocols:
        /// [`ApplicationProtocol::HTTP_2`], [`ApplicationProtocol::HTTP_11`]
        pub fn alpn_protocols_http_auto(mut self) -> Self {
            self.set_alpn_protocols(&[ApplicationProtocol::HTTP_2, ApplicationProtocol::HTTP_11]);
            self
        }
    }
```

^ so simimilar to that one make it also for HTTP1 and HTTP2 :) then use that

return the connector data:

- (can be none if not http or not target version)
- or the modified/created (Some) data  (w/ version)

This way it is only set to an HTTP ALPN if we actually use http and have that version set. Let me know on discord if you have a question

rama-tls-rustls/src/client/connector.rs
```diff
@@ -380,6 +394,47 @@ impl<S, K> TlsConnector<S, K> {
...
+    };
+
+    let target_alpn = ApplicationProtocol::try_from(target_version.0)?;
+    tracing::trace!(
+        ?target_version,
+        ?target_alpn,
+        "resolving TLS connector data to match TargetHttpVersion",
+    );
+
+    let connector_data = match (connector_data, target_alpn) {
```
GlenDC: You can simplify.. just create new connector data if it doesn't exist and after that set the custom alpn, give ntaht is anyway an option you are using for both new and existing... Might as well stop checking the ALPN's you know :)

rama-tls-rustls/src/client/connector.rs
```diff
@@ -380,6 +394,37 @@ impl<S, K> TlsConnector<S, K> {
...
+    };
+
+    let target_alpn = ApplicationProtocol::try_from(target_version.0)?;
+    tracing::trace!(
+        ?target_version,
+        ?target_alpn,
+        "resolving TLS connector data to match TargetHttpVersion",
+    );
+
+    let connector_data = connector_data.unwrap_or(TlsConnectorData::try_new()?);
```
GlenDC: Ok(Some(connector_data
    .map(Ok)
    .unwrap_or_else(TlsConnectorData::try_new)?
    .with_alpn_protocols(&[target_alpn])))

GlenDC: to replace all this code, same same

GlenDC (changes requested):
can be simpler

rama-tls-rustls/src/client/connector.rs
seuros: Elegant!

TIL

PR #853, opened 2026-03-28
refactor(time): switch runtime chrono usage to jiff

### LINKED ISSUE #849: replace chrono with jiff
Chrono is becoming deprecated. Let's replace it with jiff instead.

https://crates.io/crates/jiff

Only dependencies of rama that use chrono are:

- tracing-appender
- tracing-opentelemetry

Feel freel to replace the use of chrono there as well as extra work.

seuros:
This PR  closes #849 

Chrono still being pulled by  tui-logger.

Also i did not find how to validate https://datatracker.ietf.org/doc/html/rfc3339 in jiff, the crate has https://datatracker.ietf.org/doc/html/rfc2822 support only.

This is another breaking change that we need to add in the release note.

### DIFF
--- a/ffi/apple/examples/transparent_proxy/tproxy_ffi_e2e/Cargo.lock
+++ b/ffi/apple/examples/transparent_proxy/tproxy_ffi_e2e/Cargo.lock
@@ -46,15 +46,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android_system_properties"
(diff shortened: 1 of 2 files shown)


GlenDC:
```
cargo tree -i chrono --all-features
error: package ID specification `chrono` did not match any packages
```

Together with main patch :)

PR #854, opened 2026-03-28
rama-dns improvements

GlenDC:
- add windows native dns impl (backed by `DnsQueryEx`)
- make use of this dns impl by default for global dns resolver
- same on linux but using res_nquery / getaddrinfo
- fallback to tokio
- hickory now opt-in. Gives user the option and also without having to hide failures or make undesired choices

### DIFF
--- a/rama-dns/Cargo.toml
+++ b/rama-dns/Cargo.toml
@@ -38,7 +38,7 @@ parking_lot = { workspace = true }
 [target.'cfg(target_os = "linux")'.dependencies]
 ahash = { workspace = true }
 libc = { workspace = true }
-moka = { workspace = true }
+moka = { workspace = true, features = ["sync"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 windows-sys = { workspace = true, features = ["Win32_Foundation", "Win32_NetworkManagement_Dns"] }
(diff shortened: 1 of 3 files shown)


seuros:
We should try to have BSDs, i can add support for freebsd/openbsd for sure , netbsd maybe later.

GlenDC:
> We should try to have BSDs, i can add support for freebsd/openbsd for sure , netbsd maybe later.

Was thinking about that in last couple of days... yes
Sadly Linux is a bit of a complicated story...

Was thinking to do it using:

- `getaddrinfo` for A/AAAA
- `res_nquery` for TXT

But ... even then I will have to limit it to only _some_ linux.
For example musl doesn't contain `res_nquery` afaik, could try to use
deprecated `res_query` for musl... but that's ... global, so not sure ideal.

So there will always be some platforms where we do use hickory by default as well I think...

What did you have in mind @seuros ? Same?
Might do it tomorrow morning... Or I might just already merge this windows bit already
and move on for now.

Either way this kind of work is something I do prefer to do myself,
as there is too much I have to be paranoid about it anyway.
But all feedback and reviews welcome though, that for sure.

seuros:
I dug through the source trees for FreeBSD, OpenBSD, NetBSD, and DragonFly to see what each gives us:

FreeBSD / NetBSD / DragonFly,  All ship BIND9 resolver in libc. getaddrinfo for A/AAAA, res_nquery for TXT. Thread-safe, no -lresolv needed. spawn_blocking wraps both cleanly.

OpenBSD on other hand completely replaced BIND with ASR (Asynchronous Resolver). res_query is all what you get there. 

OpenBSD will need its own quirk later.

There are more BSDs systems, but i don't use them personally so i can't help or test.

PR #856, opened 2026-04-02
New extensions using append-only vec

soundofspace:
- Make Extensions use the new shared/append-only store.
- Adds the new access patterns get_ref, get_arc....
- Migrate everything to this new API

What we will be handled in followup PR:
- Remove blacket impl for Extension, separated so this is only focussed on api changes
- Make better use of this new logic. Goal of this PR was to implement Extensions, and do a pretty simple refactor. These other changes are more involved and will probably need some discussions, so there are better handled in individual Prs


rama-core/src/layer/get_extension.rs
```diff
@@ -126,6 +126,9 @@ impl<S, T, Fut, F> GetInputExtension<S, T, Fut, F> {
     define_inner_service_accessors!();
 }
 
+// TODO should we return Arc<T> here instead, caller can then still decide if he wants
```
GlenDC: maybe split it in 2 layers

- GetInputExtensionOwned
- GetInputExtensionRef

or however you would name them, doing only one of them is making either way a choice for the user (e.g. cloning an arc is still work that the user might not want either

rama-core/src/extensions.rs
```diff
@@ -24,133 +24,288 @@
...
+pub struct Egress<T>(pub T);
+
+#[derive(Debug, Clone)]
+pub struct Connection<T>(pub T);
+
+#[derive(Debug, Clone)]
+pub struct Stream<T>(pub T);
+
+#[derive(Debug, Clone)]
+pub struct Input<T>(pub T);
```
GlenDC, line 307: What is difference in Ingress Vs Input ?

PR #867, opened 2026-04-08
Remove blacket impl for Extension trait

soundofspace:
Remove blacket impl for Extension so everything we store needs to implement this (or derive it).

This gives a lot of power going forward, and also makes sure that everything that is stored in there is unique.

This also removes the Extension extractor like we discussed earlier since it doesn't really give a lot of useful vs just using extensions/request directly where you can use get_ref/get_arc, and properly handling missing ones


soundofspace:
@GlenDC open question: do we want to impl it for things like `Vec<T:Extension>` and some other containers. Personally I'm thinking no: everything should have a proper wrapper. But curious what you think

GlenDC:
> @GlenDC open question: do we want to impl it for things like `Vec<T:Extension>` and some other containers. Personally I'm thinking no: everything should have a proper wrapper. But curious what you think

For containers would not do no. But perhaps for something like `Arc` I can see why you might want it :)

soundofspace:
> > @GlenDC open question: do we want to impl it for things like `Vec<T:Extension>` and some other containers. Personally I'm thinking no: everything should have a proper wrapper. But curious what you think
> 
> For containers would not do no. But perhaps for something like `Arc` I can see why you might want it :)

Actually it's exactly the opposite: for Arc it's even a feature that it's not implemented. Before it was easy to save an Arc by doing insert(Arc<Value>), now this is impossible since it doesn't implement Extension. So you have to use insert_arc(Value) which is more efficient (same for get_ref(Value)/get_arc(Value)). So by not implement it for Arc it has actually become impossible to misuse this API

GlenDC:
Fair enough. In that case no need to do it for that either, makes sense. I would add a couple of lines in the module docs for extensions to clarify why we didn't implement it for things like collections, arc, etc... So that we still remember this in 5 years

rama-ua/src/profile/runtime_hints.rs
```diff
@@ -25,9 +26,24 @@ impl PreserveHeaderUserAgent {
 }
 
 /// ClientHints requested for the (http) Request.
-pub type RequestClientHints = Vec<ClientHint>;
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Default, Extension)]
+pub struct RequestClientHints(pub Vec<ClientHint>);
```
GlenDC: Was already as like this, but not better we store this as `arc<[ClientHint]>`. Not sure why I went for `Vec` here initially...

rama-ua/src/layer/emulate/service.rs
```diff
@@ -413,8 +413,8 @@ where
                             protocol,
                             Some(req.method()),
                             req.extensions()
-                                .get_ref::<Vec<ClientHint>>()
-                                .map(|v| v.as_slice()),
+                                .get_ref::<RequestClientHints>()
```
GlenDC, line 416: Perhaps we should support also `get_deref`? similar to what `Option` also implements? Woud simplify this one, no?

rama-macros/src/extension_macro/extension.rs
```diff
@@ -0,0 +1,63 @@
...
+    for param in item.generics.params.iter() {
+        if matches!(param, GenericParam::Lifetime(_)) {
+            return Err(syn::Error::new_spanned(
+                &item.generics,
+                "`#[derive(Extension)]` doesn't support lifetime parameters",
+            ));
+        }
+    }
+
+    for param in item.generics.params.iter_mut() {
```
GlenDC, line 19: took me a while to understand this as I am not every day doing macros, can you add 1 or 2 lines above this loop to describe the intent please

rama-http-types/src/request.rs
```diff
@@ -114,7 +117,7 @@ impl<T> From<Request<T>> for HyperiumRequest<T> {
         let mut hyper_extensions = parts
             .extensions
             .get_ref::<HyperExtensions>()
```
GlenDC, line 119: I wonder if we should make a shortcut for this `get_cloned` or something, or do you consider this exceptional use, e.g. here needed because cross-ecosystem?

GlenDC (approved):
Looks good to me. Some minor feedback, but nothing that would change this design or approach. So you can resolve the comments as you see fit and merge afterwards.

rama-ua/src/layer/emulate/service.rs
soundofspace: Dont think that's possible, option can do that since it knows its internal type, but extensions needs to know the  exact type it needs to get for TypeId logic

rama-ua/src/profile/runtime_hints.rs
soundofspace: Yea makes sense I did notice that we never seem to construct this type, is this future proofing?

rama-http-types/src/request.rs
soundofspace: I would consider this a special case since we need to wrap. In almost all cases where you want a clone you would either do: `get_arc()` or `get_ref().cloned()`. For the latter I don't really see much benefit in adding get_cloned(). I would rather keep Extensions impl small and focussed (it's already quite big and complex), and then let users work with an Option<T>, which gives them all the things they are used to

rama-ua/src/layer/emulate/service.rs
GlenDC: ok fair enough

rama-ua/src/profile/runtime_hints.rs
GlenDC: no idea, would have thought we aleady construct it for UA emulation, but perhaps not. Would have to look into it, it's been very long ago

PR #869, opened 2026-04-12
Tags for extensions using traits

soundofspace:
Add traits such as `TlsExtension`, `HttpExtension` that can be implement to "tag" extension. This makes it very easy to find them in rust docs and also allows us to group similar extensions. 

This also extends the derive `Extension` macro to support a syntax like this to make this ergonomic
```rust
#[derive(Debug, Clone, Extension)]
#[extension(tags(tls, net))]
pub struct SecureTransport;
```

### DIFF
--- a/rama-core/src/graceful.rs
+++ b/rama-core/src/graceful.rs
@@ -7,4 +7,5 @@ pub use ::tokio_graceful::{
     Shutdown, ShutdownBuilder, ShutdownGuard, WeakShutdownGuard, default_signal,
 };
 
+// TODO @glendc, remove usage from this in TransparentProxyEngine
 impl Extension for ShutdownGuard {}
--- a/rama-grpc/src/status.rs
+++ b/rama-grpc/src/status.rs
@@ -44,7 +44,8 @@ const ENCODING_SET: &AsciiSet = &CONTROLS
 /// assert_eq!(status1.code(), Code::InvalidArgument);
 /// assert_eq!(status1.code(), status2.code());
 /// ```
-#[derive(Clone)]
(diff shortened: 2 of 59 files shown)


rama-core/src/graceful.rs
```diff
@@ -1,10 +1,12 @@
...
+use crate::extensions::{Extension, NetExtension};
 
 #[doc(inline)]
 pub use ::tokio_graceful::{
     Shutdown, ShutdownBuilder, ShutdownGuard, WeakShutdownGuard, default_signal,
 };
 
 impl Extension for ShutdownGuard {}
+// TODO how do want to tag this one?
+impl NetExtension for ShutdownGuard {}
```
GlenDC: not, don't think we want to actually still put it in Extension, that guard? Is that still being used somewhere other than an example?

rama-grpc/src/status.rs
```diff
@@ -71,6 +75,7 @@ impl StatusInner {
 }
 
 impl Extension for Status {}
```
GlenDC: why these are manual impls instead of derive? is there a reason why it cannot be used here?

GlenDC (approved):
Looks good... Just 2 minor comments

rama-grpc/src/status.rs
soundofspace: i see its used in TransparentProxyEngine

GlenDC: Wrong comment. Can you add a TODO for glendc. I will clean that up in next week(s). Wanna make a second version of that code with what learned so far. Can also take that with me

PR #873, opened 2026-04-17
Custom Debug impl for Extensions

KoHcoJlb:
Hi
While exploring your interesting project, I noticed that recently, after it's storage was changed to AppendOnlyVec, `Extensions`' automatic `Debug` impl became rather uninformative:
`Extensions { extensions: AppendOnlyVec { count: 5, reserved: 5, data: [0x10263ac80, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0] }`
So this PR adds an impl which formats it as an array:
`[Extension1, Extension2, Extension3]`

### DIFF
--- a/rama-core/src/extensions.rs
+++ b/rama-core/src/extensions.rs
@@ -32,6 +32,7 @@
 //! ```
 
 use std::any::{Any, TypeId};
+use std::fmt;
 use std::pin::Pin;
 use std::sync::Arc;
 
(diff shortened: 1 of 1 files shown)


soundofspace (commented):
Thanks for the PR, this is indeed a nice improvement!

soundofspace (approved):

GlenDC (approved):

PR #877, opened 2026-04-24
Don't compress absent response body

maartendeprez:
Some HTTP responses are not allowed to have body content according to the standard:

- responses to HEAD and CONNECT requests
- responses with status codes 100-199 (informational), 204 (no content) and 304 (not modified)

For such responses, the server will usually not set a `content-length` header, causing he default compression `Predicate` (with `SizeAbove` defaulting its decision to `true`) to advise in favor of compressing the empty body.

### DIFF
--- a/rama-http/src/layer/compression/service.rs
+++ b/rama-http/src/layer/compression/service.rs
@@ -6,6 +6,7 @@ use crate::headers::encoding::{AcceptEncoding, Encoding, parse_accept_encoding_h
 use crate::layer::remove_header::remove_payload_metadata_headers;
 use crate::layer::util::compression::WrapBody;
 use crate::{Request, Response, header};
+use http::Method;
 use rama_core::Service;
 use rama_core::extensions::ExtensionsRef;
 use rama_http_headers::specifier::QualityValue;
@@ -163,11 +164,14 @@ where
     async fn serve(&self, req: Request<ReqBody>) -> Result<Self::Output, Self::Error> {
         let accepted_encodings: SmallVec<[QualityValue<Encoding>; 4]> =
             parse_accept_encoding_headers(req.headers(), self.accept).collect();
-
+        let req_method = req.method().clone();
         let mut res = self.inner.serve(req).await?;
         let mut respected_encoding = None;
 
         let should_compress =
+            //never compress responses without body
+            !matches!(req_method, Method::HEAD | Method::CONNECT) &&
+            !matches!(res.status().as_u16(),  100..200 | 204 | 304) &&
             //never compress responses that are ranges
             !res.headers().contains_key(header::CONTENT_RANGE) &&
             self.predicate.should_compress(&mut res) &&


GlenDC:
@maartendeprez can you please tell a bit about your compress/decompres setup ?
What behaviour are you seeing and what do you wish to see differently?

Know that there are since some weeks/months also some new things that rama provides to you to make the compression/decompression more flexible and adaptable for proxy needs?

maartendeprez:
I can make it work using the tools available, but the solution feels hacky. Maybe i am missing something?

When we send a `HEAD` request to `https://www.example.com/` with an 'Accept-Encoding' header (like a browser does), the server responds with a `Content-Encoding` header, no `Content-Length` header and no body (no data frame). Some servers respond with `Content-Encoding` and (non-zero) `Content-Length` headers. A zero-length data frame is sometimes sent in the response as well.

If this request goes through an MITM proxy with `CompressionLayer` with default predicate, that layer may decide to compress the (empty) body if the incoming request had `Accept-Encoding` set. (Since the response body length is unknown, the predicate defaults to `true`.) This results in a response with `Content-Encoding` header and a short (e.g. twenty bytes for gzip or one-byte for brotli) body. (Of course, if we decompress the body, it reveals itself as zero-length.) Some tools choke on this.

The issue can be solved by not setting an `accept-encoding` header in the original request. When the MITM proxy is configured to avoid recompression, it also works "by accident" when the `Accept-Encoding` header in the incoming request is compatible with the encoding used by the server. 

What i believe should happen, is that compression and decompression layers leave the response body for HEAD requests alone since it is not supposed to be there in the first place. Response headers like `Content-Length` and `Content-Encoding` are not supposed to apply to that response body either.


The HTTP standard [says](https://httpwg.org/specs/rfc9110.html#HEAD):

> The HEAD method is identical to GET except that the server MUST NOT send content in the response. HEAD is used to obtain metadata about the [selected representation](https://httpwg.org/specs/rfc9110.html#selected.representation) without transferring its representation data

It is a matter of interpretation whether or not sending a compressed empty body counts as sending content. 

On the other hand, MDN also [notes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD):

> Warning: If a response to a HEAD request has a body, the response body must be ignored. Any [representation headers](https://developer.mozilla.org/en-US/docs/Glossary/Representation_header) that describe the erroneous body are assumed to describe the response body that a GET request would have received.

GlenDC:
@maartendeprez two things here.

1. A MITM proxy usually should have no opinion at all on what is allowed or not, but simply touch what it needs to touch and respect anything else as-is... it is not there to validate if HTTP specifications are followed, given the wild is sadly not as pure as that.
2. I do agree that your input here is valid for servers and thus valuable

For point (1) you might want to look at https://github.com/plabayo/rama/blob/main/examples/http_mitm_proxy_boring.rs. You'll see that using a combination of :

```rust
// ensures that we support decompressing but do not add the accept encoding our header
// which means that we will only decompress if (a) the original client supports encoding and (b)
// the server encoded the content
DecompressionLayer::new().with_insert_accept_encoding_header(false)

// Compress only in case the original received response was also compressed, and if so try to respect
// that encoding
CompressionLayer::new().with_compress_predicate(MirrorDecompressed::new())
```

In some proxies I go even fancier and add an additional predicate to decompression that we only do it
in case we actually want to inspect the response payload.

For point (1), useful if you make servers or in case you have a kind of (MITM) proxy which also emulates user agents,
I would propose that you (for this PR) do the following:

- add missing predicates in `rama-http/src/layer/compression/predicate.rs` for StatusCodes and Methods and also provide constructors for the default use-case for these predicates
- add these predicates with their default use case(s) to the default predicate chain
- add unit tests to confirm this behaviour is working and remains working in future

That I would recommend instead of adding more to the compression layer,
given that would otherwise force choices for _all_ users.

Does that work for you @maartendeprez or am I missing something? Do feel free to ping me here on on discord if you have concerns or require additional feedback. Or in case you want to walk through some things together.

CC @soundofspace

maartendeprez:
@GlenDC The solution using predicates (2) would work for me, yes.

The main reason i did not implement it like that in the first place, was that i did not see a portable way to find the request method from the response in the predicate implementation.  Is that already inserted as an extension by some layer? Is there another way i am not aware of? If not, i see a few alternatives:

1. Insert a `Method` wrapper extension from the Compression service, for use by the `RequestMethod` predicate.
2. Modify the `Predicate` trait, adding an extra parameter to the `should_compress` method.

What do you prefer?


As for (1) I agree in principle that a proxy should be unopinionated and should avoid modifying the requests and responses unless it intends to do so. However, I believe the proposed solution does not achieve that either. While avoiding recompression altogether does of course solve the issue, using the `MirrorDecompressed`  predicate can still cause the compression layer to decide to recompress an empty response body, yielding a non-empty response body. The steps are as follows:

- The client sends a HEAD request with `Accept-Encoding: br` header
- The MITM proxy forwards the request to the destination server
- The server responds with "200 OK", `Content-Encoding: br` and an empty body (empty body or no data frame at all)
- In the MITM proxy:
   - The `DecompressionLayer`:
      - Sees the `Content-Encoding` response header and decides to decompress the body. (It erroneously assumes the header applies to _this_ body rather than to the body that _would_ be sent in response to a GET request.)
      - Decompresses the empty body to an empty body
      - Inserts the  `DecompressedFrom::Brotli` extension
   - In`CompressionLayer`:
      - The `MirrorDecompressed` predicate sees `DecompressedFrom::Brotli` and decides to recompress
      - The empty (decompressed) body gets recompressed to a (compressed) one-byte body
- The client sees a response with `Content-Encoding: br` as in the original response from the server, but with a one-byte body instead of an empty one

To make the process really unopinionated in the proxy, we'd need to be able to detect the absence / emptiness of the body and preserve that property.

GlenDC:
@maartendeprez you make some excellent points.
Let's for now ship this and we can worry about outfall later, if there ever is even one.

Only thing I would like to ask you is that you

1. add references to relevant specifications and their sections, so we have that properly tracked as reference (in code comments)
2. add some relevant unit tests so we can preserve this behaviour even through refactors (ideally high level enough that they do not rely on underlying implementation details), but still within that module or close enough to it

GlenDC:
cherry-picked + did my requests in https://github.com/plabayo/rama/pull/893. Will be merged soon.

Issue #876, opened 2026-04-23
Consider making Output and Error of IntoEndpointService and Router generic


KoHcoJlb:
Given that *rama* looks like a very flexible and versatile library, at least to me, where many things can be composed together in different ways using the `Layer` and `Service` traits, I think it’s somewhat limiting that the HTTP `Router`, `IntoEndpointService`, and all surrounding machinery are locked to `Output = Response` and `Error = Infallible`.

There's nothing inherently wrong with it per se, for example `axum` does the same, but I believe *rama*, with its powerful `Service` ecosystem, can do better.

#### So my proposal is this:

- Make `Router`, `IntoEndpointService`, and related types generic over `Output` and `Error`, with only slight `Error: From<_>` bounds to ensure that Router and extractor errors can be converted into the chosen error type
- Handle conversion from each endpoint's return type to the `Router` types within `with_*` and `set_*` methods, using a `Layer` stored in the `Router`
- Conversion into `Response` can be layered on top of the `Router`, for example by using existing `ErrorHandler` service

This way, library users gain greater control over how types are converted into responses and can perform additional processing, such as logging, while making use of original types. It would also allow `EndpointServiceFn` to be used in more scenarious, layering other services onto it, since it would no longer convert output types by itself.

#### Background

I came to this idea while building an http api server using *rama*.

I first ran into this when I wanted to use `eyre` for error handling. My initial thought was to write a `Service` that would convert its errors into responses, which is when I quickly discovered that function's return type must implement `IntoResponse`. This was solved by creating wrapper error type and implementing `IntoResponse` for it, but still I've lost `eyre`s convenient blanket `From<Error>` implementation.

The second case was when I wanted to add error logging to all my routes. I wasn't been able to solve this one because the error value never gets past `EndpointServiceFn` and if I wrap functions into services manually I lose automatic extractor functionality.

At the moment I only have use cases for Error type, but I believe custom `Output` could also be useful in certain situations.

If you are interested in such changes, I'd be happy to work with you on this.  
I can also share my preliminary work on the matter, which I've done to ensure that this concept plays well with the type system, in a draft PR.

GlenDC:
Hi @KoHcoJlb I am open to explore it for sure...

There is a lot in here though and I think we best explore it in small steps...

First step is we currently today have `IntoResponse`...
[...]

PR #878, opened 2026-04-24
Make HTTP services generic over Output and Error types

KoHcoJlb:
Implementation of #876


KoHcoJlb:
So @soundofspace, here is what I've got at the moment.
Some code is commented out for now to allow building the library during refactor.

Notable PoIs are:
- of course [IntoEndpointService](https://github.com/plabayo/rama/pull/878/changes#diff-1638cd7b5356f68141457e035cf82338904ff6c59ec4b0702c6161990023343b) and [Router](https://github.com/plabayo/rama/pull/878/changes#diff-f975f345178ac8f7226e02c839cb36073953ea8b940f15bd14b7760ed01dcb16) themselves
- you can see how it works together in [examples/test_new_router.rs](https://github.com/plabayo/rama/pull/878/changes#diff-868d06795c41db65e2010ef4aabfa727fdfb66e6beb6b27425fff98c66ded613)
- and also [rama-http/src/service/web/error.rs](https://github.com/plabayo/rama/pull/878/changes#diff-55f4c0d35917abd564f8f8f43926bc2d15be5c2387b763e04ae895b5f2fee7dc)
  It's my take on how `IntoResponse` implementation of errors can be kept accessible where they later will be converted into responses

soundofspace (changes requested):
This already looking promising. If we can move layers outside of Router then I think this looks really feasible.

rama-http/src/service/web/router.rs
```diff
@@ -36,43 +38,59 @@ use rama_utils::{
...
 #[allow(unused)]
-pub struct Router<State = ()> {
-    routes: MatchitRouter<Vec<(HttpMatcher<Body>, BoxService<Request, Response, Infallible>)>>,
-    sub_services: Option<Trie<String, SubService>>,
-    not_found: Option<BoxService<Request, Response, Infallible>>,
+pub struct Router<O, E, L = (), State = ()> {
+    routes: MatchitRouter<Vec<(HttpMatcher<Body>, BoxService<Request, O, E>)>>,
+    sub_services: Option<Trie<String, SubService<O, E>>>,
+    not_found: Option<BoxService<Request, O, E>>,
+    layer: L,
```
soundofspace: Why is this needed? I would think that we can keep router like it is right now, but just make it generic over `Result<O, E>`. Since router is also a Service, it should just be possible to place these layers on top of Router instead of in it.

So instead of `Router<Layers>` it would become `layers.into_layer(router)` this way it is just normal layering and the router stays simple (and it just does routing).

I might be missing something at first sight here why this might be needed, but seems like it should be possible

KoHcoJlb: Layer here serves one important role, it's used when adding endpoints to the Router to convert from endpoint's `<O, E>` into Router's `<O, E>`. Otherwise all endpoints would need to have the same return type.

It can also be used to do something with the original types before they are converted, like log all the outputs (for example when endpoints return `Json<T>` it can look better using rust's builtin Debug impl).

soundofspace: Ah yea totally overlooked that. To stay consistent with how we do this in other places I would expose it like this
- new() or new_with_state(), creates a Router that mimics how the Router works before this change. So I think internally it would create a Router with these layers `MapResponseLayer, IntoErrLayer` probably
- Add a new method `with_endpoint_layer(self, layer: L) -> Router<..., L>`. This would make it possible replace the default logic and support special use cases while keeping the default easy

And something along these lines to explain what it does
```
/// Apply `layer` to every endpoint registered after this call. Routes registered before this call keep whatever layer was in effect at the time of registration.
```

After that change I'll do a quick high level review and then you can fully implement this and all needed details

KoHcoJlb: Done
I've also reordered and added defaults to type parameters reflecting default use case.

rama-http/src/service/web/endpoint/extract/mod.rs
```diff
@@ -56,7 +56,7 @@ pub use option::{OptionalFromPartsStateRefPair, OptionalFromRequest};
 pub trait FromPartsStateRefPair<State>: Sized + Send + Sync + 'static {
     /// If the extractor fails it'll use this "rejection" type. A rejection is
     /// a kind of error that can be converted into a response.
-    type Rejection: IntoResponse;
+    type Rejection;
```
soundofspace: @GlenDC this kinds of is an Error now, should we rename it to that?

rama-http/src/service/web/endpoint/response/layer.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::service::web::response::IntoResponse;
+
+/// A [`Service`] that maps response for an inner service using [`IntoResponse`].
+#[derive(Debug, Clone)]
+pub struct IntoResponseService<S>(S);
+
+impl<S, R, E> IntoResponseService<S>
+where
+    S: Service<Request, Output = R, Error = E>,
+    R: IntoResponse + Send + Sync + 'static,
```
soundofspace:
```suggestion
impl<S> IntoResponseService<S>
where
    S: Service<Request, Output: IntoResponse>,
```
No need to name error here. Also if you use associative types the bounds become easier

rama-http/src/service/web/endpoint/response/layer.rs
```diff
@@ -0,0 +1,47 @@
...
+
+impl<S, I, O, E> Service<I> for IntoResponseService<S>
+where
+    S: Service<I, Output = O, Error = E>,
+    I: Send + 'static,
+    O: IntoResponse + Send + Sync + 'static,
+    E: Send + 'static,
+{
+    type Output = Response;
+    type Error = E;
```
soundofspace:
```suggestion
impl<S, I> Service<I> for IntoResponseService<S>
where
    S: Service<I, Output: IntoResponse>,
    I: Send + 'static,
{
    type Output = Response;
    type Error = S::Error;
```

By using the associative types these bounds can be simplified, since they already enforce Send (+Sync) + 'static

rama-http/src/service/web/endpoint/response/layer.rs
soundofspace: And just as reference, if these assiociative types become to much for a single line you can also split them up like this:
```
S: Service<Request>
S::Output: IntoResponse
```

rama-http/src/service/web/endpoint/response/layer.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    async fn serve(&self, req: I) -> Result<Self::Output, Self::Error> {
+        self.0.serve(req).await.map(IntoResponse::into_response)
+    }
+}
+
+/// A [`Layer`] that maps response for an inner service using [`IntoResponse`].
+#[derive(Debug, Clone)]
+pub struct IntoResponseLayer;
```
soundofspace, line 37: This will probably never be extended, but still to keep that option open, we use this pattern everywhere: default() / new() to create these layers
```
#[non_exhaustive]
pub struct IntoResponseLayer;
```

rama-http/src/service/web/endpoint/mod.rs
```diff
@@ -136,14 +110,14 @@ mod service;
 pub use service::EndpointServiceFn;
 
 /// Wrapper svc used for creating a endpoint service from a function.
-pub struct EndpointServiceFnWrapper<F, T, State> {
+pub struct EndpointServiceFnWrapper<F, T, O, E, State> {
     inner: F,
-    _marker: std::marker::PhantomData<fn(T) -> ()>,
+    _marker: std::marker::PhantomData<fn(T, O, E) -> ()>,
```
soundofspace:
```suggestion
    _marker: std::marker::PhantomData<fn(T) -> Result<O, E>>,
```

rama-http/src/service/web/endpoint/mod.rs
```diff
@@ -136,14 +110,14 @@ mod service;
...
     state: State,
 }
 
-impl<F: std::fmt::Debug, T, State: std::fmt::Debug> std::fmt::Debug
-    for EndpointServiceFnWrapper<F, T, State>
+impl<F: std::fmt::Debug, T, O, E, State: std::fmt::Debug> std::fmt::Debug
+    for EndpointServiceFnWrapper<F, T, O, E, State>
 {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("EndpointServiceFnWrapper")
```
soundofspace, line 123: Update _marker line also

Cargo.toml
```diff
@@ -432,6 +432,7 @@ spmc = { workspace = true }
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 tokio-test = { workspace = true }
 turmoil = { workspace = true }
+derive_more = { workspace = true, features = ["display", "error"] }
```
soundofspace: @GlenDC what do you think about using this?

rama-http/src/service/web/error.rs
```diff
@@ -0,0 +1,35 @@
+use std::error::Error;
+
+use derive_more::Display;
+use rama_http_types::Response;
+
+use crate::service::web::response::IntoResponse;
+
+pub trait IntoResponseError: Error + Send + Sync + 'static {
+    fn as_response(&self) -> Response;
```
soundofspace: It seems nice to keep the original error around, but requiring clone on it is a very high cost to do so, and might not work for a lot of uses cases, so i would consume self here

rama-http/src/service/web/error.rs
```diff
@@ -0,0 +1,35 @@
...
+}
+
+impl<T: IntoResponse + Clone + Error + Send + Sync + 'static> IntoResponseError for T {
```
soundofspace: For complex bounds we always prefer to use where: ...

rama-http/src/service/web/error.rs
soundofspace: Eg example of something that would break would be:
```rust
fn handler(req: crate::Request) -> Result<crate::Response, crate::Response> {
    let error = true;

    if error {
        return Err(Response::new(Body::empty()));
    }

    Ok(Response::new(Body::empty()))
}
```

Body doesn't implement clone so this would not implement IntoResponseError and considering this is the default path this should be supported. So it will have to consume self

rama-http/src/service/web/router.rs
```diff
@@ -1,22 +1,13 @@
-use matchit::Router as MatchitRouter;
-use radix_trie::{Trie, TrieCommon as _};
 use std::{convert::Infallible, path::Path, sync::Arc};
 
-use crate::{
-    Request, Response,
-    headers::Allow,
```
soundofspace, line 12: If possible avoid huge differences in imports just because they are placed in slightly different spot and IDE moves them

rama-http/src/service/web/router.rs
```diff
@@ -30,39 +21,50 @@ use rama_utils::{
...
+    service::{
+        fs::{DirectoryServeMode, ServeDir},
+        web::{
+            IntoEndpointService, IntoEndpointServiceWithState, ResponseError,
+            response::{Headers, IntoResponse, layer::IntoResponseLayer},
+        },
+    },
+};
+
+type DefaultLayer = (IntoErrLayer<BoxError>, IntoResponseLayer);
```
soundofspace:
```suggestion
type DefaultLayer = (IntoResponseErrorLayer, IntoResponseLayer);
```

Default logic would be to always return a Result<Response, Infallible>

soundofspace: Not sure if this should be two layers, or IntoResponseLayer should make sure both O and E are converted to just Result<Response, Infallible>. Im currently leaning towards IntoResponseLayer doing both and this being the default

rama-http/src/service/web/router.rs
```diff
@@ -574,17 +623,42 @@ impl Default for Router {
...
 where
+    O: Send + 'static,
+    E: Send + From<ResponseError> + 'static,
+    L: Send + Sync + 'static,
     State: Send + Sync + Clone + 'static,
 {
-    type Output = Response;
-    type Error = Infallible;
+    type Output = O;
+    type Error = E;
```
soundofspace, line 648: Hmm this significantly changes how a router works by default, I thought we could hide everything internally with layers, but this RouterError changes that.

Ideally all Router usages can switch to this new system without any changes or only some simple changes. Some simple change could be:
- request a IntoResponse layer (IntoResponseLayer).into_service(router)
- require handlers to return Result vs just single -> T, but this might have a bigger impact then we initially think of

But that about covers the simple changes. This could be tricky to achieve or might be easy. So instead of focussing on details like I did above, lets first figure this out.

To really see how this change will impact the rest, can you add enough code so `http_web_router.rs` compiles. This way we can really see what impact these changes have on the normal use case. If there are multiple options to solve this feel free to also share them so we can iterate on that

soundofspace (commented):
This is starting to look a lot better, but we will first have to figure out how to make this change without significantly changing/breaking the current Router usages (see [this comment ](https://github.com/plabayo/rama/pull/878/changes#r3145620266))

rama-http/src/service/web/router.rs
soundofspace: This goes hand in hand with how we configure DefaultLayer and how it all plays together. So you can ignore my comment there for now, and lets just focus on the bigger picture first

rama-http/src/service/web/endpoint/response/layer.rs
KoHcoJlb: That was a good catch, thanks, looks a lot cleaner this way

rama-http/src/service/web/router.rs
KoHcoJlb: I envisioned IntoResponseLayer as a layer that only converts Output using IntoResponse trait, independent of the Router. Hence the naming (IntoResponse), but maybe it can be confusing and should be renamed.

I've also thought about making specialized layer for Router which would convert both output and error, to have a cleaner type, but I think type alias works just fine for that.

rama-http/src/service/web/error.rs
KoHcoJlb: Limiting factor here is that we only have `&Error` (shared ref) when downcasting.
I introduced ResponseError mainly for extractors and router to be able to propagate their desired Response, for example HTTP 400 for extractors and 404 for router.

rama-http/src/service/web/router.rs
KoHcoJlb: > Some simple change could be:
>   * request a IntoResponse layer (IntoResponseLayer).into_service(router)
>   * require handlers to return Result vs just single -> T, but this might have a bigger impact then we initially think of

I agree that most code should be able to switch to the new impl using these two changes.

---

Fixed `http_web_router.rs`
Though I don't see yet what to do with nesting routers, for now I just made it compile.

It feels somewhat wrong to me that we give user the Router and not the other way around. I understand that we need to somehow pass state, but still, I wonder if there is a better solution.
`axum` has really great implementation of state in my opinion, there, instead of passing it during creation of the `Router`, you set it at the end, declaring it with type system in the meantime. [docs](https://docs.rs/axum/latest/axum/struct.Router.html#method.with_state)

rama-http/src/service/web/router.rs
KoHcoJlb: Sorry about that
I've specifically made these formatting changes in a separate commit hoping it won't cause too much trouble.

rama-http/src/service/web/router.rs
soundofspace: I don't have a strong opinion on how we now create nested routers. We did it since it made sense at that time because a subrouter had identical state/generics. Now those generics indeed change that completely, can you check if defining the state similar to axum fixes this (I assume yes), which should unblock you on making this work for all common use cases?

examples/http_web_router.rs
```diff
@@ -80,15 +83,15 @@ async fn main() {
...
 
     let router = Router::new()
-        .with_get("/", Html(r##"<h1>Rama - Web Router</h1>"##.to_owned()))
+        .with_get("/", Ok(Html(r##"<h1>Rama - Web Router</h1>"##.to_owned())))
         // route with a parameter
         .with_post(
             "/greet/{name}",
             async |method: Method, Path(PostGreetForPathParams { name }): Path<PostGreetForPathParams>| {
-                Json(json!({
+                Ok::<_, BoxError>(Json(json!({
```
soundofspace: I wonder if by setting a default like Infallible on the trait if we can work around this

rama-http/src/service/web/router.rs
KoHcoJlb: I thought about it more and I believe I know how to make it decent with closures, I'll push it later today.

---
About State its just my preference for now, that I like how it's done in axum.
I also don't have a strong conviction whether it should be implemented here, and it will require some more core changes:
- the fact that state is set _after_ the router is configured would mean that it can't be cloned into endpoint during registration, and would need to be passed later, probably by creating wrapper type for input which contains both Request and State.

If you're interested we can explore that, or if not probably just skip it.

examples/http_web_router.rs
KoHcoJlb: In this case it can't be Infallible because extractor errors must be convertible into it, but I can check with BoxError if you'd like.

rama-http/src/service/web/router.rs
KoHcoJlb: Also, maybe you have some more complex example in mind for me to implement, so we can see more edge cases.

soundofspace: I'm definitely interested in exploring that, so if you see other ways how we can improve that, please share. And no I don't really have a more complex example for you to play with, but if it works for that example (including a subrouter with same state), we are in pretty good position and it should support most use cases.

examples/http_web_router.rs
soundofspace: BoxError which gets converted later to 500 seems like a solid default if that works, but yea not sure if it will be possible at all without sacrificing other things

rama-core/src/layer/map_err.rs
```diff
@@ -144,3 +146,59 @@ where
         MapErr { f: self.f, inner }
     }
 }
+
+/// [`Service`] which converts errors using [`Into`] trait
+#[derive(Debug, Clone)]
+pub struct IntoErr<S, E> {
```
GlenDC: @soundofspace is this one strictly needed? So far this is done using MapErr(E::from), is there something limiting here? MapErr also has some specialized constructors already

rama-core/src/service/svc.rs
```diff
@@ -306,6 +306,34 @@ where
     }
 }
 
+/// A static [`Service`] that always returns pre-defined output.
+#[derive(Debug, Clone)]
+pub struct StaticService<O>(O);
```
GlenDC: Name is ambigious, as it is really about StaticOutput. 

- why is this needed btw?
- we already have something similar around into_response logic, can that not be reused, or is there something special about this?

Cargo.toml
GlenDC: Not allowed. I prefer manal implementations for these.

if there ever is soemthing in derive_more that we do want, it should be part of rama-macros... I do not trust or wish to use this dependency. For several reasons, but all out of scope for this discussion.

examples/http_web_router.rs
GlenDC: Worst case this direction might be acceptable, but I would really prefer if we can support non-result outputs as well as Result outputs, as there are plenty cases where you are truly infallible and so having to now return a result always is a bit silly. Fine that it in background gets converted to an always-ok result, but ideally we do not place this burden on the user.

Worst case, we do, but than I would like to see a small report on what has been tried, and why these didn't work. @soundofspace please assist @KoHcoJlb on this specific thing where possible.

rama-core/src/service/svc.rs
KoHcoJlb: No, it's precisely nothing special, just a service which always returns the same output.
Yes, its predecessor lived near endpoints, but now, given we're going generic, it can be a general purpose service, analogous to `RejectService`.

rama-core/src/layer/map_err.rs
KoHcoJlb: For myself I can say that this separate service, not MapErr, is needed to have a "static" typed layer that we can use in default Router definition [here](https://github.com/plabayo/rama/pull/878/changes#diff-f975f345178ac8f7226e02c839cb36073953ea8b940f15bd14b7760ed01dcb16R37).
As far as I know it's not possible to express a type of fn with generic arguments manually.

rama-core/src/service/svc.rs
KoHcoJlb: Renamed it to `StaticOutput`

examples/http_web_router.rs
KoHcoJlb: So, my best and so far the only shot here is to keep implementations for types implementing `IntoResponse`. It's a bit hacky because `IntoResponse` here is used only as a marker trait, but it should cover everything that worked before and for anything else we can always resort to using Result.
Implemented in [bf035e8](https://github.com/plabayo/rama/pull/878/changes/bf035e89cf1a5fc5633d8b2e69077c0ab1e6658e)

`negative_impls` seems to be the feature we'd want here, but it's unstable.
Otherwise I don't see how we can disambiguate between impls for `O` and `Result<O, E>`

soundofspace: Yea negative impl's will be huge if they ever fully stabilizing, but until then you can get quite far by being creative. I'm currently really busy but I'm going to review/test this over the weekend and then come back to you

rama-core/src/layer/map_err.rs
soundofspace: Came to the same conclusion here, that this would be needed for this use case. But lets first see where we settle on and then figure out details like this

soundofspace:
So i went over this again, and basically this PR is doing this:
- Extend endpoint_fn to support any Error / result (and handle infallible)
- Add endpoint_layers to router so each endpoint can be converted to router types
- Make router generic over Output and Error -> This is a pretty big change, and has side-effects in a lot of places
- Add dedicated RouterError for router failures

I like in which direction this is going, but this is a lot at once, and maybe we can achieve this in a simpler way. Currently I'm thinking this
- Make router signature: `Router<State = (), Layer = DefaultLayer>` with `Service<Request, Output = Response, Error = Infallible>`
- Pass RouterError into endpoint_layer (maybe we call this into_response_layer) -> this gives you all the extra logic to handle these logging, custom conversion... without this leaking everywhere

By doing this I think we can support all use cases listed so far (logging/modifying...), without a pretty invasive change. Also as far I see, if we do this first and later still need generic Output/Error it should be possible to add that on top of this. What do you think about this @KoHcoJlb  Am I missing use cases that this would not support?

KoHcoJlb:
I'll need some time to think about it.
To clarify, is your only concern (design wise) the need to stack additional layer on top of Router to handle errors, or is there anything else?

soundofspace:
Some of the concerns I have
- Layer on top router for which I currently see not benefit (happy to see otherwise)
- If router returns BoxError and we convert to response we need to downcast. It's especially weird around handlers, like do we use outer transform (probably not), but does the inner stuff always return Response or also error.
- Requiring this `From<RouterError> for MyCustomErr` for custom errors feels wrong. It should probably be `RouterError<HandlerError>` but again this adds indirect for only the case of handle router specific errors which could be solved by passing them to the same into_response layer

I'm definitely not opposed to what you are doing, but it if feels like this is solving 3 problems at once, of which the last one: returning generic Output/Error has currently no real use case that could not be solved by the easier solution (that I can think of). But it does heavily complicate the design and opens questions like should this be handled in into_response_layer or outer layers (but maybe these are valid). But in general I would love to see more practical use cases before committing to this fully, especially the complex interactions between RouterError/BoxError/HandlerError

KoHcoJlb:
First I'd like to explain where I stand with this PR. I think it's not about solving any particular set of cases, but giving the flexibility for endpoint functions (with extractors) and router to be used in different scenarios. So in my opinion it solves one problem, which is relaxing the bounds on output types to allow them to be used in a more modular.
If you agree then I'd propose we reverse this a bit, trying to smooth out the rough edges while keeping the core flexibility in place.
This is a lot, yes, but right now as I see it we are one step away from keeping all this user-friendly and almost backward compatible, sans the need to slap a layer on top of router to handle errors.

---
#### Now, to answer some of your questions/concerns:
- #### About generic router
  - For this one, as a matter of fact, I have a practical use right now (didn't have it in mind when I started this PR though).
    I'm building a programmable mitm proxy to analyze some restful api traffic. The idea is to pass intercepted requests through the router, adding handler functions for paths that are of interest to me which will decode them into my application's types. Be it a enum with all events or a trait with downcasting I'm not yet sure. And router would then return this decoded value as its output to be used elsewhere in the business logic of the program.
    Router here is still routing requests, but instead of returning response for the client, it returns data for my application.
  - Another thing is nested routers. As you said all this new stuff can be configured on a router basis, but when we have multiple of them nested it means duplicating configuration instead of stacking it once at the top. And while current implementation of `with_sub_router_make_fn` clones the layer into the child router making it automatic, I think it can get a bit awkward if one would want to customize layer for the nested router.

- > * If router returns BoxError and we convert to response we need to downcast. It's especially weird around handlers, like do we use outer transform (probably not), but does the inner stuff always return Response or also error.
  
  I think we can just make default endpoint layer convert errors to Responses, to stay consistent with previous behavior. And then if someone wants to bubble errors to the top to do some global processing, they can later use downcasting to convert it into response.

- > * Requiring this `From<RouterError> for MyCustomErr` for custom errors feels wrong. It should probably be `RouterError<HandlerError>` but again this adds indirect for only the case of handle router specific errors which could be solved by passing them to the same into_response layer

  MyCustomErr is there to demonstrate the possibility of using custom error types, where in practice it usually would be just BoxError or another container like eyre. If someone wants to make their own error type yes, they can do so implementing From's for all possible errors they encounter, but I believe this is in line with Rust's error handling in general. 

  In RouterError itself I see no particular use, aside from its NotFound variant.
  By the way how custom not_found handler works with nested routers?
  Seems to me that child router will return plain 404 even if parent has not_found handler set. RouterError can help with that by handling it globally in some layer.

---
**To sum it up** yes, leaving the Router with `Response, Infallible` will remove some of these questions, but also some foreseeable functionality along with them.

And thank you for working with me on this @soundofspace, I really appreciate it!

soundofspace:
Ok I can behind those arguments

> MyCustomErr is there to demonstrate the possibility of using custom error types, where in practice it usually would be just BoxError or another container like eyre. If someone wants to make their own error type yes, they can do so implementing From's for all possible errors they encounter, but I believe this is in line with Rust's error handling in general.


Yes it is default rust bahaviour, but my point there, is mostly that Router should not return any Error, but it should return `RouterError<EndpointError>`, where only EndpointError is generic. This allows use to work around most of the issues I noticed with this. Eg in the default where it returns BoxError it would convert RouterError into a BoxError which ErrorHandlerLayer would then wrongfully always convert in 500 vs detailed codes such as method not allowed. But if we always return RouterError the default behaviour stays correct, and it also becomes possible to do all the custom error handling without always forcing special logic / downcasting. You could still do all of that but it would not be needed. 

When we commit to this router can just return Output and RouterError<EndpointError> without any extra logic to convert errors `From<RouterError>`, which should not be needed layer is already there to do conversion, this just adds an extra place to do conversions, which adds complexity for no real benefit. If you dont want RouterError you can convert it in a layer on top of router, but router should not have to do that. As a huge bonus, except for adding ErrorHandlerLayer, this would probably be a drop in replacement.

So basically I would commit to these signatures:
```rust
// IntoResponseLayer converts to Result<Response, Infallible> similar to what we had now (can be multiple layers)

pub struct Router<State = (), Layer = IntoResponseLayer, O = Response, EndpointError = Infallible> {
    routes: MatchitRouter<Vec<(HttpMatcher<Body>, BoxService<Request, O, E>)>>,
    sub_services: Option<Trie<String, SubService<O, E>>>,
    not_found: Option<BoxService<Request, O, E>>,
    layer: Layer,
    state: State,
}

impl<State, L, O, EndpointError> Service<Request> for Router<State, L, O, EndpointError>
where
    O: Send + 'static,
    EndpointError: Send + 'static,
    L: Send + Sync + 'static,
    State: Send + Sync + Clone + 'static,
{
    type Output = O;
    type Error = RouterError<EndpointError>;
```

KoHcoJlb:
Ah, I see. I envisioned from the start that layer on top of Router would use downcasting to convert errors to responses, so there wouldn't be wrongful 500s, that's the whole reason behind `ResponseError`.

But I understand your idea now, it would allow to skip downcasting altogether for default router (inner layer converting handler errors directly using `IntoResponse`, and outer layer converting `RouterError` the same way).
I'm okay with it, though I've encountered several complications while implementing it:
- Keeping it inside a single type would require all Es to implement `IntoResponse` to make `ResponseError` work, so I've split it into two types.
- If EndpointError is a container like `BoxError`, which doesn't implement `Error` itself, it somewhat breaks nesting of the routers (you can see it at [`router.rs:1172`](https://github.com/KoHcoJlb/rama/blob/86149047c58ebec8c0268d13e088b025409672b1/rama-http/src/service/web/router.rs#L1172)) and also complicates downcasting ([link](https://github.com/KoHcoJlb/rama/blob/86149047c58ebec8c0268d13e088b025409672b1/examples/test_new_router.rs#L115-L119), does not compile).
  I made it work with `with_sub_router_make_fn` by passing `RouterError` through instead of converting it, but it kinda again makes `Router` a special case service where you can't easily use it in a service stack which uses BoxErrors.

I will think some more how to make it work, but in the meantime I pushed what I have.
Can you look it over and say if it aligns with what you have in mind?

soundofspace:
> Keeping it inside a single type would require all Es to implement IntoResponse to make ResponseError work, so I've split it into two types.

Can you explain this one a bit more, I'm current failing to see what this is needed. IntoResponse is already a conditional implementation?

> If EndpointError is a container like BoxError, which doesn't implement Error itself, it somewhat breaks nesting of the routers (you can see it at [router.rs:1172](https://github.com/KoHcoJlb/rama/blob/86149047c58ebec8c0268d13e088b025409672b1/rama-http/src/service/web/router.rs#L1172))

This specific example is expected behaviour. If you use a sub-service, any service at all (including a router) you need to make sure output/error matches what the router expects. Here this could for example be (but you could also map output/errors completely different as long as the types match):
```
let app = Router::new()
    .with_sub_service(
        format!("{prefix}api"),
        ErrorHandlerLayer::default().into_layer(api_router),
    )
```
With_sub_service is really meant for: I have a more special use case for which i need a sub-service, I know what I'm doing so give me the option to go crazy. If you just need a subrouter there is an easier way for that as you have already discovered. Part of this work could also be making it easier to nest router with eg different state, or this could also be followup work. Any work that improves router ergonomics can be discussed/explored.

> and also complicates downcasting

Is this because now you would need to downcast to `RouterError<WhateverThisIs>`?

> but it kinda again makes Router a special case service where you can't easily use it in a service stack which uses

Not really true, most services have an `Error: Into<BoxError>` trait bound, so this error would also work for that. And for cases where you do need this we have this already:
```
MapErrLayer::into_box_error(),
```

KoHcoJlb:
> Is this because now you would need to downcast to `RouterError<WhateverThisIs>`?

> Not really true, most services have an `Error: Into<BoxError>` trait bound, so this error would also work for that. And for cases where you do need this we have this already:
> 
> ```
> MapErrLayer::into_box_error(),
> ```

So, regarding these two, maybe I'm missing something obvious, but `RouterError<BoxError>` can't implement `Error` because `BoxError` itself does not implement `Error`.
For this reason:
- `RouterError<BoxError>` does not implement `Into<BoxError>`
- `RouterError::<BoxError>::downcast_ref` method is missing, as it is defined in `Error` trait

KoHcoJlb:
> Can you explain this one a bit more, I'm current failing to see what this is needed. IntoResponse is already a conditional implementation?

Here's the code implementing it as a single enum. Interesting bit is implementation of `Error::source`.
The `Endpoint` variant is returned as is and for everything else we need to return `ResponseError`, but to do that self must implement `IntoResponse`.
As for why I implemented `ResponseError` that way in [452c9ad](https://github.com/plabayo/rama/pull/878/changes/452c9ad33a7154ec47df68ffe2a5fa932541b376) is to keep downcasting logic inside any particular error type.
Before it was a simple wrapper type which must've been used at every point where such errors are returned, and aside from inconveniences it brings, it's also easy to forget to wrap an error at some point, silently loosing that functionality.

```rust
#[derive(Debug, Clone)]
pub enum RouterError<E> {
    Internal,
    MethodNotAllowed(NonEmptySmallVec<7, Method>),
    NotFound,
    Endpoint(E),
}

impl<E> Error for RouterError<E>
where
    E: Error + 'static,
    E: IntoResponse + Clone + Send + Sync,
{
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(match self {
            RouterError::Endpoint(err) => err,
            _ => ResponseError::new(self),
        })
    }
}

impl<E: fmt::Debug> fmt::Display for RouterError<E> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        fmt::Debug::fmt(self, f)
    }
}

impl<E: IntoResponse> IntoResponse for RouterError<E> {
    fn into_response(self) -> Response {
        match self {
            Self::Internal => StatusCode::INTERNAL_SERVER_ERROR.into_response(),
            Self::MethodNotAllowed(allowed) => (
                Headers::single(Allow(allowed)),
                StatusCode::METHOD_NOT_ALLOWED,
            )
                .into_response(),
            Self::NotFound => StatusCode::NOT_FOUND.into_response(),
            Self::Endpoint(err) => err.into_response(),
        }
    }
}
```

KoHcoJlb:
> This specific example is expected behaviour. If you use a sub-service, any service at all (including a router) you need to make sure output/error matches what the router expects. Here this could for example be (but you could also map output/errors completely different as long as the types match):
> 
> ```
> let app = Router::new()
>     .with_sub_service(
>         format!("{prefix}api"),
>         ErrorHandlerLayer::default().into_layer(api_router),
>     )
> ```
> 
> With_sub_service is really meant for: I have a more special use case for which i need a sub-service, I know what I'm doing so give me the option to go crazy. If you just need a subrouter there is an easier way for that as you have already discovered.

In fact it's just a consequence of inability to convert `RouterError<BoxError>` into `BoxError`, otherwise it would've worked because currently endpoint_layer is also applied to sub-services and default layer is `IntoErr<BoxError>`.

> I have a more special use case for which i need a sub-service, I know what I'm doing so give me the option to go crazy.

But that's a good point, I've been thinking myself about whether endpoint_layer should be applied to sub-services, and probably the answer is no.

> Part of this work could also be making it easier to nest router with eg different state, or this could also be followup work. Any work that improves router ergonomics can be discussed/explored.

Yeah, I've already implemented proposal regarding state management but it'll be in a separate PR, here we have enough as it is.

KoHcoJlb:
To summarize, the main problem with `RouterError<EndpointError>` arises when EndpointError is an "error container" like `BoxError` or `eyre::Report`. They don't implement the `Error` trait and thus `RouterError<Container>` itself can't be converted into them. Consider example implementation of `Error` for `RouterError<EndpointError>` below

```rust
impl<E> Error for RouterError<E>
where
    E: Error + 'static, // this bound is not implemented for BoxError
{
    fn source(&self) -> Option<&(dyn Error + 'static)> {
        Some(match self {
            RouterError::Endpoint(err) => err,
            _ => ...,
        })
    }
}
```

This makes RouterError kind of a terminal error type which can't be easily wrapped when Router is not the last service in the stack (last one with errors).

---

To work around that we basically need two implementations, `E: Error + 'static` for simple errors and `E: AsRef<dyn Error + Send + Sync>` for BoxError and such, but it requires an additional type parameter to avoid impl conflicts. You can see an example of that in implementation of `DowncastResponseLayer`, which is also affected by it, though to a lesser degree by it being consumer of these errors.

I wasn't been able to come up with something to save `RouterError<EndpointError>`. Closest I've got is the same solution as for `DowncastResponseLayer`, but that means dragging that additional `M` parameter up to the `Router` signature.
I'm ready to work on it some more if you have some ideas/direction, just right now I don't think that adding that additional `M` to `RouterError` and `Router` is an optimal choice.

> Eg in the default where it returns BoxError it would convert RouterError into a BoxError which ErrorHandlerLayer would then wrongfully always convert in 500 vs detailed codes such as method not allowed. But if we always return RouterError the default behaviour stays correct, and it also becomes possible to do all the custom error handling without always forcing special logic / downcasting.

But considering these two issues you outlined I have another suggestion.
As I've said before and implemented now, for downcasting approach there is `DowncastResponseLayer` which serves the same function as `ErrorHandlerLayer` but using downcasting to convert errors into responses. So there wouldn't be erroneous 500s.

And for default router, to avoid any downcasting at all, you can see what I did in [e544057](https://github.com/plabayo/rama/pull/878/changes/e544057ae58e4f5ebfffa74552fff5838cd89e3e).
Basically it now returns plain `RouterError` (without any endpoint component), you can think of it as EndpointError being Infallible. It is facilitated by a new default layer which eagerly converts both Output and Error of endpoints using IntoResponse. It also deals with extractor errors using `ErrorResponse` struct.
It may also be a point to remember, that aside from RouterError we also have custom responses in extractor errors.

What do you think about it and perhaps what are the other issues you have about errors?
I'm awaiting your comments as to further direction where this should go @soundofspace

soundofspace:
> RouterError<BoxError> into BoxError

I completely overlooked that the last time, it's a very rough edge in Rust Error handling, and I was wrong with what I suggested, sorry for that.

> What do you think about it and perhaps what are the other issues you have about errors

This new approach looks really solid, and at first sight it seems to solve all the things I was wary about, nice work! It seems the architecture is now pretty much figured out, so for a next stage I propose that you update all code so that everything fully compiles. That should show us if this works for everything, or if there are still some details that need to be figured out. Once that is finished we can go over to a full review and focus on all the smaller details.

rama-http/src/service/web/mod.rs
```diff
@@ -1,20 +1,22 @@
 //! basic web service
 
-mod service;
-#[doc(inline)]
-pub use service::{WebService, match_service};
+// mod service;
+// #[doc(inline)]
+// pub use service::{WebService, match_service};
```
KoHcoJlb: I don't quite understand it's (WebService's) relation to the Router, is it supposed to be a more simple/robust implementation of it?
I'm wondering if it should support all the new features or just remain simple as it is now.

soundofspace: It's an easier version of Router, so this can remain simple

rama-http/src/service/web/router.rs
```diff
@@ -270,157 +318,172 @@ where
...
+    //     dir: include_dir::Dir<'static>,
+    //     mode: DirectoryServeMode,
+    // ) -> &mut Self {
+    //     let service = ServeDir::new_embedded(dir).with_directory_serve_mode(mode);
+    //     match self.not_found.clone() {
+    //         Some(not_found) => self.set_sub_service(prefix, service.fallback(not_found)),
+    //         None => self.set_sub_service(prefix, service),
+    //     }
+    // }
+    //
```
KoHcoJlb: Do we need all these ServeDir specializations on the Router?
The only reason I see is to pass not_found handler through, and I think now it can be implemented using errors, so users will configure ServeDir themselves and attach it using with_sub_service.

soundofspace: I had this discussion with glen before and we were indeed planning to make the router slimmer. Is this something that is needed for this PR to work? Just thinking if we can keep the scope of this PR small if possible, and postpone things like this as followup PRs

KoHcoJlb: > Is this something that is needed for this PR to work?

No, I think not, will leave it for later then.

KoHcoJlb:
Can we run CI suite on this PR?

soundofspace:
Could you merge main in your branch first and fix merge conflicts. Also if you want you can run these tests locally with `just qa` already

KoHcoJlb:
Ready for review @soundofspace

examples/ws_chat_server.rs
```diff
@@ -62,10 +64,10 @@ async fn main() {
     let graceful = rama::graceful::Shutdown::default();
 
     graceful.spawn_task_fn(async |guard| {
-        let server = HttpServer::new_http1(Executor::graceful(guard.clone())).service(Arc::new(Router::new().with_get("/", Html(INDEX)).with_get(
+        let router = Router::new().with_get("/", Html(INDEX)).with_get(
             "/chat",
             WebSocketAcceptor::new().into_service(service_fn(
-                async |mut ws: ServerWebSocket| {
+                async |mut ws: ServerWebSocket| -> Result<_, Infallible> {
```
soundofspace, line 70: Is this one needed, or left-over from testing? Not a blocker, just curious

rama-cli/src/cmd/serve/httptest/endpoint/method.rs
```diff
@@ -1,9 +1,9 @@
-use rama::http::{Method, StatusCode, service::web::response::IntoResponse};
+use rama::http::{Method, Response, StatusCode, service::web::response::IntoResponse};
 
-pub(in crate::cmd::serve::httptest) async fn handler(method: Method) -> impl IntoResponse {
+pub(in crate::cmd::serve::httptest) async fn handler(method: Method) -> Response {
```
soundofspace:
```suggestion
pub(in crate::cmd::serve::httptest) async fn handler(method: Method) -> Result<Response, ErrorResponse> {
```
Lets return a result here

rama-core/src/layer/map_err.rs
soundofspace: Ok, so we technically don't really need this. It is possible to do this everywhere with MapError (I think). But I would still keep this one for ergonomics, since it allows us to express into error much easier (without casting to a function). @GlenDC what do you think (I know there are easier/better ways to express this specific function, but there are places where we do need to specify MapErr bounds)

```rust
fn map() -> MapErr<impl Service<(), Output = (), Error = BoxError>, fn(BoxError) -> ErrorResponse> {
    let svc = service_fn(async |_| Ok::<_, BoxError>(()));

    MapErrLayer::new(Into::into as fn(BoxError) -> ErrorResponse).into_layer(svc)
}

fn into() -> IntoErr<impl Service<(), Output = (), Error = BoxError>, ErrorResponse> {
    let svc = service_fn(async |_| Ok::<_, BoxError>(()));

    IntoErrLayer::new().into_layer(svc)
}

```

examples/test_new_router.rs
```diff
@@ -0,0 +1,172 @@
+#![allow(unused_variables, dead_code, unused_must_use, clippy::restriction)]
```
soundofspace: Can you rename this example to something like advanced_router or custom_router and really showcase what you can do with this new router. In the issue/pr some use cases were mentioned, having some of those in an example would really showcase this new logic, and would make it discoverable for other people.

rama-http/src/service/web/service.rs
```diff
@@ -411,11 +435,12 @@ where
...
     where
         I: IntoEndpointServiceWithState<T, State>,
+        I::Service: Service<Request, Output: IntoResponse, Error: Into<ErrorResponse>>,
     {
         let endpoint = Endpoint {
             matcher,
-            service: service
-                .into_endpoint_service_with_state(self.state.clone())
+            service: DefaultLayer::default()
+                .layer(service.into_endpoint_service_with_state(self.state.clone()))
```
soundofspace:
```suggestion
                .into_layer(service.into_endpoint_service_with_state(self.state.clone()))
```

rama-http/src/service/web/service.rs
```diff
@@ -426,11 +451,12 @@ where
...
```
soundofspace:
```suggestion
                .into_layer(service.into_endpoint_service_with_state(self.state.clone()))
```

rama-http/src/service/web/service.rs
```diff
@@ -442,9 +468,10 @@ where
     pub fn with_not_found<I, T>(mut self, service: I) -> Self
     where
         I: IntoEndpointServiceWithState<T, State>,
+        I::Service: Service<Request, Output: IntoResponse, Error: Into<ErrorResponse>>,
     {
-        self.not_found = service
-            .into_endpoint_service_with_state(self.state.clone())
+        self.not_found = DefaultLayer::default()
+            .layer(service.into_endpoint_service_with_state(self.state.clone()))
```
soundofspace:
```suggestion
            .into_layer(service.into_endpoint_service_with_state(self.state.clone()))
```

rama-http/src/service/web/service.rs
```diff
@@ -453,9 +480,10 @@ where
     pub fn set_not_found<I, T>(&mut self, service: I) -> &mut Self
     where
         I: IntoEndpointServiceWithState<T, State>,
+        I::Service: Service<Request, Output: IntoResponse, Error: Into<ErrorResponse>>,
     {
-        self.not_found = service
-            .into_endpoint_service_with_state(self.state.clone())
+        self.not_found = DefaultLayer::default()
+            .layer(service.into_endpoint_service_with_state(self.state.clone()))
```
soundofspace:
```suggestion
            .into_layer(service.into_endpoint_service_with_state(self.state.clone()))
```

rama-http/src/service/web/service.rs
```diff
@@ -601,8 +629,12 @@ where
 macro_rules! __match_service {
     ($($M:expr_2021 => $S:expr_2021),+, _ => $F:expr $(,)?) => {{
         use $crate::service::web::IntoEndpointService;
+        use $crate::layer::error_handling::ErrorHandler;
         use $crate::__macro_dep::__core::matcher::MatcherRouter;
-        MatcherRouter(($(($M, $S.into_endpoint_service())),+, $F.into_endpoint_service()))
+        MatcherRouter((
+            $(($M, ErrorHandler::new($S.into_endpoint_service()))),+,
+            ErrorHandler::new($F.into_endpoint_service()))
+        )
```
soundofspace, line 637: Here I think you can move ErrorHandler to the outside, so MatcherRouter stays as is, and you would use it like `ErrorHandler::new(MatcherRouter())`

soundofspace: Forcing ErrorHandler here feels wrong, but yea maybe it might be needed?

scripts/extra-checks.sh
```diff
@@ -7,6 +7,8 @@ exit_code=0
 # Make sure all examples are included in the rama book
 for example in $(cd $SCRIPT_DIR/.. && find examples -maxdepth 1 -type f -name '*.rs' -not -name 'mod.rs'); do
     echo "Checking $example..."
+    # will remove it before merge along with test_new_router.rs
+    if [[ $example == "examples/test_new_router.rs" ]]; then continue; fi
```
soundofspace: leaving this here as reminder

Cargo.toml
```diff
@@ -914,3 +914,8 @@ required-features = ["http-full", "boring"]
 [[example]]
 name = "ws_over_h2"
 required-features = ["http-full", "boring"]
+
+# will remove before merge along with test_new_router.rs
```
soundofspace: leaving this here as reminder

rama-http/src/layer/error_handling.rs
```diff
@@ -155,3 +160,100 @@ where
...
+#[derive(Default, Clone, Copy)]
+#[non_exhaustive]
+pub struct AsRefKind;
+
+/// [`Service`] that tries to downcast an Error into [`Response`] using [`DowncastResponseError`]
+///
+/// If there is no [`DowncastResponseError`] in the error chain, it returns INTERNAL_SERVER_ERROR
+pub struct DowncastErrorHandler<S, K> {
+    inner: S,
+    _kind: K,
```
soundofspace: Phantomdata for this shows intent better, and will allow some simplifications like dropping copy bound, and m: default bound)

rama-http/src/layer/error_handling.rs
```diff
@@ -155,3 +160,100 @@ where
...
+            IntoResponse::into_response,
+        ))
+    }
+}
+
+/// [`Layer`] that tries to downcast an Error into [`Response`] using [`DowncastResponseError`]
+///
+/// See [`DowncastErrorHandler`] for additional documentation
+#[derive(Debug, Default, Clone)]
+pub struct DowncastErrorHandlerLayer<M>(M);
```
soundofspace: also phantomdata

rama-http/src/layer/error_handling.rs
```diff
@@ -155,3 +160,100 @@ where
...
+            |err| {
+                if let Some(resp) = DowncastResponseError::try_as_response(&err) {
+                    resp
+                } else {
+                    StatusCode::INTERNAL_SERVER_ERROR.into_response()
+                }
+            },
+            IntoResponse::into_response,
+        ))
+    }
```
soundofspace: NIT, but i would argue this is way easier to read
```rust
    async fn serve(&self, input: I) -> Result<Self::Output, Self::Error> {
        Ok(match self.inner.serve(input).await {
            Ok(resp) => resp.into_response(),
            Err(err) => DowncastResponseError::try_as_response(&err)
                .unwrap_or_else(|| StatusCode::INTERNAL_SERVER_ERROR.into_response()),
        })
    }
```

rama-http/src/layer/error_handling.rs
```diff
@@ -155,3 +160,100 @@ where
...
+    async fn serve(&self, input: I) -> Result<Self::Output, Self::Error> {
+        Ok(self.inner.serve(input).await.map_or_else(
+            |err| {
+                if let Some(resp) = DowncastResponseError::try_as_response(err.as_ref()) {
+                    resp
+                } else {
+                    StatusCode::INTERNAL_SERVER_ERROR.into_response()
+                }
+            },
+            IntoResponse::into_response,
```
soundofspace: same remark here

rama-http/src/service/web/error.rs
```diff
@@ -0,0 +1,75 @@
+/*!
```
soundofspace: @GlenDC for me this syntax works, but I know in other places we use //!, any opinion on this?

rama-http/src/service/web/router.rs
```diff
@@ -35,39 +25,68 @@ use rama_utils::{
...
+            IntoEndpointService, IntoEndpointServiceWithState,
+            error::DowncastResponseError,
+            response::{ErrorResponse, Headers, IntoResponse},
+        },
+    },
+};
+
+#[derive(Debug, Clone, Default)]
+#[non_exhaustive]
+pub struct DefaultLayer;
```
soundofspace: Small doc comment that explain what this layer does so people are not surprised when they see this

rama-http/src/service/web/router.rs
```diff
@@ -563,33 +558,201 @@ where
...
         Self::new()
     }
 }
 
-struct SubService {
-    svc: BoxService<Request, Response, Infallible>,
+#[derive(Debug, Clone)]
+pub enum RouterError {
+    Internal,
+    MethodNotAllowed(Box<NonEmptySmallVec<7, Method>>),
```
soundofspace, line 602: I guess boxing is needed here to limit error size?

rama-http/src/service/web/error.rs
```diff
@@ -0,0 +1,75 @@
...
+where
+    T: IntoResponse + Clone + Error + Send + Sync + 'static,
+{
+    fn as_response(&self) -> Response {
+        self.clone().into_response()
+    }
+}
+
+#[derive(Debug)]
+pub struct DowncastResponseError(fn(&(dyn Error + 'static)) -> Option<Response>);
```
soundofspace: This for sure deserve a huge doc comment explaining: what it is, what it does, why it works like this...

So the way I'm seeing this now, this is basically "creative" way of implement something like [Error::provide](https://doc.rust-lang.org/std/error/trait.Error.html#method.provide) on stable rust?

@GlenDC what do you think about this? Seems like the individual pieces of this trick are quite common, but combining them like this is a first for me. To me this feels like a pragmatic solution to work around stable rust limitations, be it a very creative one (maybe even hacky since we abuse source). I tested for a bit what other options we have here, but they all require manual work by users that produce errors, which is kind of painful. Curious what you think here, or if you have better other ideas?

soundofspace (changes requested):
@KoHcoJlb this is amazing work! External usage stays almost identical yet it allows for some really advanced use cases, nice.

Only some small things left to do, and this should be ready to merge

rama-http/src/service/web/error.rs
GlenDC: I prefer we use `//!` everywhere, more consistent

rama-http/src/service/web/router.rs
KoHcoJlb: Yep, to avoid various lint-errors related to it.

rama-http/src/service/web/service.rs
KoHcoJlb: It's the same thing as with layer in Router, it allows arms to have different return types.

examples/ws_chat_server.rs
KoHcoJlb: It is needed because no Err variants are returned, so the compiler can't guess the type of Error.

rama-http/src/layer/error_handling.rs
KoHcoJlb: Agreed

rama-http/src/service/web/error.rs
KoHcoJlb: Never read about `Error::provide` until now, because yes, it's nightly.
Seems like my implementation will be replaceable by it when/if it stabilizes.
With its current api it should even be possible to do `request.provide_ref(self as &dyn AsResponse)`, making it almost as direct as it can get.

KoHcoJlb: Documentation for this one and also advanced router example will take me some more time to write.
Though maybe you can suggest what else should be added to this module beyond its existing module doc?

rama-core/src/layer/map_err.rs
KoHcoJlb: There is one more catch here: rust likes to eagerly specialize functions' generics.
Consider this example from `test_new_router.rs`

```rust
let mut router = Router::new()
    // changed this from IntoErrLayer to MapErrLayer
    .with_endpoint_layer((IntoResponseLayer::new(), MapErrLayer::into_box_error()));
router.set_match_route("/test1", HttpMatcher::method_get(), test_func1);
router.set_match_route("/test2", HttpMatcher::method_get(), test_func2);

router.set_match_route("/test3", HttpMatcher::method_get(), TestService1);
router.set_match_route("/test4", HttpMatcher::method_get(), TestService2);

// up to this point everything compiles fine, because all endpoints have BoxError error type

router.set_match_route("/test5", HttpMatcher::method_get(), test_func5);

router.set_match_route("/test6", HttpMatcher::method_get(), Ok("test"));

// but these last two don't, because MapErrLayer is already specialized to BoxError type
// while test_func5 has MyCustomErr and Ok("test") is Infallible
```

examples/test_new_router.rs
KoHcoJlb: Added `http_advanced_router` example. It's rather simple at its core, but should showcase all new tools available.

KoHcoJlb:
@soundofspace can we resolve all other comments so we can focus on the two major ones left, as I see them: documentation of downcast error and router example?

GlenDC:
Overall this looks very promising @soundofspace and @KoHcoJlb. I do have a few requests, concerns, and questions:

* Please make sure to remove `test_new_router` and any related references once you start cleaning up the PR.
* Could you add a more detailed explanation/doc comment for `DowncastResponseError`, including explicit notes that we eventually want to replace this with `Error::provide` once that becomes possible? It would also be helpful to link to the relevant Rust tracker issue in the `rust-lang` org.
* Please audit the remaining places where we still use things like `ConsumeErrorLayer::trace_as_debug` for endpoints. Are these still required now? If so, could you document why? My naive expectation would be that the advanced router can now handle endpoint services that return errors directly.
* Please improve the doc comments for `DefaultLayer` and `with_endpoint_layer`. At first it was not immediately clear to me why this was required, though I think I understand it better now. That said, it still feels a bit footgun-ish in its current form, since modifying the layer between inserts can unexpectedly affect previously inserted endpoints. Is there a strong use case for that behavior? I wonder if it would make more sense to only allow this layer to be configured at router creation time, and require subrouters when different layers are needed.
* For the advanced router example:

  * Please add a clear high level comment explaining what the example demonstrates and how it is intended to be used.
  * `tracing_subscriber` is already available through Rama, so there is no need to import it separately.
  * Looking recursively through `source` feels fairly fragile to me. As far as I understand, there is no contract guaranteeing that this is reliable behavior. Originally we considered adding this as part of `rama-error`, but it became clear that `source` is really only intended for the direct source, and anything beyond that is more of a best effort assumption. That also makes me wonder a bit about some of the other assumptions around error handling here.

KoHcoJlb:
> That said, it still feels a bit footgun-ish in its current form, since modifying the layer between inserts can unexpectedly affect previously inserted endpoints.

It does not affect previous endpoints though, it's described in the comment. Maybe you meant to say that users would still assume that new endpoint layer will be applied both to the new and old endpoints?

> Is there a strong use case for that behavior? I wonder if it would make more sense to only allow this layer to be configured at router creation time, and require subrouters when different layers are needed.

In first revision it was Router::new_with_layer(_and_state) but then Brecht asked me to make it how it is now.
I didn't think about it at the time, but now I see a few use cases:
* Reset endpoint layer type to avoid writing it in a return type of a function, e.g. when nesting routers.
  Snippet from `http_advanced_router.rs`
  ```rust
  fn api_v2<L>(
      router: Router<Arc<AppState>, L, Response, BoxError>,
  ) -> Router<Arc<AppState>, (), Response, BoxError> {
      let state = router.state().clone();
  
      router
          .with_endpoint_layer((
              layer_fn(move |inner| SerializeResponse {
                  inner,
                  state: state.clone(),
              }),
              layer_fn(ApplicationErrorService),
              layer_fn(LogHandlerOutput),
          ))
          .with_get("/greet", greet_v2)
          .with_endpoint_layer(())
  }
  ``` 
* To allow applying a layer to a group of endpoints, for example:
  ```rust
  router.
    .with_get("/a", ...)
    .with_get("/b", ...)
    .with_endpoint_layer((AuthLayer, DefaultLayer))
    .with_get("/c", ...)
    .with_get("/d", ...)
  ```
  Unless subrouters can be registered at the same level, without subpath, not sure if it's possible.

> Looking recursively through `source` feels fairly fragile to me. As far as I understand, there is no contract guaranteeing that this is reliable behavior. Originally we considered adding this as part of `rama-error`, but it became clear that `source` is really only intended for the direct source, and anything beyond that is more of a best effort assumption. That also makes me wonder a bit about some of the other assumptions around error handling here.

I don't quite see which fragility do you mean.
There is [`Error::source`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.source) method which, while maybe not very strongly described in its doc, is used to return "parent" error if there is one, which can have its own "parent" and so on. As far as I've seen, it is the case in practice.
There is also unstable [`Error::sources`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.sources) method, which returns an iterator doing exactly that.

GlenDC:
> I don't quite see which fragility do you mean.

While developing rama-error, we originally had some utilities in there so you could easily do such things. However very quickly it was realized that not all errors return the parent correctly if at all. And some are legit reasons, yet breaking the cycle, as the definition of what is "source" is not universally agreed upon..

As a result if you rely on this, depending on your stack you might not get your error, yet how would you know.. very hard to debug issues. That's why we moved away from it as it is the kind of nasty surprises I tend to want to avoid with rama, to instead have as a predictable as possible result just from reading code and have confidence.

Also in your case I do not believe you really need this to make your goals work.

rama-http/src/service/web/endpoint/extract/mod.rs
```diff
@@ -85,7 +85,7 @@ pub trait FromPartsStateRefPair<State>: Sized + Send + Sync + 'static {
 pub trait FromRequest: Sized + Send + 'static {
     /// If the extractor fails it'll use this "rejection" type. A rejection is
     /// a kind of error that can be converted into a response.
-    type Rejection: IntoResponse;
+    type Rejection: Into<ErrorResponse>;
```
soundofspace: @GlenDC technically for advanced router we dont need any bounds here. So there is argument to remove this bound, or we could keep it to force that the "default" logic of into response always works, what do you think

rama-http/src/service/web/endpoint/response/mod.rs
```diff
@@ -58,6 +58,7 @@ pub mod redirect;
 pub use redirect::Redirect;
 
 pub mod sse;
+
```
soundofspace:
```suggestion
```

rama-http/src/service/web/router.rs
```diff
@@ -563,33 +567,197 @@ where
...
+        }
+    }
+}
+
+impl From<Infallible> for RouterError {
+    fn from(_: Infallible) -> Self {
+        // unreachable
+        Self::Internal
+    }
+}
```
soundofspace, line 632:
```suggestion
impl From<Infallible> for RouterError {
    fn from(infallible: Infallible) -> Self {
        match infallible {}
    }
}
```

soundofspace (changes requested):
@KoHcoJlb except for some very minor things, and a review from @GlenDC this is looking good.

I also pushed a commit that finishes the advanced_router example and that removes all error downcasting. @GlenDC and I discussed this offline in great depth and we both came to the conclusion that we do not want to support error downcasting as a core rama feature. We both believe that everything is possible in a typed way by handling errors close to where they are created eg like it happens in the advanced_router example.

Downcasting errors is still possible if you really want to use this, but just not something that we currently want to promote in rama itself.

I also want to say thanks for all of this work. I was really impressed with the advanced_router example and how everything nicely comes together there, it clearly shows that this change opens the door for some really interesting router stacks.

examples/ws_chat_server.rs
GlenDC: Might as well return a BoxError in that case so that we can remove the `Unwrap` from the `get_ref<State>` 🤷

KoHcoJlb:
Thanks @soundofspace!
Sorry for not finishing it myself, kinda lost momentum there at the end, documentation and examples are not my strong suite.
If there's some more work to be done here - feel free to ping me, I'll try my best to do it in a timely manner.

rama-core/src/layer/map_err.rs
```diff
@@ -144,3 +146,71 @@ where
...
```
GlenDC: let's move this into an into_err.rs file, to not mix with map_err

rama-grpc/src/service/router.rs
```diff
@@ -53,7 +53,7 @@ impl Service<Request> for GrpcRouter {
                 Status::unimplemented(arcstr!("service name not found in uri path"))
                     .try_into_http::<Body>()
                     .into_box_error()
-                    .into_response(),
+                    .unwrap_or_else(|e| e.into_response()),
```
GlenDC, line 56: What is the reason we can no longer support into_response on a Result? This reads pretty awkward...

rama-http/src/service/web/endpoint/extract/mod.rs
GlenDC: I am fine in trying to remove it, I will play with it myself later.

rama-http/src/service/web/endpoint/response/into_response.rs
```diff
@@ -80,19 +80,6 @@ impl IntoResponse for BoxError {
     }
 }
 
-impl<T, E> IntoResponse for Result<T, E>
```
GlenDC, line 83: why do we need to remove this? Feels pretty awkward...

rama-http/src/service/web/endpoint/response/mod.rs
```diff
@@ -144,18 +145,6 @@ pub use partial_updates::PartialUpdates;
 /// ```
 pub type Result<T, E = ErrorResponse> = std::result::Result<T, E>;
 
-impl<T> IntoResponse for Result<T>
```
GlenDC, line 147: ?

rama-http/src/service/web/endpoint/mod.rs
```diff
@@ -20,119 +25,83 @@ pub(crate) struct Endpoint {
 
 /// utility trait to accept multiple types as an endpoint service for [`super::WebService`]
 pub trait IntoEndpointService<T>: private::Sealed<T, ()> {
-    type Service: Service<Request, Output = Response, Error = Infallible>;
+    type Service: Service<Request>;
```
GlenDC, line 28: I can understand why no longer forcing the error, as that is one of core reasons this PR exists. But why also not force Response? I am fine in having a generic response body... but why would you not be able to assume it is a http response? It is after all a http rservice...

scripts/extra-checks.sh
```diff
@@ -7,6 +7,7 @@ exit_code=0
 # Make sure all examples are included in the rama book
 for example in $(cd $SCRIPT_DIR/.. && find examples -maxdepth 1 -type f -name '*.rs' -not -name 'mod.rs'); do
     echo "Checking $example..."
+
```
GlenDC:
```suggestion

```

GlenDC (approved):
Looks good to me, couple of. changes + want to try some minor things. Will finish this one of myself. Great work @KoHcoJlb and @soundofspace ! You managed to turn this into a true gem. Also welcome to the community @KoHcoJlb , I hope you'll stick around :) And if this is still not enough for your use case(s) do ping us on discord and we can see what to do more!

rama-http/src/service/web/endpoint/response/into_response.rs
KoHcoJlb: There's now two implementations of EndpointServiceFn, one for `Future<Output = Result<O, E>>` and one for `Future<Output = O> where O: IntoResponse`.
If Result itself implements `IntoResponse` it causes ambiguity conflicts.

rama-http/src/service/web/endpoint/mod.rs
KoHcoJlb: I'd argue that they go hand in hand (Output and Error) and there's plenty of potential cases to go around, like:
- Logging original output type, before it is obscured by `Response`
- Do some stateful conversions into `Response`, like shown in advanced router example
  Or avoid dealing with wrapper types to have custom IntoResponse impl where it can be implemented more efficiently/conveniently with an external layer. 
- Reuse this machinery outside of a typical "handle request to produce response" flow

  I believe there's great deal of logic inside `Router` and all the extractors that can be reused elsewhere, where the goal is not producing a `Response`.
  For example right now I have a use case where I need to extract various data from different api requests (in a mitm proxy) and I'd like to use both Router and extractors there. In this scenario endpoints will be "deserializing" requests into my custom types, which are returned from the router to later be handled by an application logic.

It is still about http because all the logic is operating on http `Request`s, but nothing requires that output must be an http `Response`.

KoHcoJlb:
Thanks @GlenDC!
I do plan to stay a while longer, have some more ideas which might be useful that I can contribute.

rama-http/src/service/web/endpoint/mod.rs
GlenDC: Thank you for that explanation, makes a lot of sense! With that in mind, why not make it even more generic by also allowing a generic Body for the Request? That would probably make it as generic as you can get while still keeping it to http?

KoHcoJlb: After reading your initial comment I've also though about it and I'm all for it. My only concern is that I don't want to extend/delay this PR much more.
If it proves simple enough we can do it here, if not, maybe it's better to make it a followup PR.

GlenDC: I think it's up to @soundofspace as he is going to finish this PR tomorrow afaik

GlenDC:
@soundofspace , three small docs/UX nits before merge (feel free to pushback):

 - `with_dir`/`with_file` silently bypass the user's endpoint layer (hard-coded `IntoErrLayer` + `ErrorHandlerLayer`) — worth a doc note.
- `RouterError::Internal` needs a one-liner: e.g. "should never fire, sub-service prefix strip bug".
- `with_endpoint_layer(())` as a reset is invisible — a named `reset_endpoint_layer()` alias would make intent obvious.

Also once you are done make sure to squash merge 😓

rama-http/src/service/web/endpoint/mod.rs
soundofspace: Lets do this in a followup PR, changing this to generic body needs changes in a lot of places

soundofspace (approved):

PR #879, opened 2026-04-26
fix: update embedded datastar to v1.0.1

seuros:
Bumps datastar JS bundle from 1.0.0-RC.8 to 1.0.1 and renames retryMaxWaitMs → retryMaxWait per the v1.0.0 API change.


GlenDC (approved):
Thanks, looks great!

PR #880, opened 2026-04-27
Implement ExtensionsRef for SniRequest

KoHcoJlb:
This allows to use `SniRouter` directly with downstream services that require `ExtensionsRef`.
For example through `MatcherRouter`


GlenDC (approved):

GlenDC:
Thank you, great find.

PR #884, opened 2026-04-29
Connection extensions

### LINKED ISSUE #762: Provide marker wrapper types for Ingress / Egress
A proxy is one category of projects where we have both a client side and a server side. More abstractly, this corresponds to inbound and outbound, or ingress and egress, or north and south. Right now we are in a situation where we:

- either only have one type, which can cause confusion. For example, a TLS tunnel for the incoming HTTP over TLS proxy might add a “secure” marker or similar, but that marker could then confuse the proxy when making a connection on the egress side, where that detail should not matter.
- or we already have explicit wrapper types for specific extension use cases. For example, `ClientSocketInfo`, which is a wrapper around `SocketInfo`.

Now that we have cleaned up the context and extension machinery and made it read only, it might be a good moment before 0.3 to introduce more hygiene here. I propose that we introduce wrapper types whose only purpose is to indicate which side they belong to:

```rust
#[derive(Debug, Clone, PartialEq, Eq, ...)]
pub struct Egress<T>(pub T);

#[derive(Debug, Clone, PartialEq, Eq, ...)]
pub struct Ingress<T>(pub T);
```

Instead of writing:

```rust
ext.get::<SocketInfo>();
```

you would write:

```rust
ext.get::<Ingress<SocketInfo>>();

let Ingress(socket_info) = ext.get().unwrap();
```

The concept is simple enough. My main question is what naming is clearest for everyone:

- ingress and egress
- inbound and outbound
- something else?

I think this, combined with adding side specific information to types that might otherwise still be ambiguous, will help us prevent unexpected or unhygienic behaviour. This way, extension extractors can make well informed decisions without any risk of ambiguity.

Please share your feedback. Once we are aligned, I can implement this quickly. I would prefer doing this before 0.3, since it might otherwise break users subtly if they rely on these types without the new side specific wrappers.

soundofspace:
What this adds:
- Shared connection extensions: a connection can update extensions and req/resp can see the updated content
- Extensions can be forked/have a live parent. We explored this idea before, but now with this shared AppendOnlyVec this has become possible
- Extensions are smart and will traverse ingress/egress/parents to find what is needed in a way that feels natural (hopefully)

Ok, so how does all of this work? Lets have a look at a common use cases: retry_layer -> http_client (with internal connection_pool):

Start two requests at exactly the same time (so pool needs to create two connections)

First request:
1. Retry layers forks the extensions, so retry requests start with the same extensions
1. The first request goes through the connector stack: TCP creates new Extensions and stores it on the connection
1. We go up the connector stack, each connector reads from Request extensions and transfers what it uses to the connection Extensions or adds extra info/state in there
1. We pass the connection pool, so connection will be stored here after use
1. HttpClient stores the connection Extensions as Egress<Extensions> on the request, so further layers down the stack can see this
1. We do the actual request, h2 machinery create a stream. Stream forks the connection extensions, so it has access to them but it doesn't edit them (but it can through internal mutability tricks if needed). It also stores this on the request so layers layers can see this
1. We get a response, extensions are request.extensions().fork(). So all request (and connections) are reachable
1. Connection goes back 

Second request:
1. Retry layers forks the extensions, so retry requests start with the same extensions
1. We do the exact same connection logic as first request
1. HttpClient stores the connection Extensions as Egress<Extensions> on the request, so further layers down the stack can see this
1. Something fails
1. Retry layer makes a retry request, because it forked the extensions nothing we did above "leaked" into this request
1. We go through connector setup, but now pool has the connection from the first request. So we use that
1. HttpClient stores the connection Extensions as Egress<Extensions> on the request, so further layers down the stack can see this
1. We get a response, extensions are request.extensions().fork(). So all request (and connections) are reachable
1. Connection goes back, but importantly we can still read these extensions from the response. So we can do things like `resp.extensions().get_ref::<ConnectionHealth>().mark_broken()`, if this has been stored on the connection we will be able to live modify it, and this will be readable by the pool

But how does this work? The rules for finding extensions are pretty simple
- Search the AppendOnlyVec starting from the end. If you find `T`, return Some(T)
- If you find Egress/Ingress extensions, recurse into these extensions and follow the same rules
- If all extensions are checked without a match, search extensions.parents() with the same rules

The rules kind of map to this:
- If you stored something recent: this should get priority
- If you store a connection ingress/egress, you should search in that entire chain, since connection specific info should always have priority over initial config/extensions
- If not found yet, search the initial extensions since nothing "more important" has recently set it

So how does this work when we do `resp.extensions().get_ref::<ConnectionHealth>()` in a h2 response:
- Search response -> no match and no ingress/egress chain
- Search parent = request_extensions -> at some point we find egress chain
- Search egress_chain = h2 stream -> no match
- Search parent = actual_connection that h2 stream is multiplexed over(tcp/tls/...) -> we find ConnectionHealth

So how does this work when we do `resp.extensions().get_ref::<State>()` in a h2 response:
- Search response -> no match and no ingress/egress chain
- Search parent = request_extensions -> at some point we find egress chain
- Search egress_chain = h2 stream -> no match
- Search parent = actual_connection that h2 stream is multiplexed over(tcp/tls/...) -> no match
- Parent chains are depleted, so we continue in requests_extensions -> find State

Many many many different approaches have been tried and they all failed in one of these things (if you are curious ask me about it):
- Doesn't scale to a mitm situation
- Breaks around upgrades, upgrades are weird they are kind of a connection, but a request/response at the same time
- Should be easy to use. Ingress/egress should be unimportant detail in most stacks, but explicit in mitm setup

I think this works for all of them? But curious what you think

Future work / what this unblocks:
- ConnectionHealth can now be inserted in a much nicer way, it just has to be in connection extensions, order is waaay less tricky
- Custom iterator for Extensions iter -> we dont really rely on this and the boxing approach only has minor overhead, if do use this more, we should make this (will probably do this soon just as future proofing)
- We can use Extension trait (+ derive) to build on this system. Some ideas might be: give a search hint to traverse it in a different way (eg for State: search from the start), add replace-like behaviour...
- Extensions can now be used to talk between connections and the external system. Eg connectionHealth, but a lot more is possible, explore this without going overboard
- Start using this for things like TLS and see what extra utilities might be needed

### DIFF
--- a/examples/http_connect_proxy.rs
+++ b/examples/http_connect_proxy.rs
@@ -58,7 +58,7 @@
 
 use rama::{
     Layer, Service,
-    extensions::{Extension, Extensions, ExtensionsRef, InputExtensions},
+    extensions::{Extension, Extensions, ExtensionsRef},
     http::{
         Body, Request, Response, StatusCode,
         client::EasyHttpWebClient,
@@ -75,7 +75,7 @@ use rama::{
     layer::{ConsumeErrLayer, HijackLayer},
     net::{
         proxy::IoForwardService,
-        stream::{ClientSocketInfo, layer::http::BodyLimitLayer},
+        stream::{SocketInfo, layer::http::BodyLimitLayer},
         user::credentials::basic,
     },
     rt::Executor,
@@ -180,10 +180,12 @@ async fn http_plain_proxy(req: Request) -> Result<Response, Infallible> {
     let client = EasyHttpWebClient::default();
     match client.serve(req).await {
         Ok(resp) => {
+            // We can also just directly fetch SocketInfo and it will traverse into egress/ingress chains,
+            // however to be clear and to avoid confusion in a MITM setup we access the egress one directly.
             if let Some(client_socket_info) = resp
                 .extensions()
-                .get_ref()
-                .and_then(|InputExtensions(ext)| ext.get_ref::<ClientSocketInfo>())
+                .egress()
+                .and_then(|e| e.get_ref::<SocketInfo>())
             {
                 tracing::info!(
                     http.response.status_code = %resp.status(),
--- a/rama-core/src/extensions.rs
+++ b/rama-core/src/extensions.rs
@@ -37,35 +37,71 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use rama_utils::collections::AppendOnlyVec;
+use rama_utils::macros::impl_deref;
 
 pub use rama_macros::Extension;
 
 #[derive(Clone, Default)]
 /// A type map of protocol extensions.
 ///
-/// [`Extension`]s are internally stored in a type erased [`Arc`]. Since values are
-/// stored in an [`Arc`] there are extra methods exposed that build on top of this
-/// and leverage characteristics of an [`Arc`] to expose things like cheap cloning of the Arc.
+/// [`Extension`]s are internally stored in a type erased [`Arc`]. Since values
+/// are stored in an [`Arc`] there are extra methods exposed that build on top
+/// of this and leverage characteristics of an [`Arc`] to expose things like
+/// cheap cloning of the Arc.
+///
+/// [`Extensions`] may have an optional [`parent`][Self::parent]: the
+/// [`Extensions`] this one was forked from. Lookups walk the parent chain when
+/// the local [`Extension`]s don't have the requested type. The parent relationship is
+/// best described as "I'm layered on top of that, but I'm not exactly the same":
+/// - Retry attempts fork from the original request (dont leak failed extensions)
+/// - Responses fork the request (response != request)
+/// - H2 streams fork from underlying H2 connection (nested connection with isolated properties)
+///
+/// Connection's who logically map one-to-one we don't fork and we just pass the [`Extensions`]
+/// up, examples are:
+/// - TLS layered on top of TCP
+/// - HTTP layered on top of TLS
+/// - ...
 pub struct Extensions {
     extensions: Arc<AppendOnlyVec<TypeErasedExtension, 12, 3>>,
+    parent: Option<Box<Self>>,
 }
 
 impl Extensions {
-    /// Create an empty [`Extensions`] store.
+    /// Create an empty [`Extensions`] store with no parent.
     #[inline(always)]
     #[must_use]
     pub fn new() -> Self {
         Self::default()
     }
 
+    /// Create a fresh child [`Extensions`] whose parent is this [`Extensions`] store.
+    ///
+    /// The child has its own empty top-level storage. Lookups that miss
+    /// locally walk into the parent (and recursively up the chain). Inserts
+    /// land on this child only, parents are never mutated through the child.
+    #[must_use]
+    pub fn fork(&self) -> Self {
+        Self {
+            extensions: Arc::new(AppendOnlyVec::new()),
+            parent: Some(Box::new(self.clone())),
+        }
+    }
+
+    /// The parent [`Extensions`] this blob was forked from, if any.
+    #[inline(always)]
+    #[must_use]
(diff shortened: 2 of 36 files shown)


examples/http_connect_proxy.rs
```diff
@@ -180,10 +180,12 @@ async fn http_plain_proxy(req: Request) -> Result<Response, Infallible> {
     let client = EasyHttpWebClient::default();
     match client.serve(req).await {
         Ok(resp) => {
+            // We can also just directly fetch SocketInfo and it will traverse into egress/ingress chains,
+            // however to be clear and to avoid confusion in a MITM setup we access the egress one directly.
             if let Some(client_socket_info) = resp
                 .extensions()
-                .get_ref()
-                .and_then(|InputExtensions(ext)| ext.get_ref::<ClientSocketInfo>())
+                .egress()
```
GlenDC, line 187: is it worth adding also a shortcut egress_ref / egress_arc etc? Or is this not expected to be common?

rama-http-types/src/response.rs
```diff
@@ -540,6 +540,12 @@ impl<T> Response<T> {
             head: self.head,
         }
     }
+
+    /// Mutable reference to this request's extensions
+    #[inline]
+    pub fn extensions_mut(&mut self) -> &mut Extensions {
```
GlenDC: Is this only used for replacing extensions? If so not more clear (intent wise) to go for good old setter (`generate_set_and_with`)?

GlenDC: (same for other records where you added such a method)

rama-core/src/extensions.rs
```diff
@@ -104,8 +140,22 @@ impl Extensions {
...
     #[must_use]
     pub fn contains<T: Extension>(&self) -> bool {
+        self.get_ref::<T>().is_some()
+    }
+
+    /// Returns true if the [`Extensions`] store contains the given type
+    ///
+    /// This only checks this [`Extensions`] store
+    #[must_use]
+    pub fn raw_contains<T: Extension>(&self) -> bool {
```
GlenDC: raw_contains is not a clear name... would perhaps name it `self_contains` or something like that?

GlenDC (approved):
Looks good to me. A lot simpler than I had imagined. Love it. Just a small amount of comments, but nothing that would change the design or direction of this PR. Well done.

examples/http_connect_proxy.rs
soundofspace: Its only common in mitm setups and it feels like the shortcut would only offer very minor convenience, so would not do it

rama-core/src/extensions.rs
```diff
@@ -168,77 +297,232 @@ impl Extensions {
...
-    /// Get a shared reference to the oldest inserted item of type `T`
+    /// Raw flat find-or-create: return `&T` if one exists at the top level of
+    /// this [`Extensions`] store, otherwise insert the value produced by
+    /// `create_fn` at the top level and return a reference to it.
     ///
-    /// If an owned `Arc<T>` is needed prefer using [`Self::get_arc()`]
+    /// Does not follow the parent chain. Useful when you want strict "ensure T
+    /// exists on THIS blob" (e.g. materializing a direction wrapper
+    /// like [`Ingress<Connection<Extensions>>`] at the outer blob).
+    pub fn raw_get_ref_or_insert<T, F>(&self, create_fn: F) -> &T
```
GlenDC: these are still called raw_, while you already agreed with the other one to name it self_ instead? So these raw_ ones are still left-over?

GlenDC (approved):
LGTM, just one comment left that was unresolved (fully) from last review

rama-core/src/extensions.rs
soundofspace: I misread that as you only found it confusing for contains, updated all of them now

PR #888, opened 2026-05-02
http: multipart support + octet-stream extractor

### LINKED ISSUE #647: support application/octet-stream responses

GlenDC:
Adds `multipart/form-data` support to rama (server + client), gated behind
a new `multipart` cargo feature included in `http-full`. Builds on `multer`.

Also fills the gap on octet-stream: there was already a response type but
no request extractor or client helper.

Includes example, integration test, and a book page covering both.

Closes #647


rama-cli/src/cmd/send/http/request.rs
```diff
@@ -92,15 +94,90 @@ pub(super) async fn build(cfg: &SendCommand, is_ws: bool) -> Result<Request, Box
...
+            DataInput::Body { body, content_type } => {
+                request.headers_mut().typed_insert(content_type);
+                // Re-sync Content-Length with the body we're about to
+                // ship. Pure in-memory `--data` items take the fast path
+                // in `build_data_input` and arrive here as a sized
+                // `Body::from(Vec<u8>)` whose `size_hint` is exact;
+                // stdin/file-backed mixes arrive as a `Body::from_stream`
+                // with unknown length, in which case we drop any pre-set
+                // CL header so we don't ship stale framing.
+                use rama::http::StreamingBody as _;
```
GlenDC: needs to be moved to head imports

rama-cli/src/cmd/send/http/request.rs
```diff
@@ -92,15 +94,90 @@ pub(super) async fn build(cfg: &SendCommand, is_ws: bool) -> Result<Request, Box
...
+                match body.size_hint().exact() {
+                    Some(len) => {
```
GlenDC: if-else

rama-cli/src/cmd/serve/httptest/endpoint/octet_stream.rs
```diff
@@ -0,0 +1,38 @@
...
+    service::service_fn,
+    stream::io::ReaderStream,
+};
+use std::convert::Infallible;
+
+pub(in crate::cmd::serve::httptest) fn service()
+-> impl Service<Request, Output = Response, Error = Infallible> {
+    (
+        ConsumeErrLayer::trace_as_debug(),
+        BodyLimitLayer::new(8 * 1024 * 1024),
```
GlenDC: I think it makes sense that we introduce some helpers here, kb / mb, ... as this is a bit silly...

rama-http/src/service/client/ext.rs
```diff
@@ -731,6 +771,117 @@ where
...
+    /// known size, the `Content-Length` header. An existing `Content-Type`
+    /// header is overwritten.
+    #[cfg(feature = "multipart")]
+    #[must_use]
+    pub fn multipart(mut self, form: super::multipart::Form) -> Self {
+        let content_type = form.content_type();
+        let content_length = form.content_length();
+        let body = form.into_body();
+        self.state = match self.state {
+            RequestBuilderState::PreBody(mut builder) => {
```
GlenDC, line 789: this flow can be simplified

rama-http/src/service/client/ext.rs
```diff
@@ -786,6 +937,40 @@ where
...
 }
 
+/// Synchronise `Content-Length` with the new body on a pre-body request
+/// builder.
+///
+/// If the body has a known exact size, set `Content-Length` to that value
+/// (replacing any prior). If the size is unknown, remove any pre-set
+/// `Content-Length` so the request doesn't ship with stale framing.
+fn sync_content_length_pre_body(builder: &mut crate::request::Builder, body: &crate::Body) {
+    use rama_http_types::StreamingBody as _;
```
GlenDC: no inline imports please unless  it makes feature flags easier to deal with, but even than... either way doesn't apply here... no inline import

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+///
+/// Generates a random boundary on construction. Add named [`Part`]s with
+/// [`text`](Self::text), [`bytes`](Self::bytes), [`file`](Self::file), or
+/// [`part`](Self::part), then convert to a [`Body`](crate::Body) (or feed via
+/// [`RequestBuilder::multipart`](super::ext::RequestBuilder::multipart)).
+#[derive(Debug)]
+#[must_use]
+pub struct Form {
+    boundary: String,
+    parts: Vec<NamedPart>,
```
GlenDC: would a smallVec be appropriate here? how frequently do we get very large parts? Is there a common max in practice that would allow us to stack for most cases?

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+/// A multipart form body.
+///
+/// Generates a random boundary on construction. Add named [`Part`]s with
+/// [`text`](Self::text), [`bytes`](Self::bytes), [`file`](Self::file), or
+/// [`part`](Self::part), then convert to a [`Body`](crate::Body) (or feed via
+/// [`RequestBuilder::multipart`](super::ext::RequestBuilder::multipart)).
+#[derive(Debug)]
+#[must_use]
+pub struct Form {
+    boundary: String,
```
GlenDC: boundary seems usually pretty small? Can we use smollstr (rama vendors this)

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+        &self.boundary
+    }
+
+    /// The `Content-Type` header value (`multipart/form-data; boundary=…`).
+    #[must_use]
+    pub fn content_type(&self) -> HeaderValue {
+        let value = format!("multipart/form-data; boundary={}", self.boundary);
+        // Boundaries are randomly generated hex chars, so the resulting
+        // value is always a valid ASCII header value.
+        HeaderValue::try_from(value)
```
GlenDC: I would either expect this, or add warning log at the elast so we do not filent silence here...

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+    pub fn content_length(&self) -> Option<u64> {
+        let mut total: u64 = 0;
+        for np in &self.parts {
+            let part_size = np.part.content_size?;
+            let header_len = part_headers_len(&self.boundary, &np.name, &np.part) as u64;
+            total = total.checked_add(header_len)?;
+            total = total.checked_add(part_size)?;
+            total = total.checked_add(CRLF.len() as u64)?;
+        }
+        let trailer_len = (b"--".len() + self.boundary.len() + b"--\r\n".len()) as u64;
```
GlenDC: should we not add constants for these byte slice constants as they are used in multiple locations?

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+    pub async fn file<P: AsRef<Path>>(path: P) -> std::io::Result<Self> {
+        let path = path.as_ref();
+        let file_name: Option<Cow<'static, str>> = path
+            .file_name()
+            .map(|name| Cow::Owned(name.to_string_lossy().into_owned()));
+        let mime = path
+            .extension()
+            .and_then(std::ffi::OsStr::to_str)
+            .and_then(|ext| mime_guess::from_ext(ext).first())
+            .unwrap_or(mime::APPLICATION_OCTET_STREAM);
```
GlenDC, line 351: might want to add at the very least a debug trace here

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+    }
+
+    /// Resolve this spec into a [`Part`], performing any necessary I/O.
+    pub async fn into_part(self) -> Result<Part, FieldSpecError> {
+        let mut part = match self.source {
+            FieldSpecSource::Text(s) => Part::text(s.to_owned()),
+            FieldSpecSource::File("-") => Part::stream(read_stdin_stream()),
+            FieldSpecSource::File(path) => Part::file(path)
+                .await
+                .map_err(|e| FieldSpecError::InvalidModifier(format!("file open: {e}")))?,
```
GlenDC: might be cheaper that we use BoxError for invalidModifier so that we can do `e.context("...")` would avoid having to force display the error and allocate an entire string

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+                let s = tokio::fs::read_to_string(path)
+                    .await
+                    .map_err(|e| FieldSpecError::InvalidModifier(format!("file read: {e}")))?;
+                Part::text(s)
+            }
+        };
+        if let Some(ct) = self.content_type {
+            part = part
+                .with_mime_str(ct)
+                .map_err(|_| FieldSpecError::InvalidModifier(format!("type={ct}")))?;
```
GlenDC: with boxError approach this error could be preserved... I think?

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+            }
+            FieldSpecSource::FileText(path) => {
+                let s = tokio::fs::read_to_string(path)
+                    .await
+                    .map_err(|e| FieldSpecError::InvalidModifier(format!("file read: {e}")))?;
+                Part::text(s)
+            }
+        };
+        if let Some(ct) = self.content_type {
+            part = part
```
GlenDC: why not use the `set_` alternatives for these? So that we do not need to re-assign each time???

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+/// the value from the modifier list. Returns `None` if there are no
+/// modifiers.
+fn split_modifiers(input: &str) -> Option<(&str, &str)> {
+    // Naive split on the first `;` is fine here — values in field specs
+    // do not have a quoted form, by convention.
+    input.split_once(';')
+}
+
+async fn read_stdin_to_string() -> std::io::Result<String> {
+    use tokio::io::AsyncReadExt as _;
```
GlenDC: no iinline imports please

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+    let mut buf = String::new();
+    tokio::io::stdin().read_to_string(&mut buf).await?;
```
GlenDC: add context please and return `BoxError`? WDYT?

rama-http/src/service/client/multipart.rs
```diff
@@ -0,0 +1,859 @@
...
+
+fn read_stdin_stream() -> impl rama_core::futures::Stream<Item = Result<Bytes, BoxError>> + Send {
+    rama_core::stream::io::ReaderStream::new(tokio::io::stdin())
+        .map_ok(Bytes::from)
+        .map_err(BoxError::from)
+}
+
+fn gen_boundary() -> String {
+    let mut rng = rand::rng();
+    format!(
```
GlenDC: with boundary as smollstr, this can be also smollstr format

rama-http/src/service/web/endpoint/extract/body/multipart.rs
```diff
@@ -0,0 +1,835 @@
...
+    /// builder methods.
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Set the default per-field byte limit applied when no field-specific
+    /// limit overrides it.
+    #[must_use]
+    pub fn with_default_field_limit(mut self, limit: u64) -> Self {
```
GlenDC: when generating set / with methods, please use the rama `generate_set_and_with` please

GlenDC (commented):
did a self-review, looks mostly good, some minor details that I need to address, afterwards I will merge it once tests pass

PR #893, opened 2026-05-05
Fix Apple transparent-proxy CI and Swift FFI build

GlenDC:
Fix the Apple transparent-proxy CI path and Swift FFI build issues by adding the missing Swift/C target wiring, updating the Apple QA workflow to install its required tools and Rust targets, and keeping the transparent-proxy QA checks in CI.

Also fix some other stuff in rama repo that I noticed while debugging, analysing logs or reading code... (e.g. some minor http improvements)


GlenDC (commented):
Self-reviewed... looks good. Found some minor details while testing a bit more, so will commit these still.. But if that passes CI I will merge it in.

PR #895, opened 2026-05-10
rama-ws: security and correctness fixes

GlenDC:
- vendor RFC 7692 (permessage-deflate spec)
- deflate bomb: enforce `max_message_size` on decompressed payload, not just compressed wire bytes
- `Sec-WebSocket-Accept` validation: fix silent bypass when key computation errors (converted `.ok()` to hard error)
- protocol mismatch: server electing no subprotocol when client proposed some is valid per RFC 6455 (downgraded to trace log)
- `write_out_buffer`: eliminate O(n)2 `drain` loop — use cursor + single `clear()` on success
- `parse_internal`: replace `read` + length check with `read_exact` + `UnexpectedEof` pattern for header and mask bytes

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -137,7 +137,7 @@ jobs:
 
   test-ffi-apple-example-transparent-proxy-e2e:
     runs-on: macos-latest
-    timeout-minutes: 30
+    timeout-minutes: 90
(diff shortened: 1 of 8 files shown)


GlenDC (commented):
self-review looks good

PR #898, opened 2026-05-11
fix HttpExporter use runtime

soundofspace:
- Store tokio runtime so HttpExporter can spawn on tokio. Otherwise this exporter will not work
- Also add integration test for this example so we know this is working (also confirmed that this was initially broken)

### DIFF
--- a/rama-grpc/src/service/opentelemetry/http.rs
+++ b/rama-grpc/src/service/opentelemetry/http.rs
@@ -66,6 +66,7 @@ pub struct HttpExporter<S = ()> {
     temporality: Temporality,
     resource: Arc<ArcSwap<transform::ResourceAttributesWithSchema>>,
     shutdown: Arc<AtomicBool>,
+    runtime: Option<tokio::runtime::Handle>,
 }
 
 #[derive(Debug, Clone, Default)]
@@ -112,6 +113,7 @@ impl<S> HttpExporter<S> {
                 transform::ResourceAttributesWithSchema::default(),
             )),
             shutdown: Arc::new(AtomicBool::new(false)),
+            runtime: tokio::runtime::Handle::try_current().ok(),
         }
     }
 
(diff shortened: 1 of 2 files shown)


rama-grpc/src/service/opentelemetry/http.rs
```diff
@@ -333,17 +347,29 @@ where
...
+                None => service.serve(request).await,
+            }
+            .map_err(|err| OTelSdkError::InternalFailure(err.into().to_string()))?;
 
-        decode_response(response).await
+            decode_response(response).await
+        };
+
+        match self.runtime.as_ref() {
+            Some(handle) => handle
```
GlenDC, line 367: what's the value of spawning if you are anyway awaiting it with no additional work?

soundofspace: Are there better options? I spawn this to make it run on tokio executor. Otel is currently calling this from a non tokio thread so without this it breaks

GlenDC: how are they doing async await without a tokio runtime? Another runtime? o.O

soundofspace: `futures_executor::block_on`, so yea. Fyi before grpc changes to integrate this logic we also had this exact spawn logic

GlenDC: ok ok, in that case go for it.... What a mess otel is

GlenDC (approved):

rama-grpc/src/service/opentelemetry/http.rs
soundofspace: Well here i kind of agree with them... Forcing tokio on everyone seems like a bad idea, and this executor does work for async stuff, it just doesn't work with the async stuff we want to use

GlenDC: Feels to me like a pretty poor way to abstract it if you ask me. There must be better / richer ways to let users inject their runtime of choice here. Nothing has to be forced there. It can be their own trait or w/e which can than in a seaprate tokio-specific otel crate be implemented...

GlenDC: anyway out of scope for this PR and not even the right repo to discusss it

GlenDC:
@soundofspace : merge main into here, and you should have a green CI

PR #899, opened 2026-05-11
rama-fastcgi: FastCGI gateway crate

GlenDC:
Introduces `rama-fastcgi` (`rama::gateway::fastcgi`, feature `fastcgi`): FastCGI 1.0 with HTTP adapters, streaming bodies, configurable caps/timeouts, and `proto::cgi::*` constants.

Two php-fpm demos under `examples/gateway/fastcgi-php/` — gateway (HTTPS + TCP) and migration (HTTP + Unix-socket); both run end-to-end in CI on `ubuntu-latest` via `just test-fastcgi-php`, and interactively via `just example-fastcgi-php-{gateway,migration}`.


GlenDC (commented):
Self-review done. Found couple of remarks and mistakes, looking into them now. After that it will be waiting for CI and merge.

PR #901, opened 2026-05-12
otel HttpExporter support logs

soundofspace:
- Vendor grpc files for logs and logs service
- impl LogExporter for HttpExporter similar to metrics/spans
- Extend example and test to also test if this works


rama-grpc/src/service/opentelemetry/http.rs
```diff
@@ -470,6 +534,47 @@ where
     }
 }
 
+impl<S> LogExporter for HttpExporter<S>
```
GlenDC: is it ok that we do not implement the optional trait methods? Should we add the clippy lint to guard against it here?

tests/integration/examples/example_tests/http_telemetry.rs
```diff
@@ -42,23 +44,38 @@ async fn test_http_telemetry() {
...
-    while state.metrics_received.load(Ordering::SeqCst) == 0 {
+    loop {
+        let metrics = state.metrics_received.load(Ordering::SeqCst);
+        let logs = state.logs_received.load(Ordering::SeqCst);
+        if metrics > 0 && logs > 0 {
+            break;
+        }
         if Instant::now() > deadline {
-            panic!("no OTLP /v1/metrics POST reached the fake collector within 10s");
+            panic!(
```
GlenDC, line 64: not your fault, as I already didn't do before, but I wonder if we should also check if a (log) record was decoded (next to just checking if one was received)

rama-grpc/src/service/opentelemetry/transform.rs
```diff
@@ -141,6 +144,21 @@ fn instrumentation_scope_into(
     }
 }
 
+fn log_instrumentation_scope_into(
+    scope: &opentelemetry::InstrumentationScope,
+    target: Option<&str>,
+) -> ProtoInstrumentationScope {
+    match target {
```
GlenDC, line 151: is this implemented correctly? You are dropping scope metadata if target is set???

rama-grpc/src/service/opentelemetry/http.rs
GlenDC: Not entirely your fault as I had this already it seeems, but cloning this log makes this shutdown signal be shared... or if you have a http exporter + metric exporter that you create by cloning same http exporter you will also have this be shared... 

I imagine that the error of such a shutdown is not gonn ever cause real issues, but stil..... I think we have a minor footgun here?

rama-grpc/src/service/opentelemetry/transform.rs
```diff
@@ -535,6 +553,137 @@ fn exp_histogram_into<T: Numeric>(hist: &SdkExponentialHistogram<T>) -> ProtoExp
...
+            .severity_number()
+            .map(|s| s as i32)
+            .unwrap_or_default(),
+        severity_text: record
+            .severity_text()
+            .map(ToOwned::to_owned)
+            .unwrap_or_default(),
+        body: record.body().and_then(log_any_value_into),
+        attributes,
+        dropped_attributes_count: 0,
```
GlenDC: This one is always 0? Even though you have a filter map?

GlenDC (changes requested):

rama-grpc/src/service/opentelemetry/transform.rs
soundofspace: Yea, this is what official impl does. To be honest I'm not convinced that implementing this ourselves is a good idea, seems that over time we might miss out on "fixes" in otel-proto crate.

```rust
    impl
        From<(
            &opentelemetry::InstrumentationScope,
            Option<Cow<'static, str>>,
        )> for InstrumentationScope
    {
        fn from(
            data: (
                &opentelemetry::InstrumentationScope,
                Option<Cow<'static, str>>,
            ),
        ) -> Self {
            let (library, target) = data;
            if let Some(t) = target {
                InstrumentationScope {
                    name: t.to_string(),
                    version: String::new(),
                    attributes: vec![],
                    ..Default::default()
                }
            } else {
                InstrumentationScope {
                    name: library.name().to_owned(),
                    version: library.version().map(ToOwned::to_owned).unwrap_or_default(),
                    attributes: Attributes::from(library.attributes().cloned()).0,
                    ..Default::default()
                }
            }
        }
    }
    ```

rama-grpc/src/service/opentelemetry/http.rs
soundofspace: The only two not implemented:
- shutdown(), which forwards to shutdown_with_timeout() which we have
- event_enabled() -> default true matches upstream otel exporter also (also not specified there)

soundofspace: I think we mainly need to use the clippy lint if we do wrapper around things, there its really important to forward logic to the inner one, otherwise default/implementation depends on what the expected behaviour is

tests/integration/examples/example_tests/http_telemetry.rs
soundofspace: Added some extra checks for this

rama-grpc/src/service/opentelemetry/http.rs
soundofspace: Good remark, never hurts to do it properly, split it into 3 signals

rama-grpc/src/service/opentelemetry/transform.rs
soundofspace: I migrated the official code which uses unreachable and decodes everything. But we have clippy lint to not use that so i migrated to filter_map to work around that. So that means that right now we do not drop anything (so 0 like official one) but agree that it was sloppy to not future proof it and already count dropped items

soundofspace:
@GlenDC the tokio runtime fix I did, I apparently only did for the http exporter, but it is needed for the grpc one also, and I made that exact same mistake when migrating to 3 shutdown signals. So I merged http and grpc exporters into one exporter to help with that, only very little code is actually transport specific here. I will probably also move the base logic of this to rama-core overtime so we could also support the json exporter without grpc requirements.

But yea I'm still not convinced that doing everything ourselves is the best idea here, will make it harder to stay in sync with upstream. On the other hand, it is kind of nice that own everything now, so lets see

rama-grpc/src/service/opentelemetry/http.rs
GlenDC: agreed

rama-grpc/src/service/opentelemetry/transform.rs
GlenDC: agreed

tests/integration/examples/example_tests/http_telemetry.rs
GlenDC: thx

GlenDC (approved):
Thx, nice cleanup. Looks solid now!

PR #906, opened 2026-05-13
rama-http improvements p2

GlenDC:
Fix several issues spotted in rama-http, see commits.


GlenDC (commented):
self-review: all good

PR #909, opened 2026-05-15
pool BasicHttpConnIdentifier include connector/proxy target

soundofspace:
Make pool aware of ProxyAddress and ConnectorTarget so we respect this when giving out connections.

Current logic could be considered a bug, but mainly adding this because as a next step I will be adding a DnsLoadbalancer that will set ConnectorTarget to loadbalancing between multiple IPs (for grpc use cases).

### DIFF
--- a/rama-net/src/client/pool/http.rs
+++ b/rama-net/src/client/pool/http.rs
@@ -1,5 +1,6 @@
 use super::{LruDropPool, PooledConnector, ReqToConnID};
-use crate::{Protocol, address::HostWithOptPort, http::RequestContext};
+use crate::address::ProxyAddress;
+use crate::{Protocol, address::HostWithOptPort, client::ConnectorTarget, http::RequestContext};
 use rama_core::error::BoxError;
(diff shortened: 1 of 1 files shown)


GlenDC (approved):

GlenDC:
Looks good to me,.. That said, would you also not want the username (if basic auth is used) to be included, given that is often used in proxies for labels.... Because if the proxy authority is a gateway, than its pretty generic and will still give weird results, no? How do you see that?

soundofspace:
> Looks good to me,.. That said, would you also not want the username (if basic auth is used) to be included, given that is often used in proxies for labels.... Because if the proxy authority is a gateway, than its pretty generic and will still give weird results, no? How do you see that?

Honestly I completely overlooked this since I was focussed on my use case mostly. Change to include proxy address and connector target. This makes it even better since we removed the calculate target logic and we can keep logic very simple here

GlenDC (approved):

PR #911, opened 2026-05-15
Keep extensions on upgrade

maartendeprez:
Currently, extensions get lost on proxy CONNECT because `InnerHttpProxyConnector` starts out with a new, empty set of extensions. This PR makes it fork from the connection's extensions instead.

### DIFF
--- a/rama-http-backend/src/client/proxy/layer/proxy_connector/connector.rs
+++ b/rama-http-backend/src/client/proxy/layer/proxy_connector/connector.rs
@@ -4,7 +4,7 @@
 
 use super::HttpProxyError;
 use rama_core::error::{BoxError, ErrorContext};
-use rama_core::extensions::{Extension, ExtensionsRef};
+use rama_core::extensions::{Extension, Extensions, ExtensionsRef};
 use rama_core::io::Io;
 use rama_core::rt::Executor;
 use rama_core::telemetry::tracing;
@@ -58,7 +58,18 @@ impl InnerHttpProxyConnector {
     }
 
     rama_utils::macros::generate_set_and_with! {
-        /// Add a header to the request.
+        /// Set the extensions for the connector.
+        pub(super) fn extensions(
+            mut self,
+            exts: Extensions,
+        ) -> Self {
+            self.req.set_extensions_mut(exts);
+            self
+        }
+    }
+
+    rama_utils::macros::generate_set_and_with! {
+        /// Add an extension to the request.
         pub(super) fn extension(
             mut self,
             value: impl Extension,
--- a/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
+++ b/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
@@ -205,7 +205,8 @@ where
             });
         }
 
-        let mut connector = InnerHttpProxyConnector::new(transport_ctx.authority.clone())?;
+        let mut connector = InnerHttpProxyConnector::new(transport_ctx.authority.clone())?
+            .with_extensions(conn.extensions().clone().fork());
 
         if let Some(version) = self.version {
             connector.set_version(version);


GlenDC:
i'll let @soundofspace review this one, as he's working on extensions atm 

that said, can you please bundle this with 1 or more regression tests, so that we don't break in future w/e fixes we land on in the end

soundofspace:
Hey @maartendeprez , good to see you here!

Can you share a bit more about which extensions you need there / want to set. This request is definitely something I missed. I'm wondering if forking connection extensions is enough here since this would only give you access to the extensions that have been set/transferred to the connection. 

I'm thinking here it would probably be more flexible to do this instead:

```rust
let proxy_connect_ext = input.extensions().fork();
proxy_connect_ext.insert(Egress(conn.extensions().clone()));

let mut connector = InnerHttpProxyConnector::new(transport_ctx.authority.clone())?
    .with_extensions(proxy_connect_ext);
    
// or even this since extensions are a crucial piece
let mut connector = InnerHttpProxyConnector::new(transport_ctx.authority.clone(), proxy_connect_ext)?

```

This would give you access to all input/request extensions and the connection extensions similar to other places outside a connector. On the other hand, if there is a need to transfer extensions back out of this inner connector we might want to just set `conn.extenions.clone()` (without the fork) instead. Both are valid options depending on how we see this inner connector, currently I'm leaning towards the first option as it would get you access to everything, and it could be used to set extensions on the connection with `req.egress().map(|ext|ext.insert(val))`, but I'm really curious what you think about this.

maartendeprez:
The reason i did not transfer the `input` extensions as well, is that the `input` itself is returned to the caller, thus allowing continued access to those extensions. Copying them to the inner connector would create differentiation in how extensions are handled depending on whether and what kind of proxy is used.

There are three successful return paths out of this service:

- [direct connection](https://github.com/maartendeprez/rama/blob/abf683d166b3a0ce92711ba1ecdf876e0ffd259c/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs#L177)
- [proxied connection over http (no CONNECT)](https://github.com/maartendeprez/rama/blob/abf683d166b3a0ce92711ba1ecdf876e0ffd259c/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs#L202)
- [proxied connection over https (upgrade after CONNECT)](https://github.com/maartendeprez/rama/blob/abf683d166b3a0ce92711ba1ecdf876e0ffd259c/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs#L254)

In all three cases, we return `Ok(EstablishedClientConnection { input, conn })`  with a modified `conn`. In the first two cases, `conn` is simply wrapped in a variant of `MaybeHttpProxiedConnection` and we do not lose access to the original `conn` extensions. In the third case, we currently do lose access because `conn` is created through an entirely [new connector](https://github.com/maartendeprez/rama/blob/abf683d166b3a0ce92711ba1ecdf876e0ffd259c/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs#L208) and the [handshake](https://github.com/maartendeprez/rama/blob/abf683d166b3a0ce92711ba1ecdf876e0ffd259c/rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs#L239) does not transfer extensions from the original `conn` to the returned `Upgraded` object.

I´d try to avoid as much as possible any differential handling of extensions depending on the proxy situation, that would require the user to be aware of internal extension handling details (that could easily change on updates unless we explicitly define the semantics).

Regarding copying of the `input` extensions to the connection extensions, that would mean we either do this in all three cases or in none of them. I don't believe this is common in other connector services, so I think it would be surprising to do so for this one. (For pooled connections, for instance, that would mean only the extensions of the first input would be copied.)

Regarding the copying of the original connection extensions to the output connection, there are three possibilities for the upgraded case:

1. We carry them over directly to the inner connector. (`conn.extensions().clone()`).
2. We insert a forked version into the inner connector (`conn.extensions().clone().fork()`, current impl in this PR) 
3. We insert the original connection's extensions wrapped in `Egress` (`Egress(conn.extensiont().clone())`)

Option 1 most resembles the other two returns paths and would, thus, probably be the least surprising. Extensions inserted into the upgraded connection would become available on the original connection for all three cases. If I am not mistaken, options 2 and 3 are structurally equivalent (there are basically three kinds of fork), and both would introduce the same issue that one would need to special-case the upgraded conn case when inserting connection extensions supposed to be carried over into the next use of the connection. For that reason, I am currently in favor of option one.

soundofspace:
@maartendeprez I agree with you about consistency, but using `conn.extensions().clone()` will also cause inconsistencies sadly. Extensions around upgrades in general are still very rough, and this specific case just shows it.

An upgrade using H2 (without this inner connector) for example does this:
1. TCP -> Tls connectors all reuse connection extensions
2. When we create a H2 stream -> conn.extensions.fork(), since are on a new logical connection
3. Create upgrade, this gets the H2 stream connections (no fork), since this consumes the stream
4. Handle upgrade -> re-parrent upgraded extensions to request/response that caused this upgrade

If we set conn.extensions.clone() here on the request that caused this upgrade. We are skipping the H2 stream fork for example. Which might be fine in this specific example since this connection is not shared (but it could be in the future). 

Step 4 is basically needed for mitm setups where request becomes an upgraded and they are logically linked together (or things like websockets). But when creating this logic this always felt like a trade-off (and a too easy solution) since an upgrade is still a connection, and should probably never be re-parrented to the input and should only have the connection extensions. I accepted that trade-offs since everything was working as expected, and it all just seemed to work... Until it didn't. 

So I think what needs to happen (Will do this now, but would love your feedback also):
- Remove step 4, so upgrades only transfer connection extensions, and dont re-parrent (and deal with fallout of this)
- Clone request.extensions to request of inner connector, so inner connector has a similar setup as outer connectors (input extensions on request, connection extensions on connection/stream/upgrade)

I will also be checking what else I need to do around upgrades to make it more consistent and predictable. If you have ideas here or in general remarks about this setup, do let me, I'm always interested in what you think / how you would approach this.

soundofspace:
[This PR](https://github.com/plabayo/rama/pull/914) adds a test for it, and shows that removing this re-parrenting from handle_upgrade actually fixes it. 

We do need it in `UpgradeService` to prevent some tests from breaking, if this re-parrent approach there is a proper solution is another question. I also added it in comments in other places where we might also need it to see what we should do there, if anything

maartendeprez:
I can confirm that your PR works to preserve the connection extensions, but the drawback is we lose the extensions from the original `req`. For example, the `ProxyAddress` is no longer available in the connection extensions in the test. In the testcase, this works before the change, but not after: 

```
        let proxy_addr = conn
            .extensions()
            .get_ref::<ProxyAddress>()
            .expect("ProxyAddress set on the request must survive the upgrade");
        assert_eq!(proxy_addr.address, HostWithPort::example_domain_http());
```

Doing this in `handle_upgrade` fixes the `test_https?_echo` test cases, but not the above assert:

```
        let exts = msg_ext.fork();
        exts.insert(Egress(upgraded.extensions().clone()));
        Ok(upgraded.with_extensions(exts))
```

soundofspace:
@maartendeprez that also is the case for all other connectors. The idea is that connectors should transfer input extensions to the connection that become part of the connection (this part is still a wip in a lot of places). All connectors should also have access to request extensions. In my draft PR I still needed to "Clone request.extensions to request of inner connector" which would solve this issue.

And then as followup continue with cloning extensions from input to connection once they are used. At one point I had logic where extensions had a scope(input/connection) which could automate this part, but it was rather complex and didn't offer much benefit, so I decided that connectors had the responsibility to do this. What do you think about this?

maartendeprez:
@soundofspace Looking good. So you basically re-added this PR on top of the removal of the re-parenting in `handle_upgrade`.

I find it worrisome that the [special-casing in UpgradeService](https://github.com/plabayo/rama/blob/9664176fa57ee27e70f1ff5ba24e2322c96bd21a/rama-http-backend/src/server/layer/upgrade/service.rs#L127) is still required to make the cli tests pass. Something do to with the removed `connector.set_extension(map);`?

I agree with the idea of leaving it up to connectors to copy the relevant extensions (the ones they actually use) from the input to the connection. That seems safer than copying based on a scope setting that could be set on an extension at any level of the stack.

The scope idea might be a solution for the problem of "inserting an extension on the connection" discussed earlier though. Currently we'd need to "guess" the number of `.parent()` (or `.egress()`) calls needed, which can be different depending on the number of intermediate CONNECT calls, H1/H2 proto, and so on. If we could instead use `req.extensions().scope(ExtScope::APPLICATION_LAYER_CONNECTION).insert(...)` with appropriate interpretations for each protocol to pinpoint the right `fork` level (e.g. for H1, session and connection would be the same), that could make it more reliable. Then again, in such cases one could probably insert a "receiver" extension (one containing `Mutex<Option<...>>` or similar) at the right position in the stack and then push to it later.

soundofspace:
The special case in upgradeService is because in that example we insert a config into the input which is upgraded to a websocket and that config is then used later. So the input kind of becomes the connection here which is then again the input (weird, yet valid use case). I'm currently exploring this specific use case more to see if this is actually the example doing something wrong or if we can make a better solution for this.

> we'd need to "guess" the number of .parent() (or .egress()) calls needed)

This should actually not be the case. Default extensions methods such as get_ref traverse up the chain and should always be able to find the extension. Ingress/egress were only added so it would be possible to be explicit in a mitm setup where things like SocketInfo could be inserted on both sides. But in pretty much all other use cases ingress/egress/parent should never be used manually (unless you have a special use case)

maartendeprez:
@soundofspace 

> Default extensions methods such as get_ref traverse up the chain and should always be able to find the extension.

Reading is fine. I was referring specifically to the case of inserting an extension on the connection you mentioned earlier:

> and it could be used to set extensions on the connection with req.egress().map(|ext|ext.insert(val))

soundofspace:
Ah yea that was a special case indeed, arguably one that should not be used. In the beginning I liked the scope idea, but it become really complex and really big fast. 

The assumption is now is that for connections you should always insert in the active layer by default. And for cases where that doesn't apply the connector that "owns" it should have already inserted it for you eg `ConnectionHealthWatcher` is a good example of that, exactly what you describe with the receiver approach.

Now why that example/cli is failing in on the server side of a websocket this happens:
- Incoming connect request to upgrade to websocket
- Websocket upgrade "responder" detects the "to upper" protocol and inserts it into the request
- It returns a 200 response accepting the upgrade
- The extensions are now copied, in a pretty dump way to the upgraded so "handler" can read this
- Upgraded is converted to AsyncWebSocket which then reads `AcceptedWebSocketProtocol`

If seems that the proper solution here would be to not copy request/response extensions blindly over to the Upgraded. But change the upgrade "responder" signature so it gets an Extensions store or it should return one that should be applied to the upgrade. This would make this explicit and clear. This is also something we "abuse" (might be to hard statement) in other upgrade "responder" to do `req.extensions().insert(ProxyTarget(authority));` to actually set a ProxyTarget on the Upgrade, or sneak other things from a "responder" to a "handler". 

So basically where we convert a requestinto an Upgrade we should make extensions we transfer explicit so we mimic what we do in connector_stack, but now on the server side.

On the other hand, the request kind of becomes the upgraded here. So maybe we dont reparent but extend the connection extensions or we also insert them using a special marker?

soundofspace:
@maartendeprez [this PR](https://github.com/plabayo/rama/pull/914) has been merged now and should keep / transfer all needed extensions. Going to close this PR, but let me know if you think something still needs to happen or is broken.

PR #912, opened 2026-05-16
rama-dns improvements: vendor rfc3596.txt + improvements in dns resolvers

GlenDC:
- negative lookup caching
- improved response buffers on linux
- and some more minor improvements


soundofspace:
@GlenDC nice to see this one, was just going to open issue for `const DEFAULT_CACHE_TTL: Duration = Duration::from_secs(5 * 60);` being too long and restrictive. Supporting TTL on dns entries fixes that (mostly).

Having access to up to date IPs for dns is something I will need for loadbalancing, by support TTL on records that should now work as expected. Might still be surprising if users dont set this low enough, but that is expected default dns behaviour so should be fine.

PR #914, opened 2026-05-17
Improve extensions around Upgrades

soundofspace:
- Upgraded should only transfer connection extensions. If other extensions need to be transferred it's the responsibility of handle_upgrade users
- clone_to / clone_to_if_absent methods to transfer extensions. Mainly used in connectors to transfer extensions that become part of the connection to the connection extensions. This now also replaces the need for Chainable extensions
- Use clone_to_if_absent in all connectors to properly transfer extensions to connections where needed (things like config that only apply on connection setup are not transferred eg TLS config is not transferred since all info is in EstablishedTls struct, we could change this if we think this would be valuable / more consistent)
- Move UpgradeHandler to rama-http so we can use it everywhere
- UpgradeResponse instead of (Request, Output) tuple so its a typed struct, and also add Extensions to this so it becomes explicit which extensions should be transferred over the Upgraded


rama-http/src/layer/version_adapter/request.rs
```diff
@@ -3,7 +3,7 @@ use rama_core::Service;
 use rama_core::bytes::BytesMut;
 use rama_core::error::BoxError;
 use rama_core::error::ErrorContext;
-use rama_core::extensions::ChainableExtensions;
+// use rama_core::extensions::ChainableExtensions;
```
GlenDC:
```suggestion
```

?

src/http/mod.rs
GlenDC: if you want to do this you might as well just add `layer` to the above imports and drop this entire module...

that said, why? Don't think I have a strong opinion either way, just wonder....

GlenDC (approved):
LGTM, just some minor remarks. Once you resolve these, please do merge.

PR #922, opened 2026-05-21
Switch to res_nsearch

soundofspace:
Switch from `res_nquery` to `res_nsearch`:
- `res_nquery` only does fqdn lookups
- `res_nsearch` respects "search" and "ndots" in resolv.conf

The latter is consistent with `getaddrinfo` and `HickoryDnsResolver::try_new_system()` and will be drop in replacement. For fqdn one can still just use a domain ending in "."


GlenDC (approved):
Easy peasy

PR #923, opened 2026-05-22
dns loadbalancer

soundofspace:
Dns loadbalancer that resolves IPs (4, 6 or both) -> caches them -> uses IpPicker trait to choose one for every request.
- Refreshing mostly happens in the background to keep request path fast and to dedup queries
- Can be used for round-robin loadbalancing or more specialized routing
- Uses ConnectorTarget to configure the target -> works with tcp connector + pool is aware of this
- Each cached item get an Extensions store for local state. By doing it this way we link state with the item that needs state. This make it easier for IpPicker's to keep track of state without duplicating all state to an internal Dashmap or something + the burder to keep it up to date / TTL.... This could be a useful pattern to settle on in other place where we have a similar need for state linked to specific items


rama-dns/src/client/lb/cache.rs
```diff
@@ -0,0 +1,362 @@
...
+    fn spawn_refresh(self: Arc<Self>, host: Domain) {
+        let acquired = self.refreshing.lock().insert(host.clone());
+        // already refreshing, dedup
+        if !acquired {
+            return;
+        }
+
+        tokio::spawn(async move {
+            let result = self.resolve(&host).await;
+            self.refreshing.lock().remove(&host);
```
GlenDC: if the above statement panics, you never remove this host.... which in a way is a memory leak... Might need some guard? Defensive programmign wise. or a different design all together

GlenDC: I think I saw a similar pattern in one other loc.

rama-dns/src/client/lb/cache.rs
```diff
@@ -0,0 +1,362 @@
...
+            }
+        });
+    }
+
+    async fn resolve(&self, host: &Domain) -> Result<NonEmptyVec<IpAddr>, BoxError> {
+        let mut ips = Vec::new();
+        match self.mode {
+            DnsResolveIpMode::SingleIpV4 => self.collect_v4(host, &mut ips).await,
+            DnsResolveIpMode::SingleIpV6 => self.collect_v6(host, &mut ips).await,
+            DnsResolveIpMode::Dual | DnsResolveIpMode::DualPreferIpV4 => {
```
GlenDC: you are treathing them the same here. even though they are meant to have different behaviour (see other locations)

rama-dns/src/client/lb/picker.rs
```diff
@@ -0,0 +1,172 @@
...
+        // It is possible that resolution.ips has changed order because of dns refreshes.
+        // We dont consider that a problem for this RoundRobinPicker, but if you need accurate
+        // round robin, then you will also have to store a sorted listed of ips and keep track of this.
+        Ok(Some(resolution.ips[idx]))
+    }
+}
+
+#[derive(Debug, Clone, Copy, Default)]
+/// Random IP picker
+pub struct RandomPicker;
```
GlenDC, line 83: non-exhaustive

rama-dns/src/client/lb/picker.rs
```diff
@@ -0,0 +1,172 @@
...
+    }
+}
+
+impl DnsIpPicker for RandomPicker {
+    fn pick(
+        &self,
+        _host: &Domain,
+        resolution: &HostResolution,
+    ) -> Result<Option<IpAddr>, BoxError> {
+        let mut rng: SmallRng = rand::make_rng();
```
GlenDC: not entirely your fault as I also do it in some locs, but I think we migth want to move stuff like this to a thread local...

rama-dns/src/client/lb/cache.rs
```diff
@@ -0,0 +1,362 @@
...
+                // Stale but still usable: serve while refreshing in background.
+                self.clone().spawn_refresh(host.clone());
+                return Ok(entry);
+            }
+            // Too stale (this only happens if refresh is not working)
+            self.entries.invalidate(host);
+        }
+
+        // Initial lookup is (async) blocking and may double resolve (this should be cheap and rare)
+        let ips = self.resolve(host).await?;
```
GlenDC: this does mean, even if rare, that we can get an overwrite... not better to check if it exists or not? In case of such a rare double resolve?

GlenDC (changes requested):
Looks good in general. Some clever bits in there. Besides my comments I think you can improve your test coverage a bit, notabely:

* evict_after_idle path (only evict_after_stale is covered AFAIK)
* Refresh failure keeps the stale entry 
* Dual mode resolution
* serve's ProxyAddress skip path
* Picker returning None 
* Concurrent cold lookups

rama-dns/src/client/lb/cache.rs
```diff
@@ -0,0 +1,362 @@
...
+use super::HostResolution;
+
+/// Inmemory cache of resolved IPs per host.
+pub(super) struct DnsLbCache<R> {
+    resolver: R,
+    refresh_after: Duration,
+    evict_after_stale: Duration,
+    mode: DnsResolveIpMode,
+    entries: Cache<Domain, HostResolution, ahash::RandomState>,
+    refreshing: Mutex<HashSet<Domain>>,
```
GlenDC: this is essentailly a global lock... is this really required, and if so, no way to do it more clever? Perhaps even fold it in your existing cache or something similar?

GlenDC:
Also wondering. How do you see people using this? I don't see an example here that is migrated to this for example or a dedicated example. Is this useful in general or does it have specific use cases? If it's always a good idea might as well become the default behaviour of the default global cache given that is what most people end up using, but not sure that is indeed the case.

soundofspace:
> Also wondering. How do you see people using this? I don't see an example here that is migrated to this for example or a dedicated example. Is this useful in general or does it have specific use cases? If it's always a good idea might as well become the default behaviour of the default global cache given that is what most people end up using, but not sure that is indeed the case.

Added an example / special test for this, but in general I think our grpc examples will need to be restructured. In my opinion: they have a weird structure right now and are quite different then our normal examples. GRPC is also different but in general we could use some more "production" like examples, they could even spin up server/client in a single example eg to also shows like tcp_nodelay. I will give it some thought how could do this better, but if you have ideas, let me know

GlenDC:
> in general I think our grpc examples will need to be restructured
 
 Fine by me, do propose when you have concrete ideas.
 
 For now they were just mostly lifted as-is from tonic. So indeed a bit different from our own ones.

GlenDC (approved):
Thx! Looks good!

PR #940, opened 2026-05-29
sync w/ tower-rs/tower-http

GlenDC:
from tower-rs/tower-http@7551a9b8b9706ca1e11c035659b243f688b136bd to tower-rs/tower-http@b5366318c0661693e15de6d75e15a2f7924b5036


GlenDC:
FYI @soundofspace this also touches stuff you did regarding span parent, so do double check if this is desired, but seems to me like it is correct

soundofspace:
Just checked, and all looks good

PR #956, opened 2026-06-03
Extensions get_many in a single pass

soundofspace:
- get_many_ref and get_many_arc to efficiently get a tuple of Extension's
- FromExtensions macro to also easily implement this for a struct of references or arcs (or mix)
- Export crate as rama_core -> macro's also work in doc examples now, pattern that might be useful in other places also, @GlenDC unless you see something wrong with doing it this way?

This will be used for to construct TLS config from individual extension's. Basically it's just a nicer and more efficient way to get multiple extensions.


rama-core/src/lib.rs
```diff
@@ -23,6 +23,12 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![cfg_attr(test, allow(clippy::float_cmp))]
 
+// Lets `::rama_core` resolve to this crate from within itself, so the
+// `Extension` / `FromExtensions` derives (which emit `::rama_core::...`) work in
+// our own lib code, unit tests and doctests alike. `crate::...` would break
+// in doctests, which compile as a separate crate.
+extern crate self as rama_core;
```
GlenDC: you can avoid this by giving priority to the invidual crate. E.g. look first for rama-core and only than for rama. See: `rama-http-macros/src/root.rs`

GlenDC (changes requested):
Besides my comment and what discussed offline few non-blocking nits:

- `support_root_ts` is now duplicated in `extension.rs` + `from_extensions_macro.rs` (and `root.rs` is a 3rd near-copy). Could be one shared helper (also update comment to reflect what is actually preferred)
- the `const SEALED` trick works, but a sealed supertrait reads more idiomatically.
- the derive silently accepts `&mut T` / `Arc<T, A>` and only errors downstream. A clean macro diagnostic would be nicer.

GlenDC (approved):

rama-macros/src/utils.rs
```diff
@@ -0,0 +1,51 @@
...
+        return if let Some(module) = umbrella_module {
+            let module = Ident::new(module, Span::call_site());
+            quote!(::#ident::#module)
+        } else {
+            quote!(::#ident)
+        };
+    }
+
+    // Fall back to the individual_crate_name crate directly (eg rama-core)
+    if let Ok(found) = crate_name(individual_crate_name) {
```
GlenDC: prefer the invidiual crate name please, I used to think better other way around but there is no advantage and switching it opens some possibility, without a conn

PR #961, opened 2026-06-05
Refactor rustls TLS config into extension pieces

soundofspace:
Tls:
- Extensions for all common TLS config in rama-net -> these will work with both Boring and Rustls, making the common/easy use case very easy to switch between backends
- TlsClientConfig builder to easily build extension bag containing all TLS settings. Can be used as a base config, or transfer to other extensions such as request extensions
- Builder exposes setters to easily configure these
- rama-tls-rustls (and boring soon) have their own custom Extensions for custom logic (rustls only has one but boring will have way more). They also have an `Ext` trait that extends TlsClientConfig to add setters, this makes TlsClientConfig the one stop shop to configure anything TLS related and can be backend specific if needed
- Use FromExtensions to easily get all tls extensions and use that to build the final TLS connector data
- Rustls also has escape hatch that allows us to customize the final rustls config if needed or completely replace it for cases currently not supported by rama primitives
- When building EasyHttpWebClient with a custom connector, force TLS settings. Easy now with TlsClientConfig::default(). Before when using None this was a huge footgun since it didnt configure TLS alpn

Crypto
- Initial logic for shared certificate parsing -> will be extended for all things boring needs, and will do some cleanup around `DataEncoding`

Followup work (1 or more PRs):
- Migrate boring to TlsConfig and small extensions
- Extend crypto certificate logic
- Think if we also want to use a similar approach on the server side, eg TlsServerConfig. I would think yes, or at least the common api which is nice to work with


examples/mtls_tunnel_and_service.rs
```diff
@@ -88,15 +95,18 @@ async fn main() {
...
-                .build();
+        let tls_client_data = TlsClientConfig::default()
+            .with_client_auth(ClientAuth::Single(ClientAuthData {
+                cert_chain: DataEncoding::DerStack(
+                    client_cert_chain
+                        .iter()
+                        .map(|cert| cert.as_ref().to_vec())
+                        .collect(),
+                ),
+                private_key: DataEncoding::Der(client_priv_key.secret_der().to_vec()),
```
GlenDC, line 106: this is still using DataEncoding etc, so that is something you want to replace in a follow up PR I suppose?

rama-crypto/src/cert.rs
```diff
@@ -0,0 +1,19 @@
+use crate::dep::pki_types::pem::PemObject;
+use crate::dep::pki_types::{CertificateDer, PrivateKeyDer};
+use rama_core::error::{BoxError, ErrorContext};
+
+// TODO move all certificate parsing / utils in here
+
+/// Parse a PEM buffer into a DER certificate chain
+pub fn pem_to_cert_chain(pem: &[u8]) -> Result<Vec<CertificateDer<'static>>, BoxError> {
```
GlenDC: What's the advantage of these 2 utilities over just calling the PrivateKeyDer/CertificateDer API directly?

rama-net/src/tls/client/ext.rs
```diff
@@ -0,0 +1,250 @@
...
+use super::{ClientAuth, ServerVerifyMode};
+
+/// A backend agnostic builder for the common TLS configs.
+///
+/// It holds a set of fine grained config extensions (e.g. [`TlsAlpn`], [`TlsServerVerify`])
+/// and exposes typed setters for the settings both TLS backends support.
+/// Backend crates add setters for their backend-specific pieces via extension
+/// traits (`RustlsClientConfigExt`).
+#[derive(Debug, Clone)]
+pub struct TlsClientConfig(Extensions);
```
GlenDC, line 16: Bit puzzled why you also want this wrapper to exist explicitly instead of only conceptually? What if someone inerts ALPN extension directly in extensions themselves? Does that work? What is the expected order in that case?

GlenDC: Not easier if we drop TlsClientConfig as an extension and only have them in extensions? You could still have TLsClientConfig as a struct that can be created from _many_ extension references, but not sure why you would also want it to be something that is an extension itself or even holds extensions?

rama-tls-rustls/src/client/config.rs
```diff
@@ -0,0 +1,75 @@
+use crate::dep::rustls::ClientConfig;
+use rama_core::error::BoxError;
+use rama_core::extensions::{Extension, FromExtensions};
+use rama_net::tls::client::{
+    TlsAlpn, TlsClientAuth, TlsClientConfig, TlsKeyLog, TlsServerName, TlsServerVerify,
+    TlsStoreServerCertChain, TlsSupportedVersions,
+};
+
+/// Gather all the TLS extensions supported by rustls
+#[derive(FromExtensions)]
```
GlenDC, line 12: ah nice to see this in action ,love it. With this in place I am even more puzzled wabout why we still need TlsClientConfig o.o

rama-tls-rustls/src/client/connector.rs
```diff
@@ -13,12 +16,14 @@ use rama_net::address::Host;
 use rama_net::client::{ConnectorService, EstablishedClientConnection};
 use rama_net::extensions::StreamTransformed;
 use rama_net::tls::ApplicationProtocol;
-use rama_net::tls::client::NegotiatedTlsParameters;
+#[cfg(feature = "http")]
+use rama_net::tls::client::TlsAlpn;
```
GlenDC, line 20: why is this http gated? Alpn can be for non-http protocols as well?

rama-tls-rustls/src/client/connector.rs
```diff
@@ -426,33 +437,24 @@ impl<S, K> TlsConnector<S, K> {
...
+        "override TLS ALPN to match TargetHttpVersion",
     );
 
-    Ok(Some(
-        connector_data
-            .map(Ok)
-            .unwrap_or_else(TlsConnectorData::try_new)?
-            .with_alpn_protocols(&[target_alpn]),
-    ))
+    ext.insert(TlsAlpn(vec![target_alpn]));
```
GlenDC, line 456: function used to be pure, now it mutates, is that expected and if so? Will it always be fine to do so?

rama-tls-rustls/src/client/connector_data.rs
```diff
@@ -1,271 +1,121 @@
...
-        /// Set alpn protocols to only advertise [`ApplicationProtocol::HTTP_2`].
-        pub fn alpn_protocols_http_2(mut self) -> Self {
-            self.set_alpn_protocols(&[ApplicationProtocol::HTTP_2]);
-            self
-        }
-    }
-
-    rama_utils::macros::generate_set_and_with! {
-        /// Set certificate verifier that will be used to verify certs
-        pub fn cert_verifier(mut self, verifier: Arc<dyn ServerCertVerifier>) -> Self {
```
GlenDC, line 227: we  no longer seem to easily allow using custom verifiers, is this expected? There still seems to be a way but not as easy as it used to be?

GlenDC (changes requested):
Looks good, just a couple of questions and nits.

examples/mtls_tunnel_and_service.rs
soundofspace: Yes, but will do this in followup since the proper refactor I want to do also touches boring code

rama-net/src/tls/client/ext.rs
soundofspace: So TlsClientConfig is actually not what we store in Extensions. It's basically a builder that makes it easy to configure TLS options with typed setters. I could expose these setters with an Ext trait on Extensions directly but that already has so many methods that i wanted to isolate this. So it works like this for these reasons:
- TlsClientConfig has methods for all common TLS configs, and is extended with Ext traits for each backend. This single struct groups all setters basically
- It stores Extensions internally so it supports all the individual config pieces even the ones that are backend specific
- Backends can store this config as the "base" config eg passed in EasyHttpWebClient
- TlsClientConfig also has a method to transfer these extensions to a request eg `tls_config.write_to(request.extensions())`. This is just to make it easier to group all setters and provide a better user experience, but nothing is stopping you/us from just setting the individual pieces directly in request extensions
- Tls backends do merge the "base" config with the ones from extensions like this `input.extensions().with_base(base.as_extensions())`. From here on out we are working with just Extensions again so everything we expect to work with extensions just works again. Config in from request extensions has priority since Extensions are stacked like this, and fallbacks just kick in naturally because of how we iterate extensions from newest (request) to oldest (base_config)

This could be slightly changed in how expose this and how we store this. Eg instead of builder put this behind Ext trait and dont store TlsClientConfig as base config in Connector but instead store base_extensions. However I think having TlsClientConfig type makes this cleaner, but really curious about how you would change this?

Basically the goal is an easy way to set a base config with typed setters and easily merge this in connector

rama-tls-rustls/src/client/connector.rs
soundofspace: Yes but we are importing alpn here only to configure it is TargetHttpVersion is set which is http a feature

rama-tls-rustls/src/client/connector.rs
soundofspace: I think it makes more sense if we do an override like this to also store it in extensions to reflect this and so extensions contains the real state. And yes this is fine todo since we actually already did it, but less clear

rama-tls-rustls/src/client/connector_data.rs
soundofspace: No forgot to add this one, going todo it now

rama-net/src/tls/client/ext.rs
GlenDC: Inserting extensions isn't difficult either. So only thing I can think is that you are worried about discoverability, but that is what those extension tags are bout, no? So still not entirely sure what is the problem being solved here. 

That said I do understand where you are coming from and it read reasonable. I just wonder if it's really all that much or required.

rama-tls-rustls/src/client/connector.rs
GlenDC: Ok soon rama-net should no longer have a http feature, will see what that means for this.

rama-tls-rustls/src/client/connector.rs
GlenDC: ok gotcha

rama-crypto/src/cert.rs
soundofspace: Preparing for the future, this will grow and boring will also rely on this

rama-net/src/tls/client/ext.rs
soundofspace: Not required, but lets say we remove this how would you configure TLS config once on connector level. We dont want to be inserting extensions in each request with the config?

rama-tls-rustls/src/client/connector.rs
soundofspace: even better

rama-net/src/tls/client/ext.rs
GlenDC: Not sure? How do you do it know without doing that? And what if someone does insert those extensions in between the connector and using it?

soundofspace: Well thats exactly what this config solves, to make that easy, or we are back to were before.
What do you mean with "does insert those extensions in between the connector and using it"?

soundofspace: Eg a common use case case which maybe makes this clearer.
- Create `TlsClientConfig::default()` -> this has H2/http1.1 enabled
- Create EasyHttpWebclient with custom connector using this config
- Do a request
- Tls connector use `req.extensions().with_base(base_config)` to build config
- Config fetches all pieces including ALPN. We go over all extensions to find ALPN, we will find ALPN all the way at the bottom in base_config -> alpn = H2/http1.1

With override
- Do a second request but on this request with do `request.extensions().insert(TlsAlpn(vec![H2]))`
- Or `TlsClientConfig::new().with_http2().write_to(request.extensions())`.
- Reuse the same client from before
- Tls connector uses `req.extensions().with_base(base_config)` to build config
- We again go over all extensions, and since request extensions are the first ones we iterate we see ALPN = H2
- We use H2

The nice thing about this setup is that the logic that builds the config doesn't know/care where the extensions come from. It just goes over them from top to bottom until it finds what it needs, its just looks like one long chain of extensions.

soundofspace: If we dont want this global struct (which i think we do), we could also just use extensions directly. And give connectors their own "base" extensions instead. This would work the same then but without the wrapper
eg
```rust
/// Define the base [`Extensions`] for this [`TlsConnector`].
///
/// Per connection pieces inserted in the request's extensions are layered
/// on top of this base (newest-wins).
pub fn base_extensions(mut self, base: Option<Extensions>) -> Self {
    self.base_extensions = base;
    self
}
```

```
let tls_extensions = Extensions::new();
tls_extensions.insert(TlsAlpn::h2());
tls_extensions.insert(TlsKeylogIntent::Env());

let client = EasyHttp.connector_builder().with_tcp().with_tls(tls_extensions)...;
```

So yea this global struct just makes it easier to read (its a config) and groups everything. Which makes it much nicer to work with I think instead of searching all the individual pieces.

rama-crypto/src/cert.rs
GlenDC: Sure but abstractions need a good trade off and I feel the abstraction is very shallow here to the point that I feel it might be better to just have people use the logic directly? Including us. As in those types + the method(s) on it? 

For example maybe someone wants an iterator of results etc... or some filtering in between or this or that, all easily possible without this thin abstraction, but now hidden away, to buy us... what?

rama-net/src/tls/client/ext.rs
GlenDC: > Create TlsClientConfig::default() -> this has H2/http1.1 enabled

In a way that's already a trap though, as depending you as a rama user it might not occur to you that default means advertising http... you might not be an http user at all... fine to have that as `http_auto` as a constructor or something like that but I don't think we should claim that default for tls is about http. Wdyt?

--

FYI introducing that `with_base` thing might feel nice but it does make extensions less easy to understand though. As so far everything in rama you can read and understand the flow correctly... even for extensions things can only eve be appended... You could say that is not so for ingress/egres, or connecitons etc, but don't agree as even there it is clear parent-wise etc...

Introducing `Tls connector use req.extensions().with_base(base_config)` makes it that now you need to read all bodies to really know if something is first or not, as with_base can be called at any time adding suddenly stuff to front. So I wonder if it is really worth it..

--

I just still do not really see what we are missing by just having those invidiual extensions, given at usage time you already ahve your aggregator by reference collecting all possibilities...

GlenDC: nvm part about base config etc.... just make sure default tls client config doesn't enable http-specific things (e.g. ALPNs and all is good)

GlenDC (approved):
LGTM, only concern left over already discussed offline. Once you ensured that shared tls client config is no longer (by default) http specific, you can ship it.

soundofspace:
@GlenDC is integrating pki-types like this good (nested in dep) or do we move it up one level so it becomes `rama-crypto::pki_types` vs `rama-crypto::dep::pki_types`?

GlenDC:
> @GlenDC is integrating pki-types like this good (nested in dep) or do we move it up one level so it becomes `rama-crypto::pki_types` vs `rama-crypto::dep::pki_types`?

If we plan to make it prominent first class types in the API of our crates it should no longer be under `dep`.

And if ever blocking in the future we can fork it in place. For now hope is it can be a stable dep.

PR #962, opened 2026-06-07
support rss-podcast alternate + strong typed Uris (rss/atom)

GlenDC:


GlenDC (commented):
looks good, there's one improvement i'll see but i'll fix that when I integrate in my other branch

PR #966, opened 2026-06-10
tls: default rama tls clients to native system trust store

GlenDC:
New feature-gated rama-crypto native_certs module
(forked rustls-native-certs) loads the platform trust store once into shared pki_types anchors; rustls and boring both verify against it by default, falling back to bundled webpki roots (with a warning) when no native roots are found.

Note that this work is unrelated to a platform verifier, where we make use of the platform sdk (windows/linux/macos/...) to verify certificates. Advantage of the latter isd that it also works in contained environments (e.g. system extensions on MacOS) and always have the latest view of certificates. In contrast these certificates are static and require a cold start (to load them).

Regardless, it's a good default. And works for majority of use cases just fine.
That said I do think one of these weeks either I or you @soundofspace need to look into integrating code similar to 'platform-verifier' as an opt-in feature (just not sure where) and when enabled have that be the default for rustls/boring tls (client) connectors.

### DIFF
--- a/rama-crypto/src/native_certs/mod.rs
+++ b/rama-crypto/src/native_certs/mod.rs
@@ -1,0 +1,436 @@
+//! Load the platform's native certificate store (system trust chain) in a
+//! tls-implementation agnostic way, as [`pki_types`] certificates.
+//!
+//! The certificates returned here can be fed into any tls backend (e.g.
+//! `rustls` or `boring`), which is why this lives in `rama-crypto` rather than
+//! in one of the tls backend crates.
+//!
+//! The main entry points are:
+//!
+//! - [`shared_native_trust_anchors`]: the cached, process-wide default trust
+//!   anchors used by rama tls clients. Loads the native store once; if nothing
+//!   is found it warns and falls back to the bundled webpki roots.
+//! - [`load_native_certs`]: a one-shot (uncached) read of the platform store,
+//!   for callers that want to manage caching/merging themselves.
+//! - [`bundled_root_certs`]: the bundled Mozilla (CCADB) root certificates used
+//!   as the fallback.
+//!
+//! # Attribution
+//!
+//! The platform readers and `SSL_CERT_FILE`/`SSL_CERT_DIR` handling are an
+//! adapted fork of [`rustls-native-certs`] (Apache-2.0 OR ISC OR MIT), with the
+//! pending [permission-skip fix][pr228] folded in and the public surface
+//! reshaped around rama's [`pki_types`] re-export, error and tracing
+//! conventions.
+//!
+//! [`pki_types`]: crate::pki_types
+//! [`rustls-native-certs`]: https://github.com/rustls/rustls-native-certs
+//! [pr228]: https://github.com/rustls/rustls-native-certs/pull/228
+
+use std::error::Error as StdError;
+use std::path::{Path, PathBuf};
+use std::sync::{Arc, OnceLock};
+use std::{env, fmt, fs, io};
+
+use rama_core::telemetry::tracing::{debug, warn};
+
+use crate::pki_types::CertificateDer;
+use crate::pki_types::pem::{self, PemObject};
+
+#[cfg(all(unix, not(target_os = "macos")))]
+mod unix;
+#[cfg(all(unix, not(target_os = "macos")))]
+use unix as platform;
+
+#[cfg(windows)]
+mod windows;
+#[cfg(windows)]
+use windows as platform;
+
+#[cfg(target_os = "macos")]
+mod macos;
+#[cfg(target_os = "macos")]
+use macos as platform;
+
+/// Returns the cached, process-wide default trust anchors used by rama tls
+/// clients (both the `rustls` and `boring` backends consume these).
+///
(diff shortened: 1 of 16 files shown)


rama-tls-boring/src/client/connector_data.rs
```diff
@@ -736,106 +734,46 @@ fn shared_default_verify_store() -> Result<&'static X509Store, BoxError> {
...
-    );
+    trace!("boring connector: shared verify store built: added {added} anchor(s), {failed} failed");
 
-    if total_added_cert_count == 0 {
+    if added == 0 {
         return Err(OpaqueError::from_static_str(
-            "failed to add windows certs from system (user/machine x Root/CA)",
+            "no native trust anchors could be added to the boring x509 verify store",
         )
         .into_box_error());
```
soundofspace, line 778: Is there still argument for using OpaqueError::from_static_str + into_box_error vs just BoxError::from("")?

rama-tls-boring/src/client/connector_data.rs
```diff
@@ -736,106 +734,46 @@ fn shared_default_verify_store() -> Result<&'static X509Store, BoxError> {
...
+                failed += 1;
+                debug!("boring connector: failed to parse native trust anchor as x509: {err}");
             }
         }
     }
 
-    trace!(
-        "boring connector: windows: final result: added {total_added_cert_count} certs of {total_cert_count} certs"
-    );
+    trace!("boring connector: shared verify store built: added {added} anchor(s), {failed} failed");
```
soundofspace: NIT, but rustls is structured logging and here its free text. Going forward do we want to commit to structured logging, or not really? Personally i prefer structured, and depending on where logs are exported it could potentially save string formatting
```
        rama_core::telemetry::tracing::trace!(
                added,
                ignored,
                "rama-tls-rustls: initialised client root cert store from shared native trust anchors"
            );
```

rama-crypto/src/native_certs/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+/// native store yields no certificates, a warning is logged and the bundled
+/// webpki roots ([`bundled_root_certs`]) are used instead so that clients on
+/// minimal systems (e.g. distroless containers) still have a sane default.
+///
+/// The result is cached for the lifetime of the process: the (potentially
+/// expensive) native read happens at most once.
+pub fn shared_native_trust_anchors() -> Arc<[CertificateDer<'static>]> {
+    static ANCHORS: OnceLock<Arc<[CertificateDer<'static>]>> = OnceLock::new();
+    ANCHORS
+        .get_or_init(|| {
```
soundofspace: LazyLock seems better here

rama-tls-rustls/src/client/connector_data.rs
```diff
@@ -129,12 +129,25 @@ fn rustls_client_auth(
...
 
+/// The default client root certificate store used to verify servers.
+///
+/// By default this is built from the platform's native trust store (the system
+/// root certificates), loaded once and shared process-wide via
+/// [`rama_crypto::native_certs::shared_native_trust_anchors`]. On systems where
+/// no native roots are found, that loader warns and falls back to the bundled
+/// webpki (Mozilla CCADB) roots.
 pub fn client_root_certs() -> Arc<RootCertStore> {
     static ROOT_CERTS: OnceLock<Arc<RootCertStore>> = OnceLock::new();
```
soundofspace: Also lazylock?

rama-crypto/src/native_certs/mod.rs
```diff
@@ -0,0 +1,435 @@
...
+            let paths = CertPaths::from_env();
+            let result = load_native_certs_with_paths(&paths);
+            for err in &result.errors {
+                debug!("rama native-certs: error while loading native root certificate: {err}");
```
soundofspace:
```suggestion
                debug!(%err, "rama native-certs: error while loading native root certificate");
```

soundofspace (changes requested):
All nit stuff, but something we probably want to align on

rama-tls-boring/src/client/connector_data.rs
GlenDC: afaik a String allocation is still larger than a Box allocation, so yes? Would be cleaner if they didn't string-alloc a static str in std lib... but oh well? Feel free to push back if I'm wrong by re-opening it.

rama-crypto/src/native_certs/mod.rs
GlenDC: done

rama-tls-rustls/src/client/connector_data.rs
GlenDC: done

rama-crypto/src/native_certs/mod.rs
GlenDC: done

rama-tls-boring/src/client/connector_data.rs
GlenDC: done

soundofspace (approved):

PR #967, opened 2026-06-10
BoxErrorExt trait to create BoxError from_static_str

soundofspace:
- shadow into_box_error with a dedicated one for opaqueError to prevent to double boxing
- BoxErrorExt trait which we can use to easily created special boxedErrors: for now only from_static_str, but could be extended

### DIFF
--- a/examples/gateway/fastcgi-php/migration/main.rs
+++ b/examples/gateway/fastcgi-php/migration/main.rs
@@ -191,8 +191,10 @@ mod unix_impl {
 use unix_impl::run;
 
 #[cfg(not(target_family = "unix"))]
-async fn run() -> Result<(), rama::error::extra::OpaqueError> {
-    Err(rama::error::extra::OpaqueError::from_static_str(
+async fn run() -> Result<(), rama::error::BoxError> {
+    use rama::error::BoxErrorExt;
+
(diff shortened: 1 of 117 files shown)


examples/gateway/fastcgi-php/migration/main.rs
```diff
@@ -191,8 +191,10 @@ mod unix_impl {
 use unix_impl::run;
 
 #[cfg(not(target_family = "unix"))]
-async fn run() -> Result<(), rama::error::extra::OpaqueError> {
-    Err(rama::error::extra::OpaqueError::from_static_str(
+async fn run() -> Result<(), rama::error::BoxError> {
+    use rama::error::BoxErrorExt;
+
+    Err(rama::error::BoxError::from_static_str(
```
GlenDC, line 197: I wonder if it might not make sense to provide instead of this ext just a function `rama::error::static_str_error`... or something like that with a different name, but mostly talking about the same name... not sure if a Trait Ext helps here much?

GlenDC (approved):

examples/gateway/fastcgi-php/migration/main.rs
GlenDC: Discussed offline, fine approach as-is.

PR #970, opened 2026-06-11
Refactor boring tls to TlsClientConfig and remove legacy ClientConfig

soundofspace:
- Refactor boring to new TlsClientConfig
- rama-ua store client hello instead of intermediate (legacy) ClientConfig
- rama-ua slightly restructure TlsProfile so it lines up with extensions.insert_arc/get_arc


examples/acme_http_challenge.rs
```diff
@@ -72,7 +72,7 @@ use rama::{
     layer::ConsumeErrLayer,
     net::tls::{
         DataEncoding,
-        client::ServerVerifyMode,
+        client::{ServerVerifyMode, TlsClientConfig},
```
GlenDC, line 75: not something new in this PR, but your work here triggered me to recheck `TlsClientConfig` as I remember it wraps `Exensions` yet you derive `Clone`. That means if you clone this you have not actually cloned tlsClientConfig you just have 2 instances to the same thing. Shouldn't that Clone be manually implemented so it doesn't clone underlying extensions? Seems like a footgun waiting to happen...

rama-net/src/tls/client/config.rs
```diff
@@ -1,73 +1,202 @@
...
+    /// Transfer this config's pieces onto `extensions` (appending, so they
+    /// override existing entries of the same type — newest-wins). Use this to
+    /// transfer the tls config to e.g. request extensions
+    pub fn write_to(&self, extensions: &Extensions) {
+        extensions.extend(&self.0);
+    }
+
+    generate_set_and_with! {
+        /// Set the ALPN protocols to offer.
+        pub fn alpn(mut self, protocols: Vec<ApplicationProtocol>) -> Self {
```
GlenDC: not introduced by this pr, but wouldn't a NonEmptySmallVec of cap 2 be a better fir for this? I doubt it is often more than 2 protocols?

rama-ua/src/profile/tls.rs
```diff
@@ -1,29 +1,22 @@
...
-    tls::{
-        ProtocolVersion,
-        client::{ClientConfig, ClientHello, ServerVerifyMode},
-    },
+    tls::{ProtocolVersion, client::ClientHello},
 };
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, Extension)]
+#[derive(Debug, Clone, Extension, Serialize, Deserialize)]
```
GlenDC, line 10: not sure if we ever used it, too long ago since I worked on this, but you do drop the `insecure` property here? Not gonna cause issues in existing flows in rama-fp? Again maybe was never used, but just saying it's a difference

rama-tls-boring/src/client/emulate_ua.rs
```diff
@@ -44,108 +43,35 @@ where
...
     async fn serve(&self, input: Input) -> Result<Self::Output, Self::Error> {
-        let tls_profile = input.extensions().get_ref::<TlsProfile>().cloned();
-
-        // Right now this is very simple, but it will get a lot more complex, which is why it is separated from the connector itself
-        if let Some(profile) = tls_profile {
-            let mut domain_overwrite = None;
-            let mut emulate_config = Cow::Borrowed(profile.client_config.as_ref());
+        if let Some(profile) = input.extensions().get_arc::<TlsProfile>() {
+            // Create a config for for the provided client_hello
+            // TODO we could cache this in the future
```
GlenDC, line 48: what stops us from doing it already? and why would we want to do so?

GlenDC (approved):
Looks good, great even. Just a couple of nits/questions.

rama-ua/src/profile/tls.rs
soundofspace: As far as i could tell, we added insecure there, so we could easily set it when we tested / used these profiles. But with how this new config/extensions work its easy to set it where we actually use. So I removed it so it would say more true to the source of the profile, but can definetely add it again if needed, just think that this is cleaner unless we have a real need for it, which I didnt see now

rama-net/src/tls/client/config.rs
soundofspace: Chrome also adds h3 here when using quic together with the default ones so yea we could make this a smallvec, but i would not make it non empty. Eg right now you could overwrite a TlsProfile to set no alpn by giving it an empty list still

rama-tls-boring/src/client/emulate_ua.rs
soundofspace: This is me assuming that building this from a client hello might be expensive, but without doing any benchmarking here I dont want to really commit to caching. Caching here could be complex depending on the amount of discing client hello's. So without proof that we need it I can of left this here as an easy to find marker in case we do ever come across such evidence

examples/acme_http_challenge.rs
soundofspace: yea good safety

rama-tls-boring/src/client/emulate_ua.rs
GlenDC: ok fair

rama-ua/src/profile/tls.rs
GlenDC: fair

rama-net/src/tls/client/config.rs
GlenDC: yeah for quic we'll have to see how we'll do this, but fair, making it a regular SmallVec can work as well

PR #981, opened 2026-06-15
refactor(net): dedup ja4 hash12 helper into shared module

seuros:
Moved similar code to avoid duplication.


GlenDC (approved):
Nice catch! And nice to see you around!!

PR #987, opened 2026-06-15
refactor(net): dedup serde str (de)serialize boilerplate via macro

seuros:
Replace 11 hand-written serde::Serialize/Deserialize impl pairs across the address, forwarded, socket and uri modules with a single crate-local impl_serde_str! macro.


GlenDC (approved):
Nice! I didn't really mind the duplication, but appreciated none the less

PR #989, opened 2026-06-16
refactor(http): dedup HTTP layer/io boilerplate

seuros:
Removes duplicated logic across the http layers and io writers, found with AST
clone detection. No behavior change; rama-http checks clean with
`--all-features` and the relevant layer/io tests pass.

- compression/decompression body: share the identical `StreamingBody::poll_frame`
  and the eight `DecorateAsyncRead` impls (`compressed_body_poll_frame!` and
  `impl_decorate_async_read!` macros)
- macro-generate the three `SetSensitive*HeadersLayer` `new`/`from_shared`
  constructors (docs rebuilt per type via `concat!`/`stringify!`)
- macro the five sized-body `IntoResponse` impls (`Cow<str>`, `ArcStr`,
  `&ArcStr`, `Bytes`, `Cow<[u8]>`)
- share the HTTP/1 header- and body-writing blocks between `write_http_request`
  and `write_http_response` (`write_http1_header_map` / `write_http1_body`)
- dedup the unbounded/bounded `BidirectionalWriter` receive loop via a
  `drive_bidirectional_writer!` macro

### DIFF
--- a/rama-http/src/io/mod.rs
+++ b/rama-http/src/io/mod.rs
@@ -1,5 +1,12 @@
 //! http I/O utilities, e.g. writing http requests/responses in std http format.
 
+use crate::{Body, HeaderMap, StreamingBody, body::util::BodyExt};
+use rama_core::bytes::Bytes;
+use rama_core::error::{BoxError, ErrorContext as _};
+use rama_core::extensions::Extensions;
+use rama_http_types::{Version, proto::h1::Http1HeaderMap};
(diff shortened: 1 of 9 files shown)


GlenDC (approved):
Looks good. Thanks. One I pushed back on as I think the abstraction is not clear and makes the code harder to read. Some others are also not something i would have persuit myself, but they are reasonable enough

PR #991, opened 2026-06-16
refactor(otel): scope grpc to opt-in consumers under opentelemetry

seuros:
Make the opentelemetry feature enable grpc instrumentation only when a consumer opts into grpc, via rama-grpc?/opentelemetry alone. OTel for the core/http/net layers stays available without the grpc transport.

Gate telemetry::opentelemetry::collector behind the grpc feature so the re-exported rama_grpc OTLP exporter is present exactly when grpc is enabled, and the crate builds with opentelemetry on its own.

collector::OtelExporter is reachable with the grpc + opentelemetry feature combination.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -464,7 +464,6 @@ opentelemetry = [
     "rama-core/opentelemetry",
     "rama-http?/opentelemetry",
     "rama-net?/opentelemetry",
-    "grpc",
     "rama-grpc?/opentelemetry",
 ]
 # Optional [dial9] runtime-telemetry support across the rama crate
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -730,7 +730,7 @@ required-features = ["http-full"]
 
 [[example]]
 name = "http_telemetry"
-required-features = ["http-full", "opentelemetry"]
+required-features = ["http-full", "grpc", "opentelemetry"]
 
 [[example]]
 name = "http_user_agent_classifier"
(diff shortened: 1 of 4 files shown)


GlenDC (approved):

GlenDC:
@seuros I think you will need to add grpc feature opt-in in rama dep of rama-cli now

soundofspace (changes requested):
This will not work as is. Grpc is not only used for the transport but is also needed for the http protobuf exporter in OtelExporter. So we kind of need grpc also for the http exporter

GlenDC:
> This will not work as is. Grpc is not only used for the transport but is also needed for the http protobuf exporter in OtelExporter. So we kind of need grpc also for the http exporter

Correct, exportor http or grpc both use grpc. But I suppose that shouldn't mean we hardcode grpc enabled when otel is enabled, or am I missing something @soundofspace ?

soundofspace:
> > This will not work as is. Grpc is not only used for the transport but is also needed for the http protobuf exporter in OtelExporter. So we kind of need grpc also for the http exporter
> 
> Correct, exportor http or grpc both use grpc. But I suppose that shouldn't mean we hardcode grpc enabled when otel is enabled, or am I missing something @soundofspace ?

No just means it needs some more feature work eg in rama-cli, but probably also in examples

seuros:
I'm on it.

GlenDC:
still failing @seuros , you'll want to make sure to run all (full) QA checks locally, when making a contribution, makes it more pleasent for yourself than to have to ping-pong with a CI system

seuros:
I did run the full qa check.

But  CI additionally runs --run-ignored and --examples steps.  

Both PR should be green now.

GlenDC:
> But CI additionally runs --run-ignored and --examples steps.

Yes but there are also `just` targets for those :) Not saying to run them always, but in cases like this can help so you do not need to rely on CI multiple times

PR #992, opened 2026-06-16
refactor(grpc-otel): extract shared send_proto scaffolding

seuros:
The HTTP and gRPC OTLP transports duplicated the same send_proto control flow: a shutdown guard, a timeout-or-await match arm, and the `InternalFailure(err.to_string())` error mapping repeated ~12 times.

Introduce three shared helpers in the module root:
- internal_failure(err): map any Display error into InternalFailure
- await_with_timeout(timeout, fut): enforce the optional timeout and surface OTelSdkError::Timeout, shared by both transports
- ensure_not_shutdown(signal): the AlreadyShutdown guard, also reused by force_flush_signal

No behavior change; rama-grpc tests and clippy --all-features pass.

### DIFF
--- a/rama-grpc/src/service/opentelemetry/mod.rs
+++ b/rama-grpc/src/service/opentelemetry/mod.rs
@@ -108,6 +108,25 @@ pub trait OtlpTransport {
         Resp: Message + Default + Send + 'static;
 }
 
+/// Map any displayable error into an [`OTelSdkError::InternalFailure`].
+pub(super) fn internal_failure(err: impl fmt::Display) -> OTelSdkError {
+    OTelSdkError::InternalFailure(err.to_string())
+}
+
+/// Await `fut`, enforcing `timeout` when set and surfacing
+/// [`OTelSdkError::Timeout`] on expiry. Shared by the HTTP and gRPC transports.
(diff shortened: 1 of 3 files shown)


rama-grpc/src/service/opentelemetry/mod.rs
```diff
@@ -108,6 +108,25 @@ pub trait OtlpTransport {
...
 }
 
+/// Map any displayable error into an [`OTelSdkError::InternalFailure`].
+pub(super) fn internal_failure(err: impl fmt::Display) -> OTelSdkError {
+    OTelSdkError::InternalFailure(err.to_string())
+}
+
+/// Await `fut`, enforcing `timeout` when set and surfacing
+/// [`OTelSdkError::Timeout`] on expiry. Shared by the HTTP and gRPC transports.
+pub(super) async fn await_with_timeout<F: Future>(
```
GlenDC:
```suggestion
pub(super) async fn await_with_optional_timeout<F: Future>(
```

+ fixes in call sites

GlenDC (changes requested):
looks fine for me, just a single request for change

GlenDC (approved):

PR #1001, opened 2026-06-18
FromExtensions support rank and one of enum variant

soundofspace:
When getting multiple extensions support getting the rank (0=newest). This allows users to make informed decisions about mixed priority. One user of this logic is an Enum variant that basically allows you to give it an enum and it will give you the most recent variant of this enum.

I need this for TlsServerConfig extensions where we will have ServerAuth (static certs) and CertIssuer (boring specific). If both are specific we want to use the "newest" one, so we are consistent with how we use extensions everywhere. Adding rank unlocked this for me, but I'm kind of lurking around toasty aswel and seeing how they leverage types + traits + macros to make all of this nicer to use I also added enum support to FromExtensions. Basically for TLS in boring it will work like this:
```rust
#[derive(FromExtensions)]
pub struct BoringTlsAcceptorConfig<'a> {
    ....
    pub server_auth: Option<BoringServerAuth<'a>,
}

#[derive(FromExtensions)]
pub enum BoringServerAuth<'a> {
    CertIssuer(&'a BoringServerCertIssuer),
    StaticAuth(&'a TlsServerAuth)
}
```

instead of this, which would give questions like:
- What has priority?
- Lets assume we add rank here: we still need to look inside logic to see how we actually use...

But with enum its clear, its one of those options, and its always the newest one
```rust
#[derive(FromExtensions)]
pub struct BoringTlsAcceptorConfig<'a> {
    ....
    pub server_auth: Option<&'a TlsServerAuth>,
    pub cert_issuer: Option<&'a BoringServerCertIssuer>,
}

```

### DIFF
--- a/rama-macros/src/from_extensions_macro.rs
+++ b/rama-macros/src/from_extensions_macro.rs
@@ -262,6 +262,9 @@ fn expand_enum(
                 __offset: usize,
             ) -> ::core::option::Option<Self> {
                 let __candidates: [::core::option::Option<(Self, usize)>; #n] = [ #( #builders ),* ];
+                // `min_by_key` keeps the first of equal keys, so on a rank tie
+                // (only possible when two variants name the same type and thus
+                // resolve to the same entry) the earlier-declared variant wins.
                 ::core::iter::IntoIterator::into_iter(__candidates)
                     .flatten()
                     .min_by_key(|(_, __rank)| *__rank)
@@ -277,7 +280,10 @@ fn expand_enum(
             /// candidate type, lookup uses the same rule as `Extensions::get_ref`
             /// (newest-wins, walks wrappers and the parent chain). When several
             /// candidates are present the one with the lowest traversal rank
-            /// (newest) wins, returns `None` if none are present.
+            /// (newest) wins, returns `None` if none are present. If two variants
+            /// name the same type they resolve to the same entry (equal rank), the
+            /// tie is broken deterministically in favour of the earlier-declared
+            /// variant.
             pub fn from_extensions #fn_generics (
                 ext: & #lt #root::extensions::Extensions,
             ) -> ::core::option::Option<Self> {
--- a/rama-macros/src/from_extensions_macro.rs
+++ b/rama-macros/src/from_extensions_macro.rs
@@ -470,3 +470,210 @@ fn index_tuple(ty: &Type) -> Option<(&Type, &Type)> {
 fn is_usize(ty: &Type) -> bool {
     matches!(ty, Type::Path(p) if p.path.is_ident("usize"))
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use syn::{DeriveInput, parse_quote};
+
+    fn expand_err(input: DeriveInput) -> String {
+        expand(input)
+            .expect_err("expected expansion to fail")
+            .to_string()
+    }
+
+    fn expand_ok(input: DeriveInput) -> String {
+        expand(input)
+            .expect("expected expansion to succeed")
+            .to_string()
+    }
+
+    #[test]
+    fn rejects_type_parameters() {
+        let e = expand_err(parse_quote! {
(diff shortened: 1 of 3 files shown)


rama-macros/src/lib.rs
```diff
@@ -222,12 +222,21 @@ pub fn derive_extension(item: TokenStream) -> TokenStream {
...
+/// `Option<Arc<T>>` (owned Arc clone), and the two may be mixed. A field may
+/// also be `Option<(&'a T, usize)>` / `Option<(Arc<T>, usize)>` to additionally
+/// capture the entry's traversal rank (`0` is the newest value seen, growing for
+/// older ones, so ranks order fields by recency). A borrowed field requires the
+/// struct to carry the matching lifetime (`struct View<'a>`); an all-`Arc`
+/// struct needs no lifetime. Generates `fn from_extensions(ext: &Extensions) ->
+/// Self`, where each field uses the same lookup as `Extensions::get_ref` but the
+/// store is traversed only once.
+///
+/// On an*enum, each variant is a one-field tuple variant naming a candidate
```
GlenDC:
```suggestion
/// On an enum, each variant is a one-field tuple variant naming a candidate
```

rama-macros/src/from_extensions_macro.rs
```diff
@@ -1,9 +1,27 @@
...
 
+/// How a struct field is filled.
+enum FieldShape<'a> {
+    /// `Option<&'a T>`, `Option<Arc<T>>`, or `(_, usize)` variant
+    Piece(FieldSpec<'a>),
+    /// `Option<G>` where `G` is itself a `#[derive(FromExtensions)]` group
+    Nested(&'a Type),
+}
+
+/// An extension piece and if its indexed
```
GlenDC:
```suggestion
/// An extension piece and whether it's indexed
```

rama-macros/src/from_extensions_macro.rs
```diff
@@ -182,3 +455,18 @@ fn option_inner(ty: &Type) -> Option<&Type> {
     };
     Some(inner)
 }
+
+/// Match a two size tuple
```
GlenDC:
```suggestion
/// Match a 2-element tuple
```

rama-macros/src/from_extensions_macro.rs
```diff
@@ -38,96 +56,350 @@ pub(crate) fn expand(input: DeriveInput) -> syn::Result<TokenStream> {
...
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+    let (fn_generics, lt) = fn_lifetime(lifetime);
 
-    // The lifetime tying the borrowed slots to `ext`: reuse the struct's
-    // lifetime if it has one, otherwise introduce a fresh one on the method
-    // (a struct with only `Arc<T>` fields needs no lifetime of its own).
-    let (fn_generics, lifetime) = if let Some(lt) = lifetimes.first() {
-        (quote!(), quote!(#lt))
+    // Trait impl header: the group lifetime is the enum's own lifetime, or a
+    // new one introduced on the impl if only arc's are used
```
GlenDC:
```suggestion
    // new one introduced on the impl if only `Arc`s are used
```

GlenDC (changes requested):
Nice new feature. As much as possible we ensure there's 1 extension type that covers something... But there will always be situations where you can derive it from multiple types and there this features comes in greatly, so well done.

One big thing I notice is that we do not have any tests in rama-macros itself. There are some in rama-core, that do cover the happy and some other paths, including for this new logic. But would be nice to have some macro-specific tests. Wasn't added for this functionality in past either.

Other than this and my comments I have also a small nit. I think it would be good if we document somewhere somehow that:

a. These ranks are to be seen as an opaque ordering, not a continigious index
b. min_by_key returns the first on ties, so if two enum variants name the same type the earlier-declared one wins. It's deterministic, but not documented

GlenDC (approved):
Thank you looks solid.

PR #1004, opened 2026-06-20
chore(datastar): update embedded datastar bundle to v1.0.2

seuros:
Re-vendor datastar.js and datastar.js.map from starfederation/datastar@1.0.2 and bump the source URLs in DatastarScript/DatastarSourceMap docs.


GlenDC (approved):
Thank you! Looks great!

PR #1007, opened 2026-06-20
refactor similar pattern and document/extract

seuros:
A few of these are judgment calls, so please tell me if you dislike any of the patterns and I'll drop or rework that commit. Each one is isolated in its own commit so they are easy to take or leave individually.

where the same body was copied 10–14 times, a bug or behavior change now lives in one place instead of every copy, and the intent is clearer. 

I kept the per-type docs and doctests intact and left idiomatic one-liner boilerplate alone, since collapsing that would hurt readability more than it helps.


rama-net/src/stream/mod.rs
```diff
@@ -11,6 +11,66 @@ mod socket;
...
 
+/// Implements [`tokio::io::AsyncRead`] and [`tokio::io::AsyncWrite`] for a
+/// pin-projected newtype by delegating every method to the named inner field.
+///
+/// The wrapped type must obtain its projection through `pin_project!` with
+/// `$field` marked `#[pin]`. Used by the `TcpStream`/`UnixStream` wrappers,
+/// which only differ in the inner stream they hold.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! rama_delegate_async_read_write {
```
GlenDC: pattern is to name this `__rama_delegate_async_read_write` and than re-export from within here using `rama_delegate_async_read_write` so that we do not polute our root namespace

GlenDC (approved):
Looks good thank you. Reverted 1 commit and cleaned up one other thing.

PR #1015, opened 2026-06-22
Migrate server TLS to TlsServerConfig similar to clients

soundofspace:
- Migrate boring to TlsServerConfig
- Migrate rustls to TlsServerConfig
- Use pki types for Authentication
- Remove opaque DataEncoding

- Move cert generation logic to rama-crypto using rcgen/aws-lc/boring depending on features
- Get rustls/rcgen and boring cert gent mostly to feature parity (rcgen doesnt support all rsa types)
- Fix subject_alternative_names wrongfully applied in boring cert


rama-net/Cargo.toml
```diff
@@ -69,6 +69,7 @@ pin-project-lite = { workspace = true }
 psl = { workspace = true }
 radix_trie = { workspace = true }
 rama-core = { workspace = true }
+rama-crypto = { workspace = true }
```
GlenDC: optional = true, so we opt-in to it for tls feature, no?

GlenDC (changes requested):
Lovely, a whole lot simpler. Very happy this was still done as part of 0.3.

Besides my comment, I think there's also one less visible side effect. And maybe I'm wrong but as far as I understand it from just reading this PR:

* self-signed: now generates a new cert for every new connection? Besides that being computational expensive it also means each server connection has a different fingerprint?
* in same vein we are building an entire config for each conneciton, before we shared it between connections

Nor sure if/how/what we need to improve here, but this feels like a move in the wrong direction ... Yet I like the simplicity of it all, so is there a way to marry both worlds?

soundofspace:
Hmm one thing im not sure cert logic has now moved to rama-crypto so i make Domain -> String, since that is in rama-net, and rama-net depends on rama-cyrpto. Any opinions here or how how we want to tackle import order? And how we see this going forward with rama-crypto growing more

GlenDC:
> Hmm one thing im not sure cert logic has now moved to rama-crypto so i make Domain -> String, since that is in rama-net, and rama-net depends on rama-cyrpto. Any opinions here or how how we want to tackle import order? And how we see this going forward with rama-crypto growing more

rama-net does _not_ depend on rama-crypto?

soundofspace:
> > Hmm one thing im not sure cert logic has now moved to rama-crypto so i make Domain -> String, since that is in rama-net, and rama-net depends on rama-cyrpto. Any opinions here or how how we want to tackle import order? And how we see this going forward with rama-crypto growing more
> 
> rama-net does _not_ depend on rama-crypto?

I does now `tls = ["dep:hex", "dep:md5", "dep:sha2", "dep:jiff", "dep:rama-crypto"]` for pki types

GlenDC:
> > > Hmm one thing im not sure cert logic has now moved to rama-crypto so i make Domain -> String, since that is in rama-net, and rama-net depends on rama-cyrpto. Any opinions here or how how we want to tackle import order? And how we see this going forward with rama-crypto growing more
> > 
> > 
> > rama-net does _not_ depend on rama-crypto?
> 
> I does now `tls = ["dep:hex", "dep:md5", "dep:sha2", "dep:jiff", "dep:rama-crypto"]` for pki types

yeah but that's a mistake than. rama-net only should contain rama network stuff. It can contain tls / http specific bits but only minimal and usually in form of types or extensions, only in function and there because they require  to be avaialble as info or w/e in amny crates.

We should have no need to ever put rama-crypto in there. Networking by itself doesn't do crypto. That is more for other crates (e.g. tls crates, authentication crates etc...)

examples/acme_tls_challenge_using_boring.rs
```diff
@@ -91,10 +81,17 @@ use rama::{
...
+        },
+    },
+    tls::{
+        client::{ClientHello, ServerVerifyMode, TlsClientConfig},
+        server::{DynamicCertIssuer, ServerAuthData, TlsServerConfig},
     },
+    utils::collections::smallvec::smallvec,
 };
-use rama_net::tls::KeyLogIntent;
+use rama_tls::KeyLogIntent;
```
GlenDC: probably another candidate for your lovely extra checks, to never have `rama_*` imports in examples (so that examples can be copy pasted by users only having `rama + tokio` as dependencies (perhapos sometimes also other ones, but in general that's the idea

GlenDC (changes requested):
PR looks good overall and passes `cargo check` incl. all-features. Two small issues: `TlsAcceptorData` per-connection override seems silently removed/broken while README still documents it, so restore or update docs/API. Also `rama-tls-boring` README still mentions removed `DataEncoding`.

GlenDC (approved):
Great. Looks good now. And sorry if there were any merge conflicts since yesterday 😓

Issue #1019, opened 2026-06-23
Switch IO traits to owned variants


soundofspace:
### What problem does this solve?

Native Tokio AsyncRead and AsyncWrite are incompatible with io-uring. By switching to owned variants we will open support to io-uring and potentially other runtimes such as compio (this is however not the goal, just a potential nice side-effect).

### Proposed solution

How these traits will look is still a WIP, but they will have to support taking ownership of a buffer. To also be able to play nicely with Hyper and other poll based async code they will also have to be designed around a poll based API, so traits native async traits like used in compio/glommio/tokio-io-uring are off the table.

Since we will be owning these traits we will also have to integrate all utilities tokio offers for these.

PR #1037, opened 2026-06-26
Initial owned IO buffer traits

soundofspace:
Owned IO traits which will replace tokio io traits so we can support completion based io

### DIFF
--- a/rama-core/src/io/mod.rs
+++ b/rama-core/src/io/mod.rs
@@ -18,6 +18,13 @@ pub mod timeout;
 mod bridge;
 pub use bridge::BridgeIo;
 
+mod owned;
+#[doc(inline)]
+pub use owned::{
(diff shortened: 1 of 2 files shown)


soundofspace:
@GlenDC when you have time (no rush), can you give this a review. Will fix ci later, but these are the traits for owned IO that will unlock io-uring while still supporting poll based futures.

GlenDC:
> @GlenDC when you have time (no rush), can you give this a review. Will fix ci later, but these are the traits for owned IO that will unlock io-uring while still supporting poll based futures.

looks fine to me, don't see immediate concerns. your explanation + code so far makes sense to me. I can however imagine that as you start implementing and using it that some more refining and corrections might be neeeded, but for now looks ok to me.

Will be easier to tell once some more interesting setups demonstrate it in more depth.

GlenDC (approved):

soundofspace:
part 1 of #1019

PR #1041, opened 2026-06-27
rmove DNS out of rama-tcp

### LINKED ISSUE #841: rmove DNS out of rama-tcp
Instead of handling DNS resolution implicitly, we should provide a dedicated connector service backed by a DNS provider.

Together with the work in #840, this would make connectors such as `TcpConnector` much simpler and more composable.

The idea is that connectors are stacked, each with a single responsibility:

```
... other connectors
|
ProxyTargetConnector
  Uses the proxy target if present as the connection target
|
DnsAddressResolverConnector
  Resolves the target if it is a domain name
|
TcpConnector
  Only accepts a socket address (or something convertible to it).
  If the target is still a domain name, it should fail.
```

This keeps each connector simple and explicit.

### Benefits

* Clear separation of responsibilities between connectors
* Simpler connectors such as `TcpConnector`
* More flexible connector stacks
* Enables deeper control over the DNS stack in the future (for example allowing `rama-dns` to rely on `rama-tcp`)

### Implications

For higher level utilities such as `EasyHttpWebClient`, the stack builder can inject the required connectors automatically depending on the needs.

This approach also avoids leaking connection related data through extensions.

### Extensions

We should stop inserting `ProxyAddress` into extensions.

Instead we should inject:

* `ProxyTarget`
* `ProxyCredentials`

`ProxyAddress` can still exist as a type, but it would no longer have a Rama defined meaning inside extensions and would not be injected automatically. Users can still insert it themselves if they want.

GlenDC:
* move rama-dns out of rama-tcp and rama-udp
* simplify rama-tcp (eg drop factory, was bit silly isolated concept in the end)
* make dns optional in rama-socks5 and if used just part of connector
* integrate and cleanup all that follows

Closes #841


rama-dns/src/client/connector.rs
```diff
@@ -0,0 +1,523 @@
...
+        Result<EstablishedClientConnection<C, Request>, BoxError>,
+    ),
+>;
+
+#[derive(Debug, Clone)]
+/// A [`Layer`] which wraps a transport connector with DNS resolution.
+///
+/// The produced [`DnsConnector`] accepts higher-level inputs, resolves their
+/// [`ConnectorTargetInputExt::connector_target`] domain target to IP targets,
+/// and calls the inner transport connector with a fresh L4 [`Request`] per IP.
```
soundofspace: Why make a new request for this?

rama-dns/src/client/connector.rs
```diff
@@ -0,0 +1,523 @@
...
+            let transport_input = make_transport_input(&input, target.port, ip, None);
+            let EstablishedClientConnection { conn, .. } = self
+                .inner
+                .connect(transport_input)
+                .await
+                .map_err(Into::<BoxError>::into)?;
+            input.extensions().insert(ConnectorTarget(HostWithPort::new(
+                Host::Address(ip),
+                target.port,
+            )));
```
soundofspace: It feels wrong to replace input/request and then also ignore output with potentially changed input/request

soundofspace: Eg having to insert ConnectorTarget is already an example of something going wrong here make_transport_input forks extensions so all extensions set in lower layers eg tcp layer or in make_transport_input are effectively lost

soundofspace (changes requested):
@GlenDC only just looked at connector.rs but I think replacing input (+forking extensions) is the wrong approach here? Why are you doing it like this as far as i can see you still rely on ConnectorTargetInputExt in tcp connector so why not just leave input and just make sure that is properly set? Or in case we do want to only support l4 request, really make this is forced input type?

rama-dns/src/client/connector.rs
```diff
@@ -0,0 +1,299 @@
...
+pub struct DnsAddressCandidates<R> {
+    resolver: R,
+    domain: Domain,
+    port: u16,
+}
+
+impl<R> DnsAddressCandidates<R> {
+    /// Create a new [`DnsAddressCandidates`] resolving `domain:port` via `resolver`.
+    #[must_use]
+    pub const fn new(resolver: R, domain: Domain, port: u16) -> Self {
```
GlenDC: Is there a reason this is public? Seems like a private detail to me?

rama-socks5/src/server/udp/relay.rs
```diff
@@ -474,45 +483,63 @@ impl UdpSocketRelay {
...
+        let domain = host
+            .try_into_domain()
+            .context("host is not resolvable as a domain via SOCKS5 udp relay")?;
+        let dns_resolver = self
+            .dns_resolver
+            .as_ref()
+            .context("domain cannot be resolved: no dns resolver defined")?;
+
+        match self.dns_resolve_mode {
+            DnsResolveIpMode::SingleIpV4 => Ok(IpAddr::V4(
```
GlenDC, line 511: "single" here does not mean only 1 ip, it means 1 ip family, so just lookup ipv4, but can be multiple address ,np. Same goes for SIngleIpv6.

Incase I had that as a bug before, sorry

rama-tcp/src/client/service/connector.rs
```diff
@@ -2,28 +2,25 @@ use rama_core::{
...
     stream::{Socket, SocketInfo},
     transport::TransportProtocol,
 };
 
 use crate::TcpStream;
 use crate::client::connect::TcpStreamConnector;
 
-use super::{CreatedTcpStreamConnector, TcpStreamConnectorCloneFactory, TcpStreamConnectorFactory};
+/// Max number of resolved-candidate connection attempts raced concurrently.
+const MAX_IN_FLIGHT_CONNECT_ATTEMPTS: usize = 3;
```
GlenDC: I hadn't exposed it before, but I think as a default great, but we probably should allow it to be overwritten, WDYT? in the tcp connector that is

GlenDC (commented):
looks amazing, just a couple of remarks @soundofspace , WDYT?

rama-tcp/src/client/service/connector.rs
soundofspace: agree

rama-socks5/src/server/udp/relay.rs
soundofspace: All this logic here in socks udp expects a single so how would this look, or do we want complete happy eyeballs and multiple attempts also?

soundofspace: Discussed offline this is expected behaviour here

GlenDC (commented):
Approved, cannot approve my own PR, but great teamwork @soundofspace

soundofspace (approved):

PR #1044, opened 2026-06-30
Multiplexing connection pool (default for EasyHttpWebClient)

### LINKED ISSUE #1040: Multiplexed connection pool
### What problem does this solve?

Connection pool currently works by checking out a single connection from the pool and giving it out to the users, effectively a connection is only used by one consumer at the same time. This works perfectly for things like TCP or http1, but to use h2 efficiently we want to be able to multiplex connections.

### Proposed solution

Add a different Pool using the already provided building blocks that instead of giving out the actual connection hands out a virtual one, which internally maps map to the original connection. By doing that we can multiplex and really on the fact that services/connection are `&self `.

This will effectively replace the current connection pool for HTTP use cases. The other one will still continue to exist for other other cases where single ownership is needed.

This will also automatically detect h2 settings such as steam concurrency and automatically configure the pool for that (or max concurrency 1 for http1)

### Alternatives considered

_No response_

soundofspace:
Adds a pool where one backend connection can serve multiple concurrent requests, and makes it the default HTTP client pool.

Pool:
- New MultiplexPool. It keeps connections in storage and hands out cheap virtual handles instead of checking a connection out exclusively. The old exclusive pool is just this with capacity 1.
- Connections set a MaxConcurrency extension: h1 sets 1, h2 sets whatever it negotiates. The pool multiplexes up to min(MaxConcurrency, configured max_concurrent_streams).
- A MaxConcurrency bump (h2 raising its SETTINGS) wakes parked pool waiters, with no per-connection tasks and no leaks.
- MuxSelection picks which connection serves a request when several can.
- Moved the existing pool to a new exclusive module. Import paths stay the same.

Connection follows the body:
- A connection now returns to the pool when the response body finishes (or is dropped), not when the headers arrive. Before this it could be reused too early and threw off concurrency and idle tracking. The exclusive pool had the same latent bug.
- GuardedBody plus a BindBodyToConn layer move the connection into the response body so the release fires on body end or drop.

Reactive:
- New Reactive<T>: lock free atomic reads plus change notification. Backs both MaxConcurrency and ConnectionHealthWatcher. 

HTTP pool logic (HttpPooledConnectorConfig, connection identity) now lives in rama-http-backend, so all HTTP pooling behavior is tested there: reuse after the body is consumed, streaming bodies, h1 reuse safety, multiplex vs no multiplex, and stream limits.


examples/http_pooled_client.rs
```diff
@@ -67,7 +67,11 @@ async fn main() {
         .with_proxy_support()
         .without_tls_support()
         .with_default_http_connector(Executor::default())
-        .try_with_connection_pool(HttpPooledConnectorConfig::default())
+        .try_with_connection_pool(HttpPooledConnectorConfig {
+            max_concurrent_streams: 20,
```
GlenDC, line 74: why these overwrites are needed? or just to show off? If latter might be useful for some explanation as a doc comment here

GlenDC (changes requested):
Looks solid, though dense to review. Had to read couple of times, but guess that is just the subject that isn't all that easy.

I do think I found 2 possible mistakes, can you please look?

1. h2 SETTINGS capacity increases don’t wake pool waiters, if my assumption is correct that might mean requests can stall until timeout/stream drop;
2. as far as I can see h2 stream capacity is released when response headers arrive, not when the body stream finishes, skewing limits/idle tracking.

If your own findings align with mine I would suggest adding some tests for it first and than fix it.

GlenDC:
Regression tests worth adding: full `MultiplexPool`, two same-id waiters; freeing a slot lets one create a cap=2 conn and the other should wake/reuse it. Also assert h2 `MaxConcurrency(0)` means no new stream admission, not clamp-to-1.

E.g.  example test that fails:

```rust
    struct SlowConnector {
        created: AtomicUsize,
        delay: Duration,
    }

    impl<Input> Service<Input> for SlowConnector
    where
        Input: Send + 'static,
    {
        type Output = EstablishedClientConnection<Conn, Input>;
        type Error = Infallible;

        async fn serve(&self, input: Input) -> Result<Self::Output, Self::Error> {
            tokio::time::sleep(self.delay).await;
            let serial = self.created.fetch_add(1, Ordering::Relaxed);
            let conn = Conn {
                serial,
                extensions: Extensions::new(),
            };
            conn.extensions.insert(ConnectionHealthWatcher::default());
            Ok(EstablishedClientConnection { input, conn })
        }
    }

    #[tokio::test]
    async fn new_multiplexed_connection_wakes_waiters() {
        let pool = MultiplexPool::try_new(2, 1).unwrap();
        let svc = PooledConnector::new(
            SlowConnector {
                created: AtomicUsize::new(0),
                delay: Duration::from_millis(100),
            },
            pool,
            id_fn as fn(&ServiceInput<u32>) -> Result<TestId, BoxError>,
        )
        .with_wait_for_pool_timeout(Duration::from_millis(500));

        let c1 = svc.connect(ServiceInput::new(1u32)).await.unwrap();

        let waiter1 = svc.connect(ServiceInput::new(2u32));
        let waiter2 = svc.connect(ServiceInput::new(2u32));
        tokio::pin!(waiter1);
        tokio::pin!(waiter2);

        tokio::time::sleep(Duration::from_millis(20)).await;
        drop(c1);

        let (r1, r2) = tokio::join!(waiter1, waiter2);
        assert!(r1.is_ok(), "first waiter should create a new connection");
        assert!(r2.is_ok(), "second waiter should reuse the spare stream slot");
    }
```

GlenDC (approved):

PR #1047, opened 2026-07-01
rama-net, rama-core and rama are now also nostd compatible (with std gates)

GlenDC:
* Make rama core and net no-std friendly
* Make rama facade no-std friendly
* Gate Unix examples on unix feature


GlenDC:
with rama-net added to the mix and possible to use directly from `rama`, this is now good enough nostd-compatible wise for 0.3. It allows windows-driver (so called examples of proxy companion projects) also to use rama for the surface where it matters to such projects.

PR #1054, opened 2026-07-02
Fix H2 early connection WINDOW_UPDATE replay

### LINKED ISSUE #1052: H2 replayed connection WINDOW_UPDATE emits wrong size_increment
### What happened?

When an emulation/embedded profile carries an early connection-level `WINDOW_UPDATE` frame, rama replays it with the wrong `size_increment` on the wire. The captured value is treated as an *absolute target window size* rather than the frame's actual increment, so the emitted increment is off — and, because it interacts with the default 5 MiB connection window, the error is not even a constant offset.

This makes the akamai_h2 fingerprint (and its hash) not match the browser being emulated.

### How can we reproduce it?

Profile with an early frame:
```json
{ "WindowUpdate": { "size_increment": 15663105, "stream_id": 0 } }
```
Route a request through the proxy to echo.ramaproxy.org.

Expected (real Chrome, no proxy)

WINDOW_UPDATE size_increment = 15663105
akamai_h2 = 1:65536;2:0;4:6291456;6:262144|15663105|0|m,a,s,p
hash = 52d84b11737d980aef856699f885ca86

Actual (through the proxy)

WINDOW_UPDATE size_increment = 15597570        # == 15663105 - 65535
akamai_h2 = 1:65536;2:0;4:6291456;6:262144|15597570|0|m,a,s,p
hash = 77632e31924ef556b6b218ddc6179f66

Setting the profile value to 15728640 (15663105 + 65535) produces the correct on-wire value, confirming the start window (65535) is being subtracted.

### rama version

5d7c98db73571f6d6b26c032f6982edb2a838fe0

### Platform(s)

macOS

### Relevant logs or output

```shell

```

ntrippar:
Fix #1052 Corrects early connection-level WINDOW_UPDATE handling by treating the increment as an absolute replay target, adds regression coverage for WINDOW_UPDATE(0, 15663105), validates against the H2 max window size, and documents that the helper is intended for connection-level frames only.


tests/http-core/h2/flow_control.rs
```diff
@@ -2836,3 +2836,65 @@ async fn poll_capacity_window_update_settings_race() {
...
+
+    let srv = async move {
+        let settings = srv.assert_client_handshake().await;
+        assert_default_settings!(settings);
+        srv.recv_frame(frames::window_update(0, INCREMENT)).await;
+    };
+
+    let client = async move {
+        let (_client, conn) = client::Builder::new()
+            .with_initial_connection_window_size(5 * 1024 * 1024)
```
GlenDC:
```suggestion
            .with_initial_connection_window_size(mib(5))
```

GlenDC (approved):
Looks good to me, just a small nit.

GlenDC:
Not a new issue, but I think next to documenting that we only care about the first windows update for these replays, given that is afaik what fingerprinting also worries about and more windows updates get anyhow more complicated quickly.... would be to perhaps also ensure in code that we only ever apply the first windows update.

In `EarlyFrameStreamContext::new_replayer`, before `frames.reverse()`:

```
let mut seen_connection_window_update = false;
frames.retain(|frame| match frame {
    EarlyFrame::WindowUpdate(window_update) if window_update.stream_id.is_zero() => {
        let keep = !seen_connection_window_update;
        seen_connection_window_update = true;
        keep
    }
    _ => true,
});
```

soundofspace:
Looks good, thanks for the fix!

GlenDC (approved):

PR #1064, opened 2026-07-09
rama-ttrpc

soundofspace:
- Add rama-ttrpc and rama-ttrpc-build
- Followed same patterns and structure from rama-grpc where it makes sense
- Also added example server + test with client


soundofspace:
Added _example_ttrpc and _example_grpc features so we can put ttrpc/grpc example in the root examples folder. I think this is pretty nice and if you agree will also upgrade some grpc example to there vs the weird structure we have now

rama-ttrpc-build/src/service_generator.rs
```diff
@@ -0,0 +1,218 @@
...
+
+        let trait_methods = service.methods.iter().map(|m| trait_method(&service, m));
+        let client_methods = service.methods.iter().map(|m| client_method(&service, m));
+        let dispatch_branches = service.methods.iter().map(|m| dispatch_branch(&service, m));
+
+        let tokens = quote! {
+            #[allow(dead_code, non_snake_case)]
+            pub fn #service_ident<T: #service_ident>(
+                target: impl std::convert::Into<std::sync::Arc<T>>,
+            ) -> impl rama_ttrpc::__codegen_prelude::Service {
```
GlenDC: Seems you are hardcoding `rama_ttrpc` which will fail if users only have `rama` as dependency. You'll need to apply our usual trick (try `rama_ttrpc` first and if that doesn't work try `rama`)

rama-ttrpc/src/client/request_handlers.rs
```diff
@@ -0,0 +1,421 @@
...
+                payload: (),
+                metadata,
+                timeout_nano: timeout.as_nanos(),
+            },
+        };
+
+        let fut = self.spawn_stream(frame, move |res, mut stream| async move {
+            res.await.map_err(Status::send_error)?;
+
+            let rx = RwLock::new(&mut stream.rx);
```
GlenDC: not super important, but how frequent do you expect this to be written too? RwLock isn't all _that_ cheap... better then a regular lock but still a lock. But I wonder how frequently we plan to write to it to begin with

rama-ttrpc/src/client/request_handlers.rs
```diff
@@ -0,0 +1,421 @@
...
+
+fn handle_server_unary<'a, Output: prost::Message + Default + 'a>(
+    rx: &'a RwLock<&'a mut StreamReceiver>,
+    tx: oneshot::Sender<Output>,
+) -> impl Future<Output = Result<()>> + Send + 'a {
+    #[expect(
+        clippy::unwrap_used,
+        reason = "try_write runs synchronously before any await on a lock owned solely by this handler, so it cannot be contended"
+    )]
+    let mut rx = rx.try_write().unwrap();
```
GlenDC: Seems like a design smell, might need to do better so we do not even need such an implicit assumption

rama-ttrpc/src/server/controller.rs
```diff
@@ -0,0 +1,21 @@
+use tokio_util::sync::CancellationToken;
+
+#[derive(Clone, Default)]
+pub struct ServerController {
+    pub(super) shutdown: CancellationToken,
```
GlenDC: Why the split of these 2 ? Shutdown vs Abort?

rama-ttrpc/src/server/method_handlers.rs
```diff
@@ -0,0 +1,298 @@
...
+        stream: &'a mut StreamIo,
+    ) -> Pin<Box<dyn Future<Output = Result<()>> + Send + 'a>> {
+        Box::pin(async move {
+            let rx = RwLock::new(&mut stream.rx);
+
+            if flags.bits() != (Flags::REMOTE_OPEN | Flags::NO_DATA).bits() {
+                // REMOTE_OPEN must be set (as the client is a stream)
+                // NO_DATA must be set, as the request doesn't include a stream payload
+                return Err(Status::invalid_request_flags(
+                    Flags::REMOTE_OPEN | Flags::NO_DATA,
```
GlenDC: Should `NO_DATA` to also belong to Request frames? How is this handled in containerd?

rama-ttrpc/src/server/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+pub(crate) mod method_handlers;
+
+pub use controller::ServerController;
+
+/// A ttRPC server as a rama [`Service`](rama_core::Service): it serves one
+/// [`ServerConnection`] per accepted stream, so it can be handed straight to a rama
+/// listener (`rama-tcp`, `rama-unix`, ...) instead of writing the per-connection loop by hand.
+///
+/// ```rust,ignore
+/// let server = rama_ttrpc::TtrpcServer::new().register(my_service(handler));
```
GlenDC: in general not too big of a fan of these example blocks, PITA to maintain

rama-ttrpc/src/types/protos/code.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// `OUT_OF_RANGE` if asked to read from an offset past the current
+    /// file size.
+    ///
+    /// There is a fair bit of overlap between `FAILED_PRECONDITION` and
+    /// `OUT_OF_RANGE`.  We recommend using `OUT_OF_RANGE` (the more specific
+    /// error) when it applies so that callers who are iterating through
+    /// a space can easily look for an `OUT_OF_RANGE` error to detect when
+    /// they are done.
+    ///
+    /// HTTP Mapping: 400 Bad Request
```
GlenDC: why these mentions?

rama-ttrpc/src/types/protos/raw_bytes.rs
```diff
@@ -0,0 +1,96 @@
...
+pub struct RawBytes(Bytes);
+
+impl RawBytes {
+    pub fn decode<Msg: prost::Message + Default>(&self) -> Result<Msg, DecodeError> {
+        Ok(Msg::decode(self.0.clone())?)
+    }
+}
+
+impl Buf for RawBytes {
+    fn remaining(&self) -> usize {
```
GlenDC, line 18: you can force inlining for many of the methods in this module, no?

rama-ttrpc/src/types/protos/raw_bytes.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.0.chunk()
+    }
+
+    fn advance(&mut self, cnt: usize) {
+        self.0.advance(cnt);
+    }
+}
+
+pub trait ProstField: Send + Sync {
+    fn encode(&self, tag: u32, buf: &mut impl BufMut);
```
GlenDC, line 44: given this is a publc trait, might want toa dd some documentation to what the expected contract of the trait and its methods is

rama-ttrpc/src/types/protos/request.rs
```diff
@@ -0,0 +1,132 @@
+use super::raw_bytes::{ProstField, RawBytes};
+use crate::types::message::{Message, MessageType};
+use crate::types::protos::KeyValue;
+
+#[derive(Clone, PartialEq, Debug, Default)]
+pub(crate) struct Request<Payload: ProstField + Default = RawBytes> {
+    pub service: String,
+    pub method: String,
```
GlenDC: perhaps out of scope here, but seems that it's a bit wasteful to have to declare things like service and methods as strings. Normally would shrug it away but as you say it's for low-memory environments. This isn't an open set within the context of the user, no? E.g. I imagine there is only a fixed amount of methods and services possible for a given project? One could expec they need to define an enum for these or something like it, or auto-generated from proto... That is than used in the request... I guess would make it all more complex, but at hte same time... yeah dunno feels wasteful. As it is essentially part of each request, twice.

rama-ttrpc/src/types/protos/request.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub timeout_nano: i64,
+    pub metadata: Vec<KeyValue>,
+}
+
+impl<Payload: ProstField + Default> Message for Request<Payload> {
+    const TYPE_ID: MessageType = MessageType::Request;
+}
+
+/*
+This is a modified version of the code generated by cargo expand on:
```
GlenDC: what is missing is the reason why this manual vs the macro based approach

rama-ttrpc/src/types/frame.rs
```diff
@@ -0,0 +1,133 @@
...
+use prost::bytes::{Buf, BufMut, Bytes, BytesMut};
+use tokio::io::{AsyncRead, AsyncReadExt as _};
+
+use crate::types::encoding::{BufExt as _, DecodeError, Decodeable, Encodeable, InvalidInput};
+use crate::types::flags::Flags;
+use crate::types::message::{FallibleBytesMessage, Message};
+use crate::types::protos::raw_bytes::ProstField;
+use crate::types::protos::{Response, Status};
+
+const MAX_DATA_LENGTH: usize = 4 << 20;
```
GlenDC: why expressed so weird instead of using the mib constructor, same for the discard page size?

rama-ttrpc/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+[dependencies]
+ahash = { workspace = true }
+bitflags = { workspace = true }
+prost = { workspace = true }
+prost-types = { workspace = true }
+rama-core = { workspace = true, features = ["std"] }
+rama-net = { workspace = true, features = ["std"] }
+rama-ttrpc-build = { workspace = true, optional = true }
+tokio = { workspace = true, features = ["io-util", "rt", "sync", "time", "macros"] }
+tokio-stream = { workspace = true }
```
GlenDC: tokio-stream is re-exported in rama-core, same for tracing. So shouldn't be direct dependnecies

src/lib.rs
```diff
@@ -126,6 +126,15 @@ pub mod gateway {
     pub use ::rama_fastcgi as fastcgi;
 }
 
+/// ttRPC ("gRPC for low-memory environments") support.
+///
+/// Unlike `grpc`, ttRPC does not ride on HTTP/2, it is a length-prefixed framing
```
GlenDC, line 131: in general I still feel we portray too much grpc as if it is bound to h2... yes it is the common, just like h2 and h3 are usually tls, but none of this is enforced... so dunno I would at least rephrase these kind of sentencs and docs accross the PR

examples/ttrpc_server.rs
```diff
@@ -0,0 +1,108 @@
...
+        clippy::all,
+        clippy::pedantic,
+        clippy::allow_attributes,
+        reason = "generated ttRPC/prost stubs are exempt from rama's lints"
+    )]
+    // Generated by the workspace `build.rs` into `OUT_DIR/ttrpc/` (namespaced by RPC
+    // flavour so it never collides with gRPC codegen of the same package name). This
+    // prefixing is something we do manually in `build.rs`, so normally this would be
+    // imported with `rama::ttrpc::include_proto!("rama.examples.greeter.v1")`
+    rama::ttrpc::include_proto!("ttrpc/rama.examples.greeter.v1");
```
GlenDC, line 69: this example works atm because you are part of the rama examples and thus also have access to rama's dependnecies. Which includes `rama_ttrpc`. Should you have it as a stand-alone crate similar to the grpc examples you would have caught the error that i commented on earlier.

rama-ttrpc/src/io/mod.rs
```diff
@@ -0,0 +1,292 @@
...
+            if let Err(MpscSendError(frame)) = stream_tx.send(frame) {
+                // the stream was already closed, return the message and let consumers handle it
+                return Some((id, frame));
+            }
+        }
+        None
+    }
+
+    fn stream(&mut self, id: u32) -> Option<StreamReceiver> {
+        let (tx, rx) = unbounded_channel();
```
GlenDC: perhaps should be opt-in, but these channels are now always unboudned. However, a peer can send valid frames of up to 4 MiB faster than the application consumes them, allowing one connection to exhaust process memory.

Not sure if should be the default behaviour or not, but I do think you'll want for production systems to allow some kind of mechanism to apply backpressure... or drastic measures such as closing streams whose buffers remain full (hard stop :P)

rama-ttrpc/src/types/frame.rs
```diff
@@ -0,0 +1,133 @@
...
+
+    buf.resize(HEADER_LENGTH + data_length, 0);
+    readable.read_exact(&mut buf[HEADER_LENGTH..]).await?;
+
+    Ok(buf.into())
+}
+
+async fn discard_bytes(reader: &mut (impl AsyncRead + Unpin), mut n_bytes: usize) -> IoResult<()> {
+    let mut buf = [0u8; DISCARD_PAGE_SIZE];
+    while n_bytes > 0 {
```
GlenDC: When the peer sends an oversized length and then closes, read returns Ok(0), leaving n_bytes unchanged...

rama-ttrpc/src/client/request_handlers.rs
```diff
@@ -0,0 +1,421 @@
...
+    Err(Status::timeout())
+}
+
+fn handle_input_stream<T: Send>(
+    input: impl Stream<Item = T> + Send,
+) -> (
+    UnboundedReceiverStream<T>,
+    impl Future<Output = Result<()>> + Send,
+) {
+    let (tx, rx) = unbounded_channel();
```
GlenDC: an infinite always-ready stream can monopolize the task and prevent network and timeout branches from running

rama-ttrpc/src/server/controller.rs
```diff
@@ -0,0 +1,21 @@
...
+    pub(super) abort: CancellationToken,
+}
+
+impl ServerController {
+    pub fn terminate(&self) {
+        self.abort.cancel();
```
GlenDC: terminate cancels the abort token, but nothing observes that token

rama-ttrpc-build/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        /// Add an argument to the `protoc` invocation.
+        ///
+        /// Passed directly to `prost_build::Config::protoc_arg`.
+        pub fn protoc_arg(mut self, arg: impl AsRef<str>) -> Self {
+            self.protoc_args.push(arg.as_ref().into());
+            self
+        }
+    }
+
+    rama_utils::macros::generate_set_and_with! {
```
GlenDC, line 144: prost-build requires file_descriptor_set_path whenever skip_protoc_run is enabled, but this builder provides no way to configure that path

GlenDC (changes requested):
Plenty of feedback, but that said. I love the prospects of this. Seems like an amazing new addition to the rama family of protocols! Great job.

Some of these comments make me think you might need to start playing with `cargo mutants` and also add some fuzzing tests... That would catch some of my feedback already + might even uncover stuff that I didn't notice myself ... As it's a big enough PR that there is probably still some stuff that might need improvement yet that I do not see. Some might be patches soon enough once you start using it in production for a while, but some might be lurking for a pretty long time...

rama-ttrpc/src/types/protos/code.rs
soundofspace: this is vendored from google source file, so comments are kept as is

GlenDC: Why? Not like they buy you anything? Not sure why we need to mix HTTP mapping? The mapping as done here is not all that logical to begin with, and ttrpc users might not even know http ? Don't see what preserving comments like this buys you?

Fine for me either way though, just do not see the value in them.

GlenDC (approved):

PR #1072, opened 2026-07-13
Add TLS certificate pinning

GlenDC:
This PR adds backend-agnostic TLS certificate pinning and custom trust anchors for rustls and BoringSSL.
Pins can be scoped per server, composed for key rotation and used together with normal certificate verification.


rama-tls/src/client/config.rs
```diff
@@ -155,6 +186,267 @@ pub struct TlsServerName(pub Host);
...
+#[derive(Debug, Clone, Extension)]
+#[extension(tags(tls))]
+pub struct TlsServerTrustAnchors(Arc<[CertificateDer<'static>]>);
+
+impl TlsServerTrustAnchors {
+    /// Create a non-empty set of server trust anchors.
+    pub fn try_new(
+        certificates: impl IntoIterator<Item = CertificateDer<'static>>,
+    ) -> Result<Self, BoxError> {
+        let certificates: Vec<_> = certificates.into_iter().collect();
```
soundofspace:
```suggestion
        let certificates: Arc<[CertificateDer<'static>]> = certificates.into_iter().collect();
```
does this not work?

tests/integration/examples/example_tests/tls_boring_cert_pinning.rs
```diff
@@ -0,0 +1,47 @@
...
+        [("RAMA_TLS_BORING_DYNAMIC_CERTS_ADDR", "127.0.0.1:64901")],
+    );
+    runner.get("https://127.0.0.1:64901").send().await.unwrap();
+
+    let secure_output = utils::ExampleRunner::run_with_args_output(
+        "tls_boring_cert_pinning",
+        ["https://127.0.0.1:64901", "examples/assets/example.com.crt"],
+    )
+    .await;
+    assert!(!secure_output.status.success(), "{secure_output:?}");
```
soundofspace, line 20: I wonder if we should using this pattern way more, this is kind of what i meant in the past by making examples configurable with clap/envs. Eg perfect uses cases i see is grpc client server pair, where both are examples and test is just basically connecting them? But not sure might be too much noise to make that possible

soundofspace (approved):
Nice, especially since it works for both boring and rustls alike

rama-tls/src/client/config.rs
GlenDC: Yes that does work, had it first as a Vec, seems I forgot to remove this. I would think you can implement that conversion zero-copy anyway between Vec and Arc slice, give nyou just drop the capcity , but not sure if it is implemented like that. And anyway why botter, your suggestion is anyway the correct one.

tests/integration/examples/example_tests/tls_boring_cert_pinning.rs
GlenDC: I am not a fan of general rules. As such I never was against using clap or similar stuff where it helps. However neither it is something I wanna promote in general.

If an example can be simple and self-contained you often do not need it. This one is not that useful as an example if we try to make it self-contained as you would need to run a server or w/e which complicates the example much more than a bit of clap config.

GlenDC:
> Nice, especially since it works for both boring and rustls alike

Thx, did a couple of minor improvements to the examples / docs, so nothing that should affect your assesment. Your feedback is applied as well. Will merge once CI = green.

PR #1074, opened 2026-07-16
Make examples a crate so these are fully external

soundofspace:


Cargo.toml
```diff
@@ -3,6 +3,10 @@ name = "rama"
 readme = "README.md"
 documentation = "https://docs.rs/rama"
 description = "modular service framework"
+# Examples live in the standalone `rama-examples` crate at ./examples. Cargo still
```
GlenDC: a one liner at max would suffice here, we can lookup full explanation in cargo.toml docs, as that goes for any property either way. Neither somethng we would by accident easily flip anyhow

Cargo.toml
```diff
@@ -27,6 +31,7 @@ ignored = ["rustversion"]
 [workspace]
 members = [
     ".",
+    "examples",
     "examples/grpc",
```
GlenDC, line 34: why is this a separate one but nothing else?

GlenDC (changes requested):
Seems to go in the right direction. I think this is the right choice. However... I wonder. Why not go the full mile than and group them all? E.g. all tcp under tcp, all http under http, this way you can also remove that prefix from the exmaple name. 

You could even that way only add the features for those project in rama that you need for that kind of group of examples. Not sure that buys you much though.

Or you can also just have those sub-modules anyway but keep it all in examples project, but still have some more order in those examples instead of all of them under `/src`.

GlenDC (approved):

PR #1078, opened 2026-07-24
feat: add http-backend feature for slimmer http builds

seuros:
The `EasyHttpWebClient` stack was only reachable via `http-full`, which drags in ws, idna (icu), compression, embedded ua profiles, multipart, html and rss — none of which are required by the http backend itself.

The new `http-backend` feature enables just `dep:rama-http-backend`, `dep:rama-http-core` and `dep:tokio` on top of `http`. `http::client` is gated on `http-backend` + `dns` + `tcp`, and `http::tls` (`CertIssuerHttpClient`) now shares those gates, fixing a pre-existing `http,tls` build break. `http-full` stays a strict superset (its redundant `dep:ahash`, already enabled via `std`, is dropped), so existing users are unaffected.

A client-only build (`http-backend,dns,tcp`) drops ~50 crates from the dependency tree (229 → 176 without default features). Optional features such as compression, rustls, ring, socks5, opentelemetry and unix compose with it as before.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -463,21 +463,26 @@ http = [
     "rama-fastcgi?/http",
     "rama-tower?/http",
 ]
-http-full = [
+# minimal http client stack: connector builder + `EasyHttpWebClient`,
+# without the extras (ws, idna, compression, embedded ua profiles, ...)
+http-client = [
     "http",
     "tcp",
     "dns",
-    "ws",
-    "idna",
     "dep:rama-http-backend",
     "dep:rama-http-core",
+    "dep:tokio",
+]
+http-full = [
+    "http-client",
+    "ws",
+    "idna",
     "dep:ahash",
     "ua-embed-profiles",
     "compression",
     "multipart",
     "html",
     "rss",
-    "dep:tokio",
     "rama-grpc?/transport",
 ]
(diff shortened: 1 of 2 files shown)


GlenDC (changes requested):
You are correct and you are going in the right direction. However I would not focus so much on client here. Your problem can be made more generic. 

What about if we introduce a feature named `http-backend` which enables `dep:rama-http-core` and `dep:rama-http-backend`. And drop the `http-client` feature introduced by you. `http-backend` would also enable `http` and `http-full` would enable `http-backend`. `http-backend` would also enable `dep:tokio`.

For the `EasyHttpClient` I would feature-gate than on `http-backend`, `dns` and `tcp`. 

I think that makes more sense than specifically focussing on client. As in general I want to keep the amount of features as little as possible. And by doing it like I propose we can also use the same gating for proxy and server stuff without having to also add features for those. Should we have the need for it in future.

seuros:
What i did here is introduce `http-client` which is a minimal  version of `http-full` .

This was a quick hack and i was not sure if that align with your vision of the project.

Your solution is better, and i will think more tomorrow about the edge cases before updating this PR.

GlenDC (approved):

PR #1080, opened 2026-07-26
IoForwardService first byte timeout

soundofspace:
First byte timeout to be able to detect broken connections during initial connection. In general this timeout is something that is really useful to configure with a short timeout, while idle_timeout in a lot of cases will be much longer so they serve different purposes

### DIFF
--- a/rama-core/src/stream/forward.rs
+++ b/rama-core/src/stream/forward.rs
@@ -63,6 +63,12 @@ pub enum BridgeCloseReason {
     /// handler that never invokes `signalServerDrain`) instead of
     /// wedging the bridge indefinitely.
     PausedTimeout,
+    /// The upstream (right / egress) half wrote no byte within the
+    /// configured first-byte window. Surfaces a silent origin that
+    /// accepts the connection but never responds, distinct from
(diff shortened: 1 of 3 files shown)


GlenDC:
LGTM overall, nice work — one fix before merge: disarm `first_byte_timeout` once `r_to_l` completes Ok (upstream EOF before any byte), otherwise the still-armed timer kills the half-close drain and reports a misleading `FirstByteTimeout` instead of `PeerEofRight` (a test for that case would be welcome too).

Question: is it intentional that the window starts at bridge open rather than at the client's first sent byte? Thinking of protocols/flows where the server isn't expected to speak until the client has said something — a slow-to-speak client would get cut even though the origin is fine.

GlenDC (approved):
Both points addressed nicely — approved. Two non-blocking nits:

- doc opener of `first_byte_timeout` still says "bridge opening"
- make `FirstByteTimeoutStart` `#[non_exhaustive]` per convention (in that module). Not even sure it has to be even exposed, but it does make sense so didn't comment on it as the alternative is probably worse.

PR #1087, opened 2026-08-01
feat: use systemd-resolved (varlink) for linux dns when available

GlenDC:
linux dns lookups now go through the systemd-resolved varlink socket if it's there, fully async so no more blocking pool threads. falls back per query to the old res_nsearch/getaddrinfo path. txt uses ResolveRecord, also added a fuzz target for the rr parsing.

based on feedback from some attendees at the copenhagen rust meetup where we introduced rama from core concepts up to advanced network stacks. cc @soundofspace

### DIFF
--- a/rama-dns/src/client/linux/mod.rs
+++ b/rama-dns/src/client/linux/mod.rs
@@ -509,6 +509,10 @@ fn lookup_txt_uncached_stream(
 
 /// Serve the lookup via systemd-resolved when a usable answer comes back,
 /// and stream from the native libc backend otherwise.
+///
+/// The native fallback deliberately runs with its own full timeout: during
+/// the short window before the breaker trips on a wedged daemon, a slow
+/// success beats failing queries at the configured deadline.
 fn resolved_first_stream<T, F, N, S>(
     varlink: Option<F>,
     native: N,
--- a/rama-dns/src/client/linux/mod.rs
+++ b/rama-dns/src/client/linux/mod.rs
@@ -1,5 +1,10 @@
 //! Linux-native DNS resolver.
 //!
+//! When systemd-resolved's varlink socket is reachable, lookups are served
+//! through it first — fully async, no blocking-pool worker — falling back to
+//! the libc paths below per query and whenever the daemon is unavailable
+//! (see [`super::systemd_resolved`]).
+//!
 //! On targets with `res_nsearch` support, `A` / `AAAA` / `TXT` lookups are
 //! backed by the native resolver stub. `res_nsearch` (not `res_nquery`) is
 //! used so the resolver walks the `search` list from `/etc/resolv.conf` and
@@ -30,7 +35,10 @@ use rama_utils::{
     str::arcstr::ArcStr,
 };
 
-use super::resolver::{DnsAddressResolver, DnsResolver, DnsTxtResolver};
+use super::{
+    resolver::{DnsAddressResolver, DnsResolver, DnsTxtResolver},
+    systemd_resolved::{self, ResolvedLookup, SystemdResolved},
+};
 
 mod cache;
 
@@ -70,6 +78,8 @@ pub struct LinuxDnsResolverBuilder {
     negative_cache_ttl: Duration,
     cache_capacity: u64,
     response_buffer_size: usize,
+    systemd_resolved: bool,
+    systemd_resolved_config: systemd_resolved::Config,
 }
 
 impl Default for LinuxDnsResolverBuilder {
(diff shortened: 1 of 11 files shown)


rama-dns/src/client/systemd_resolved.rs
```diff
@@ -0,0 +1,1767 @@
...
+        self: &Arc<Self>,
+        domain: &Domain,
+        timeout: Duration,
+    ) -> ResolvedLookup<Bytes> {
+        let name = wire_name(domain);
+        // ResolveRecord never applies resolv.conf search domains; keep
+        // single-label names on the native path where that expansion happens
+        if !name.contains('.') {
+            return ResolvedLookup::Unavailable;
+        }
```
soundofspace: This feels like a hack to make this work, in places like K8 it often happens that dns does contain `.` but not the complete search domain, so for certain cases this would work, but others not. Eg example to address a stateful service `http://router-0.router:80` (specific instance) or general `http://router:80`, full dns name contains more suffixes, but the main issue here is that it would be very weird why first one does not work, but second does. so i wonder if we should just document and never have this fallback

rama-dns/src/client/linux/mod.rs
```diff
@@ -139,10 +228,46 @@ impl LinuxDnsResolverBuilder {
...
                 self.negative_cache_ttl,
             )),
+            systemd_resolved: self
+                .systemd_resolved
+                .then(|| Arc::new(SystemdResolved::new(self.systemd_resolved_config))),
         }
     }
 }
 
+fn systemd_resolved_selected_by_nss() -> bool {
```
soundofspace, line 241: Should this be a onceLock so places where we create dns resolvers dont always have to do this io

rama-dns/src/client/systemd_resolved.rs
```diff
@@ -0,0 +1,1767 @@
...
+
+use std::{
+    fmt,
+    net::{Ipv4Addr, Ipv6Addr},
+    path::PathBuf,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
+    time::{Duration, Instant},
```
soundofspace: I think we should probably always use tokio version of instant so tests can be run in paused mod and no real sleep is needed

rama-dns/src/client/systemd_resolved.rs
```diff
@@ -0,0 +1,1767 @@
...
+        // single-label names on the native path where that expansion happens
+        if !name.contains('.') {
+            return ResolvedLookup::Unavailable;
+        }
+        let Some(claim) = self.claim() else {
+            return ResolvedLookup::Unavailable;
+        };
+        if !self.txt_supported.load(Ordering::Acquire) {
+            return ResolvedLookup::Unavailable;
+        }
```
soundofspace, line 249: Not sure if this should happen after the claim, i guess it should be before it?

soundofspace (changes requested):
Some minor remarks, but looks good

rama-dns/src/client/systemd_resolved.rs
GlenDC: good catch, and it was actually worse than inconsistent: a
negative from ResolveRecord for a relative name was treated
as authoritative, so `router-0.router` would silently
resolve to nothing while res_nsearch would have found it
via the search list. changed it so only rooted names
(`example.com.`) take a negative as final; relative names
now fall back to the native path on a negative, so both
your examples work, and the same way. kept the single-label
check purely as a fast path (that roundtrip can never
answer). remaining divergence is documented: a positive
as-is answer wins where ndots > 1 would have preferred a
search candidate first.

rama-dns/src/client/linux/mod.rs
GlenDC: yes, made it a OnceLock. process-lifetime snapshot is fine
here: the builder flag overrides whenever runtime control
is needed.

rama-dns/src/client/systemd_resolved.rs
GlenDC: done, the module uses tokio::time::Instant everywhere now
(free in prod, it is a transparent wrapper). the two
stale-probe tests run start_paused with advance, which also
drops the Instant::checked_sub hack that could fail on
hosts with a young monotonic clock. kept real time for the
socket tests, mixing paused time with actual io is where
auto-advance gets racy.

rama-dns/src/client/systemd_resolved.rs
GlenDC: the order is deliberate: claim() resets the pin inside its
unavailable->reprobe arm, so reading it after claim is what
lets a txt lookup drive the recovery probe and re-test
ResolveRecord support on a replaced daemon (pinned by
recovery_probe_rechecks_pinned_txt_capability). moved
before it, a txt-only workload could never un-pin. added a
comment saying so, and the early return now hands back the
claim overload-style so not even the theoretical race can
strand a probe slot for the 15s stale window.

soundofspace (approved):
lgtm

PR #1090, opened 2026-08-03
fix(cli): treat `resolve` as a known subcommand in default-send fallback

### LINKED ISSUE #1089: rama-cli: `resolve` missing from default-send fallback list, so `rama resolve <bad args>` reports a `send` error
## Problem

`rama-cli` falls back to the implicit `send` command when the first CLI argument is not a recognised subcommand (`rama-cli/src/main.rs`, the `Cli::try_parse()` error arm).

The guard list is:

```rust
["-V", "--version", "-h", "--help", "send", "serve", "probe"]
```

`resolve` is a real subcommand (`CliCommands::Resolve`) but is absent from that list. So when `rama resolve` is invoked with arguments clap rejects, the error is not surfaced. The arguments are re-parsed as `send` instead, and the user gets an error about `send` for a command they never asked for.

`-V` / `--version` / `-h` / `--help` and the other three subcommands are all handled correctly; `resolve` is the only one missing.

## Expected

`rama resolve <bad args>` should exit with the error clap produced for `resolve`, the same way `rama probe <bad args>` does today.

## Notes

Happy to send a PR. I have a fix that also derives the test expectation from clap's own subcommand list, so adding a future subcommand without updating the guard list fails a test instead of silently misrouting.

zeeshan8281:
Closes #1089

## What

`rama` falls back to the implicit `send` command when the first argument is not a recognised subcommand. `resolve` was missing from that guard list, so `rama resolve <bad args>` was re-parsed as `send` and reported an error about `send` rather than the resolve error clap had already produced.

`resolve` was the only `CliCommands` variant missing from the list.

## How

- Extract the guard list into `KNOWN_ROOT_ARGS` with an `is_known_root_arg` helper, and add `"resolve"`.
- Add a test that derives the expectation from clap's own subcommand list (`Cli::command().get_subcommands()`), so adding a subcommand without updating the guard list now fails a test instead of silently misrouting.

I verified the test actually catches the bug by removing `"resolve"` again:

```
thread 'tests::all_subcommands_are_known_root_args' panicked at rama-cli/src/main.rs:150:13:
subcommand `resolve` is missing from KNOWN_ROOT_ARGS
```

## Testing

```
cargo test -p rama-cli --bin rama     # 52 passed, 0 failed
cargo fmt -p rama-cli -- --check      # clean
cargo clippy -p rama-cli --bin rama --all-targets   # clean
```

I understand these changes in full and can respond to review comments. AI assistance was used in preparing this change; I have reviewed every line.

### DIFF
--- a/rama-cli/src/main.rs
+++ b/rama-cli/src/main.rs
@@ -53,6 +53,24 @@ enum CliCommands {
     Probe(cmd::probe::ProbeCommand),
 }
 
+/// Root arguments that must surface their own clap error rather than falling
+/// back to the implicit `send` command. Every [`CliCommands`] subcommand has to
+/// be listed here; `all_subcommands_are_known_root_args` enforces that.
+const KNOWN_ROOT_ARGS: [&str; 8] = [
+    "-V",
+    "--version",
+    "-h",
+    "--help",
+    "resolve",
+    "send",
+    "serve",
+    "probe",
+];
+
+fn is_known_root_arg(arg: &str) -> bool {
+    KNOWN_ROOT_ARGS.contains(&arg.trim())
+}
+
 #[tokio::main]
 async fn main() {
     #[expect(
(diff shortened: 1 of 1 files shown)


GlenDC:
Replaced by https://github.com/plabayo/rama/pull/1092

(see for explanation in the issue. Thanks for reporting it though @zeeshan8281 as well as the effort on trying to fix it yourself!)

PR #1093, opened 2026-08-05
Easily support grpc with custom codecs and no build steps

soundofspace:
Grpc does not require protobuf and supports basically any codec. This PR does mainly two things to build on that:
- Add support for a serde based codec (json included), but very easy to extend to any serde based encoding. Or any codec for that matter with serde as the reference implementation
- Macro to generate service contract without any build steps (including proto toolchain) using rust native types

Using grpc without protobuf is a very nice choice when everything you are doing is rust native. Protobuf requires proto types, generated types, conversion from rust to generated types... (Some common rust native types such as enums with some data variants are hard/annoying to define in proto). But by using a codec that supports these, you can circumvent almost all of these problems. The only real missing thing after switching codec is generated server/client definitions. This again could be done manually, but can be made very easy with a macro that is heavily inspired by the proto syntax / proto-build setup, but references native rust types, and codecs directly. Combine both of these and it is possible (and easy) to use grpc completely from within in rust using the business types you already have (as long as they are support by the codec), no need for proto files or generated types


GlenDC:
@soundofspace can you add context please to discuss motivation and also what use cases it actually unlocks outside the abstracted?

rama-grpc-build/src/manual.rs
```diff
@@ -297,21 +319,44 @@ impl RamaGrpcMethodBuilder {
...
+
+    /// Build a Method, falling back to `default_codec_path`
+    /// in case this Method defines no codec path of its own.
     ///
-    /// Panics if `name`, `route_name`, `input_type`, `output_type`, or `codec_path` weren't set.
-    #[must_use]
-    pub fn build(self) -> RamaGrpcMethod {
+    /// Panics if `name`, `route_name`, `input_type` or `output_type` weren't set,
+    /// or if there is no codec path to use.
+    fn build(self, default_codec_path: Option<&str>) -> RamaGrpcMethod {
```
GlenDC, line 346: Could we preserve the public zero-argument `build()` and built-method flow here to avoid breaking downstream users of the manual builder API?

rama-grpc-macros/src/service.rs
```diff
@@ -0,0 +1,439 @@
...
+    if first.ident == "crate" {
+        return path;
+    }
+
+    if first.ident == "self" {
+        first.ident = Ident::new("super", first.ident.span());
+        return path;
+    }
+
+    let mut segments = Punctuated::new();
```
GlenDC, line 431: Could this avoid rewriting normal dependency paths such as `rama_grpc::...` to `super::rama_grpc::...`, which fails in downstream crates?

rama-grpc-macros/src/tests.rs
```diff
@@ -0,0 +1,266 @@
...
+    assert!(code.contains("super :: Request"), "{code}");
+    assert!(!code.contains("super :: self"), "{code}");
+}
+
+#[test]
+fn generic_codecs_are_turbofished_and_their_arguments_qualified() {
+    let code = generate(
+        r#"
+        package = "test";
+        codec = SerdeCodec<JsonFormat>;
```
GlenDC: Could this become a compile test using `SerdeCodec<JsonFormat, _, _>`, since the generated `SerdeCodec<JsonFormat>` currently fails with E0107?

GlenDC (approved):
Besides the given inline comments, all looks good—the rest is solid.

rama-grpc-build/src/manual.rs
soundofspace: We could do that, but this was already broken I think. We did self.codec.unwrap() but didn't provide a setter for it, so as far as I could tell this never worked

rama-grpc-macros/src/service.rs
soundofspace: I could probably do some more magic here to make this work, but that means always checking if the first part is a crate name, which is kind of hard and might cause false positives/negatives. The way I saw it: by default everything is relative, if you need full path use `::rama_grpc` syntax, which was kind of made for this use case.

rama-grpc-macros/src/tests.rs
soundofspace: fixed

rama-grpc-macros/src/service.rs
GlenDC: ok but put it the other way around, why would rewriting to super be the assumd the default?

Been a while since I saw this code, so might take a refresher if you have the time

rama-grpc-build/src/manual.rs
GlenDC: hmm ok, would it be a lot of work to fix it now that we see it?

rama-grpc-macros/src/service.rs
soundofspace: This here is different in that the macro is supposed to be used inline, and have access to all the rust types you have in the place, code is generated in mod, so we super everything so it has access to all types / imports in that repo

soundofspace: So this works, without any special syntax
```rust
use rama::JsonCodec;
use api::OtherType;

impl_service! {
codec: JsonCodec
 fn Function(MyTypes) -> OtherType
 }
 
 struct MyTypes{}
 
 ```

rama-grpc-build/src/manual.rs
soundofspace: I mean this is the fix... What do you mean by fixing it, now you can use the manual build api, having a build() without any default would either need to manic if coded is not set anywhere, so i dont see a benefit in providing that option? vs just passing in None here

GlenDC: Fair. I guess I was too worried about something none existent as it turns out. Thx for pushing back

rama-grpc-macros/src/service.rs
GlenDC: Ok in that case let's try what we have now, I had a feeling that there is something to be fixed here, but feelings only go so far I suppose. Tests pass and your arguments are solid, so let's go for it.

GlenDC (approved):

PR #1095, opened 2026-08-06
Add route-aware connector services

GlenDC:
Generalizes client connectors around `ConnectRequest` and structured errors. Adds ordered proxy-route fallback, ProxyDB integration, pooling support, and bounded failure backoff.

Originally motivated by PAC-style routing, this foundation enables much broader routing flexibility and connector features in general.


rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
```diff
@@ -105,25 +120,33 @@ where
...
                 tracing::trace!(
                     "http proxy connector: no proxy required or set: proceed with direct connection"
                 );
                 let EstablishedClientConnection { input, conn } =
-                    self.inner
-                        .connect(input)
-                        .await
-                        .context("establish direct connection (no http proxy given or required)")?;
+                    self.inner.connect(input).await.map_err(|error| {
+                        error.context(
```
soundofspace, line 146: I guess map_err is needed because this error type has override for context? I guess this works, but will be slightly tricky in review to spot this

rama-http-backend/src/client/conn.rs
```diff
@@ -169,48 +262,56 @@ impl<S, Body> HttpConnector<S, Body> {
...
-    OpaqueError,
->
+) -> Result<EstablishedClientConnection<HttpClientService<BodyConnection>, Input>, OpaqueError>
 where
     IO: Io + Unpin + ExtensionsRef,
-    BodyIn: StreamingBody<Data: Send + 'static, Error: Into<BoxError>> + Unpin + Send + 'static,
+    Input: AuthorityInputExt
+        + ExtensionsRef
+        + HttpVersionInputExt
+        + TargetHttpVersionInputExt
```
soundofspace, line 285: Nice these traits are really starting to pay of

rama-net/src/client/conn.rs
```diff
@@ -20,6 +26,12 @@ impl<S: fmt::Debug, Input: fmt::Debug> fmt::Debug for EstablishedClientConnectio
     }
 }
 
+impl<S, Input: ExtensionsRef> ExtensionsRef for EstablishedClientConnection<S, Input> {
+    fn extensions(&self) -> &Extensions {
+        self.input.extensions()
+    }
+}
+
```
soundofspace: Im not sure this is a good idea. Basically in connectors you check input for the "input extensions", but all stored extension which become part of the connection or are connection info should be stored on the connection. So its really important to know that there are two options and both serve different goals, this shortcut hides that more and will probably cause more issues since its not explicit anymore so harder to see mistakes

soundofspace (approved):
Nice, amazing work, will be fun to play with

rama-net/src/client/conn.rs
GlenDC: good point, gonna revisit it.

rama-http-backend/src/client/conn.rs
GlenDC: agreed

rama-http-backend/src/client/proxy/layer/proxy_connector/service.rs
GlenDC: No it's mostly a performance enhancement. So that we do not need to box and unbox. which is what it will do otherwise, given that `Result<T, E>` turns `E` always into a BoxErorr for context. It will also work if we didn't do the map, but you are than just boxing for no reason.

soundofspace (approved):

PR #1120, opened 2026-08-16
Establish proxy egress during handshakes by default

GlenDC:


GlenDC:
FYI @soundofspace working on porting a TUI/GUI proxy tool's internal network stack to make use of Rama. That and also your earlier comments from past made it clear that it is time for me to start supporting this. I am tagging you as I know you wanted to contribute something like this as well. I will merge it later today, but please have a look when you can in case it is not what you had in mind or is missing something or wrong in any other way. This way we can still rectify it prior to releasing rama 0.4 coming week.

rama-http/src/layer/upgrade/http_proxy_connect/eager.rs
```diff
@@ -0,0 +1,281 @@
...
+) -> Response {
+    tracing::debug!(
+        server.address = %authority.host,
+        server.port = authority.port,
+        error = ?err,
+        "HTTP proxy CONNECT egress connection failed",
+    );
+
+    if err.kind() == ConnectionErrorKind::Timeout {
+        StatusCode::GATEWAY_TIMEOUT.into_response()
```
soundofspace, line 127: If i'm not mistaken this should work with TimeoutLayer which makes me doubt if we need connect_timeout on this service, having a TimeoutLayer in the connector stack seems more explicit and is probably a pattern we should explicitly do. I see connectors stuck in a lot of cases so promoting a TimeoutLayer pattern seems like a good default. The only downside with that layer is we need a downcast from boxedError internally, but that should be OK?

rama-http/src/layer/upgrade/service.rs
```diff
@@ -36,20 +42,138 @@ pub struct UpgradeResponse<I, O> {
...
+}
+
+struct PreparedUpgrade<O> {
+    response: O,
+    request: Request,
+    extensions: Extensions,
+    continuation: UpgradeContinuation,
+}
+
+impl<O: Send + Sync + 'static> UpgradeHandler<O> {
```
soundofspace: Is Sync bound needed on output?

soundofspace (commented):
Looks great, I always though it was somewhat weird that upgradeLayer forced a split between responder and handler. This new API is way more natural and cleaner then forcing everything through extensions

rama-http/src/layer/upgrade/service.rs
GlenDC: nope, good catch, removed it

rama-http/src/layer/upgrade/http_proxy_connect/eager.rs
GlenDC: good point, removed the custom timeout stuff and use `TimeoutLayer` now

PR #1141, opened 2026-08-27
fix http client pool reuse and waiter wakeups

GlenDC:


GlenDC:
FYI @soundofspace , these were fixes noticed while making changes in the codebase in and around those locations in regards to some other tasks... Will be merged once green, but letting you know in case you think I did something stupid here.

PR #1147, opened 2026-08-31
update dial9 to 0.5 and make tokio_unstable optional

jv1i:
Upgrades dial9 0.3 -> 0.5

Moves rama to the `dial9` 0.5 crate (dial9-tokio-telemetry is no longer the main crate to depend on), and drops `--cfg tokio_unstable` as a build requirement.

## API

0.5 removed `Dial9Config` and `TracedRuntime`, users now build a `Recorder` (used to say what and where to record), and attach a Tokio runtime to it (see [release page](https://github.com/dial9-rs/dial9/releases/tag/dial9-v0.5.0)). 
Before you also had to pass your Tokio settings into dial9 and it built the runtime, now you build the runtime and pass it in. 

In summary, the main changes are:

- `with_dial9_config(Dial9Config)` becomes `with_dial9_recorder(dial9::Recorder)`, on both `rt::blocking::RuntimeBuilder` and `DefaultTransparentProxyAsyncRuntimeFactory`.
- A new `with_dial9_attach_options(TokioAttachOptions)` on both, for the per-runtime knobs (task tracking, task dumps, Tokio hooks, runtime name) that used to go inside `Dial9Config`.
- `OwnedRuntime::from_dial9` takes a `dial9::AttachedRuntime` (a recorder + its runtime) and keeps the recorder up.

Some minor cleanup that 0.5 made possible:

- `shutdown_bounded` doesn't need a dispose thread: in 0.3 `TracedRuntime` bundled the runtime and the guard, handing out only references, so neither `shutdown_timeout` nor `graceful_shutdown` (both consuming) were reachable. Holding the two separately now makes both reachable, so the `ManuallyDrop` + dispose-thread + channel trick is gone (including unsafe)
- `OwnedRuntimeHandle` doesn't need the dial9 handle field, `spawn_in` resolves the session on the target worker at first poll, so nothing needs capturing up front.
- dial9 used to build the runtime, so rama's `with_worker_threads` could never reach it and pairing the two was rejected. Now rama builds it, so the worker count applies. Enabled telemetry still needs multi-thread, so `with_current_thread()` is still rejected.
- Decoding is serde-based now and driven by the on-wire schema, so `TraceField` no longer has a decode side. rama's impls drop `type Ref<'a>` and `decode_ref`.

## tokio_unstable

dial9 now builds without tokio unstable, so we remove it from `.cargo/config.toml` and CI jobs (left just one `test-dial9-tokio-unstable`).

Tradeoff for default builds without unstable (users can still opt in): poll events come from dial9's own spawn helpers rather than from tokio, so the task timeline only covers what goes through `rama_core::rt::Executor`. 
That is most of it, since the accept loop and rama-http-core's h2 internals already spawn that way. 
I see that there are some places using bare `tokio::spawn` (socks5 DNS racing, the HTTP proxy connection driver, HAR workers, DNS cache refresh, fastcgi accept) and will go untraced. Task spawn/terminate events and per-worker queue depth need the flag regardless.

Up for discussion / follow-up:
Perhaps `rt::executor::spawn` could be made public and have the places mentioned above start using it? (I'm guessing it wouldn't be `Executor::spawn_task` since that also opts into graceful shutdown)

## Checklist

- [x] This change is linked to an issue we have discussed (see above). (discussed on discord)
- [x] `just qa` passes locally (or I have explained below what I could not run).
- [x] I agree to the [contribution terms](https://github.com/plabayo/rama/blob/main/CONTRIBUTING.md#contribution-terms)
      and the Developer Certificate of Origin.

### DIFF
--- a/rama-core/src/rt/dial9_test_util.rs
+++ b/rama-core/src/rt/dial9_test_util.rs
@@ -13,7 +13,7 @@ pub(crate) fn recorder_slot() -> parking_lot::MutexGuard<'static, ()> {
 pub(crate) fn recorder(trace_dir: &std::path::Path) -> ::dial9::Recorder {
     let writer = ::dial9::DiskBuffer::builder()
         .base_path(trace_dir)
-        .max_file_size(1024 * 1024)
+        .max_file_size(mib(1))
         .max_total_size(4 * 1024 * 1024)
         .build();
     let recorder = ::dial9::recorder_or_disabled(writer).build();
--- a/rama-core/src/rt/dial9_test_util.rs
+++ b/rama-core/src/rt/dial9_test_util.rs
@@ -14,7 +14,7 @@ pub(crate) fn recorder(trace_dir: &std::path::Path) -> ::dial9::Recorder {
     let writer = ::dial9::DiskBuffer::builder()
         .base_path(trace_dir)
         .max_file_size(mib(1))
-        .max_total_size(4 * 1024 * 1024)
+        .max_total_size(mib(4))
         .build();
     let recorder = ::dial9::recorder_or_disabled(writer).build();
     assert!(
--- a/rama-core/src/rt/dial9_test_util.rs
+++ b/rama-core/src/rt/dial9_test_util.rs
@@ -1,5 +1,7 @@
 //! Shared setup for the tests that exercise the dial9 integration.
 
+use rama_utils::octets::mib_u64;
+
 /// Serializes tests that build an enabled dial9 recorder (a second `build()`
 /// while one is alive returns a disabled recorder).
 pub(crate) fn recorder_slot() -> parking_lot::MutexGuard<'static, ()> {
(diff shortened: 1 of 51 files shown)


GlenDC:
> Up for discussion / follow-up:
Perhaps rt::executor::spawn could be made public and have the places mentioned above start using it? (I'm guessing it wouldn't be Executor::spawn_task since that also opts into graceful shutdown)

Yes sounds reasoanble to me. Gonna look into it as a follow-up PR. Good idea for sure :)

rama-core/src/rt/dial9_test_util.rs
```diff
@@ -0,0 +1,25 @@
...
+    SLOT.lock()
+}
+
+/// Build an enabled recorder writing into `trace_dir`.
+///
+/// Callers must hold [`recorder_slot`].
+pub(crate) fn recorder(trace_dir: &std::path::Path) -> ::dial9::Recorder {
+    let writer = ::dial9::DiskBuffer::builder()
+        .base_path(trace_dir)
+        .max_file_size(1024 * 1024)
```
GlenDC:
```suggestion
        .max_file_size(mib(1))
```

rama-core/src/rt/dial9_test_util.rs
```diff
@@ -0,0 +1,25 @@
...
+        .max_total_size(4 * 1024 * 1024)
```
GlenDC:
```suggestion
        .max_total_size(mib(4))
```

GlenDC (approved):
Thanks a lot, looks mostly good, found some minor things also tracked your open questions, gonna fix it while I am at it

GlenDC:
Gonna watch CI and once green merge it. But @jv1i please do check my last commit and see if what I did make sense to you.

jv1i:
Your changes look good! good catch on the `#[track_caller]`. I just noticed the CI errors: the semver one is from my changes (I think maybe missing `--cfg tokio_unstable`), and the apple FFI errors jobs be due to my changes well, it's late here so I'll go over these tomorrow :)

PR #1183, opened 2026-09-16
feat(http-types): push-driven sse EventDecoder

bitterpanda63:
## What does this PR do?

This adds `EventDecoder<T>`, a push-driven, synchronous SSE decoder. It exposes the decoder that already sits behind `EventStream` rather than adding a new one.

No issue exists for this yet. The draft is here to show the API we'd like; happy to open an issue and rework it.

**Why.** `EventStream` needs to own the body as a `Stream`. A proxy that inspects a response while forwarding it unchanged (for example inside `BodyExt::inspect_frame`) only sees each chunk as it passes, so today it has to write its own line splitter. We do exactly that in endpoint-protection to read tool calls off AI provider SSE streams.

**API:**

```rust
let mut decoder = EventDecoder::<String>::new().with_max_line_len(1024 * 1024);
decoder.decode(chunk, |event| { /* ... */ })?; // once per chunk
decoder.finish()?;                             // at end of body
```

- **Decoding:** `decode` runs the same decode path as `EventStream`, including CR/LF/CRLF, BOM and UTF-8 carry across chunks. It drains the internal ready queue between chunks, so the existing soft caps still apply.
- **Errors:** an error is fatal, and later calls keep failing, matching how `EventStream` terminates on a decode error.
- **Line cap:** `max_line_len` is optional and off by default. It bounds a single line, including one still being buffered, so a long-lived stream can be capped per line instead of only by a cumulative body limit.

## Checklist

- [ ] This change is linked to an issue we have discussed (see above).
- [ ] `just qa` passes locally (or I have explained below what I could not run).
- [ ] I agree to the [contribution terms](https://github.com/plabayo/rama/blob/main/CONTRIBUTING.md#contribution-terms)
      and the Developer Certificate of Origin.

**What ran:** `cargo test -p rama-http-types --lib sse` (67 passed, 7 of them new) and `cargo clippy -p rama-http-types --tests`. Both ran on `0155390` with this patch applied, because the machine they ran on can't fetch some of current `main`'s newest dependencies. `just qa` has not been run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


GlenDC:
I understand what you are trying to achieve. Going to close this in favour of a PR where I'll make sure there is a common push based decoder available that is shared between both approaches.

That said, what are you currently doing with the decoded events? In the end these are stuck in a closure so I imagine you are for now also sending something as a result over a channel back to where you need the data? 

Know that body's also allow you to "capture". In contrast to frame inspection do you get a body stream via a sink of your choice. Thus allowing you to make an event stream on top of this duplicate stream. Under the hood this works with bytes clones which is an atomic cheap operation (in same space as Arc clones).

Would that not achieve the same, even without needing an API change within rama?

Like I said I will make it possible in a follow-up PR to achieve what you a re trying to achieve here. That said, I am not sure you even truly need it for the use case you are currently doing.

To be continued.

PR #1188, opened 2026-09-17
fix(http): gate server re-export behind http-backend

seuros:
The `server` re-export from rama-http-backend was gated on `http-full`, but the dependency itself is pulled in by `http-backend`. 

Building with `http-backend` alone made the module unreachable.


GlenDC (approved):
good catch, you did miss the proxy one, so I also fixed that for you while you were at it.

PR #1193, opened 2026-09-21
Less fragile build for rama-dns

soundofspace:
Having it crash on warnings is fine if we fully control the build, however since external users also run this build, crashing on any warning is painful. Seeing how simple this is, I just removed that logic, but we could also add a dedicated CI job to check this, but exclude it from the build.rs (or have it behind some special env flag)

Official docs on this flag are:
```
Warning: turning warnings into errors only make sense if you are a developer of the crate using cc-rs. Some warnings only appear on some architecture or specific version of the compiler. Any user of this crate, or any other crate depending on it, could fail during compile time.
```


GlenDC:
Perhaps make env variable for it that is clear this is for dev purposes only and set it in the relevant CI jobs. Should be easy enough to add as part of this PR. WDYT?

GlenDC (approved):

GlenDC:
For historical purposes @soundofspace can you here in this PR also as a comment add a log snippet to show the errors you were getting, together with the target/platform info , rama hash and rama features used

soundofspace:
I was building in an environment that set `-D_FORTIFY_SOURCE=3` which caused a warning when dependencies where being build with the default `CFG_OPT_LEVEL` on a debug build. Setting that to `CFG_OPT_LEVEL=3` fixed my issues, but it also made me "fix" this, since an unrelated warning should no break a complete build.

The error seen was only this, and it took some time to figure out what happened
```
  error occurred in cc-rs: command did not execute successfully (status code exit status: 1): LC_ALL="C" "clang" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-g" "-gdwarf-4" "-fno-omit-frame-pointer" "-mno-omit-leaf-frame-pointer" "-m64" "--target=x86_64-unknown-linux-gnu" "-Werror" "-Wall" "-Wextra" "-o" "/home/runner/work/proxy-stack/proxy-stack/sdk/tauri/target/debug/build/rama-dns-cf97e223d9a78ea2/out/f743ee7cfe25a682-resolv_wrapper.o" "-c" "src/client/linux/resolv_wrapper.c"
  ```

All of this is mostly unrelated to which version/features were used, and were caused by external env warning

PR #1194, opened 2026-09-22
Use DropGuard for otel http metrics

soundofspace:
Use dropguard so we don't forget to count cancellations.


GlenDC (approved):
Good catch. Do you have numbers in some of your production systems on how high that active counter climbed up over time?

soundofspace:
> Good catch. Do you have numbers in some of your production systems on how high that active counter climbed up over time?

Could get up into the millions with only a few thousand active, but it was slow to grow. It's just something that barely restarts, so over time this was very noticeable.

